PR #2, opened 2022-01-24
Fixes for custom labels dropdown

CirkaN:
This fixes the bug with a dropdown on the custom labels settings page.
After selecting a word from a dropdown and removing it from the list, the word was removed from the dropdown, this fix fixes this behavior.


beganovich:
Thanks for your contribution! :+1:

PR #10, opened 2022-02-10
Tests for Format Number class

turbo124:


src/common/helpers/invoices/invoice-sum-inclusive.ts
```diff
@@ -0,0 +1,249 @@
...
+  protected async calculateDiscount() {
+    this.totalDiscount = this.discount(this.invoiceItems.subTotal);
+    this.total -= this.totalDiscount;
+
+    return this;
+  }
+
+  protected async calculateInvoiceTaxes() {
+
+    let amount = this.total;
```
beganovich, line 52: This variable isn't used anywhere. Should we return it or update some state based on it, or?

beganovich (changes requested):
Looks good, slight change needed.

src/common/helpers/invoices/invoice-sum-inclusive.ts
turbo124: pushed changes for this. needed to actually use the var :)

beganovich:
Thank you for your contribution! :+1:

PR #11, opened 2022-02-11
extraction of resourceful actions into separate component

notks:


src/components/ResourcefulActions.tsx
```diff
@@ -0,0 +1,42 @@
...
+type Props = {
+  resource: string;
+  actions: Actions[];
+};
+
+export default function RecorcefulActions(props: Props) {
+  const [t] = useTranslation();
+  return (
+    <Dropdown label={t('actions')}>
+      {props.actions.map((action) => {
```
beganovich, line 28: As far as I know, this should throw a warning in both console & editor with missing key prop?

beganovich (commented):
Looks good, slight change.

src/components/ResourcefulActions.tsx
notks: I did that just forgot to add it to pr. It should be good now.

beganovich:
Thanks for your contribution! :+1:

PR #14, opened 2022-02-14
 link to resource create from the sidebar

notks:


beganovich:
Looks good, needs minor changes.

1. Labeling is wrong. I see you made use of the resource itself, but this produces "New Clients" (plural). This is not good. 
To fix this, make rightButton implement the following interface:

```ts
interface RightButton {
  icon: ReactNode;
  to: string;
  label: string;
}
```

With this, you fix the label, eliminate the need for checking if the item is the dashboard. Code change would go something like this:
   
```tsx
{
  item.rightButton && (
    <Link
      to={item.rightButton.to}
      title={item.rightButton.label}
      className="..."
    >
      <item.rightButton.icon />
    </Link>
  );
}
```

2. Rename RightButton to ActionButton (actionButton for variable).

3. Update action button CSS to:

```css
hover:bg-gray-200 hover:bg-opacity-10 rounded-full p-1.5
```

beganovich:
Looks awesome, thanks for your contribution! :+1:

PR #27, opened 2022-03-14
changed layout to culomns in vendor resourceful pages

notks:


beganovich:
Thanks for your contribution :+1:

PR #28, opened 2022-03-16
Product custom actions

notks:


beganovich:
Thanks for the PR. Changes needed:

1. Remove cloning to invoices. This is fundamentally wrong since you technically create a new page for the invoice. This is not how it should be done, so it can be removed (for now). We'll need to discuss this.

2. We'll need to unite form for create/clone. This shouldn't be driven by the "initial values" concept, but the response from `/products/create` response. Take that response and pass it to the create component, modify the data & push it to the backend. Same for cloning. Just pass existing products entity into clone component.

3. Structuring the files. If it's a reusable component, it should be in the top-level `common` directory. Eg. `pages/products/common/components`. If it's a locally used component, it should be in `products/create/components`.

4. Default quantity is missing in create/clone component.

beganovich:

PR #31, opened 2022-03-25
Custom actions for client

notks:


src/common/queries/clients.ts
```diff
@@ -33,3 +33,17 @@ export function useClientQuery(
...
+  action: 'archive' | 'delete'
+): Promise<AxiosResponse> {
+  return request(
+    'POST',
+    endpoint('/api/v1/clients/bulk'),
+    {
+      action,
+      ids: Array.from(id),
+    },
+    { 'X-Api-Token': localStorage.getItem('X-NINJA-TOKEN') }
```
beganovich, line 47: Please use `defaultHeaders` object.

src/pages/clients/show/Client.tsx
```diff
@@ -76,7 +84,34 @@ export function Client() {
       href: generatePath('/clients/:id/recurring_expenses', { id }),
     },
   ];
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const onSave = (password: string, passwordIsRequired: boolean) => {
+    const toastId = toast.loading(t('processing'));
+    if (password == '') {
+      toast.dismiss();
+      toast.error(t('password_error_incorrect'));
```
beganovich, line 92: This is wrong. We need to integrate validation errors into `ValidationBag` interface so we can access errors as if they were from the backend.

src/pages/clients/show/Client.tsx
```diff
@@ -76,7 +84,34 @@ export function Client() {
...
+      return 1;
```
beganovich, line 93: Not entirely sure what's return 1. Return itself should be enough.

src/pages/clients/show/Client.tsx
```diff
@@ -87,7 +122,54 @@ export function Client() {
             {t('edit_client')}
           </Button>
 
-          <Dropdown className="divide-y" label={t('more_actions')}></Dropdown>
+          <Dropdown className="divide-y" label={t('more_actions')}>
+            <DropdownElement
+              key={'archive'}
+              onClick={() => {
+                const toastId = toast.loading(t('processing'));
```
beganovich, line 129: Please extract this into the method. Mix and match between inline functions/extracted methods shouldn't be present.

src/pages/clients/show/Client.tsx
```diff
@@ -87,7 +122,54 @@ export function Client() {
...
+                    toast.error(t('error_title'));
+                  });
+              }}
+            >
+              {t('archive')}
+            </DropdownElement>
+            <DropdownElement
+              key={'delete'}
+              onClick={() => {
+                const toastId = toast.loading(t('processing'));
```
beganovich, line 148: Needs method extraction.

beganovich (changes requested):
Minor changes needed for the first pass.

PR #33, opened 2022-03-25
line item arrows

notks:


beganovich:
This is good, but we have to ensure the number is inserted. This is not doing anything regarding that.

src/pages/invoices/common/hooks/useResolveInputField.tsx
```diff
@@ -73,11 +73,14 @@ export function useResolveInputField() {
...
           value={invoice?.line_items[index][property]}
-          onChange={(event: ChangeEvent<HTMLInputElement>) =>
-            onChange(property, parseFloat(event.target.value), index)
-          }
+          onChange={(event: ChangeEvent<HTMLInputElement>) => {
+            if (isNaN(Number(event.target.value)) || event.target.value == '') {
+              event.target.value = '0';
+              return event;
+            }
+            return onChange(property, parseFloat(event.target.value), index);
```
beganovich, line 82: This is not good and not convention. Please avoid placing inline logic whenever possible. The extracted method would be much nicer.

Regarding this example, you'll need to extract it into method anyways, because we want to perform this validation for the recurring invoices as well.

beganovich (changes requested):
Minor changes needed.

beganovich:

PR #34, opened 2022-03-25
disabled adding empty client 

notks:


beganovich:
This is not the usual way we show validation errors. Please use ValidationBag interface as a reference and use the errorMessage prop in the input field to show the validation message. 

Also return 1 can be changed to return, itself.

src/pages/clients/create/Create.tsx
```diff
@@ -66,9 +66,23 @@ export function Create() {
...
   const onSave = () => {
     set(client as Client, 'contacts', contacts);
-
     const toastId = toast.loading(t('processing'));
+    setErrors(undefined);
+    if (
+      !(
+        client?.name != '' ||
+        contacts[0].first_name != '' ||
+        contacts[0].last_name != ''
```
beganovich, line 75: Requirements have not changed since the original task. In the original commit, only the client name was referenced for this, so I am not entirely sure where the validation for contact came from. Even more, why only validate the first contact. 

Please stick to original requirements and validate the client name only.

src/pages/clients/create/Create.tsx
```diff
@@ -66,9 +66,23 @@ export function Create() {
...
+      )
+    ) {
+      setErrors({
+        message: 'Invalid data',
```
beganovich, line 79: Application uses dynamic translations. Please translate ALL strings.

beganovich:
I will take this & make changes.

PR #35, opened 2022-03-25
product create modal

notks:


beganovich:
This is the wrong approach to doing this. When you run this modal background is non-existing. We should render modal in top-level (on invoices page) and then pass isVisible prop and setter to parent component to toggle it from the input field (denounced combobox).

beganovich:

PR #37, opened 2022-03-29
added missing titles to pages

notks:


beganovich:
If you inspect useTitle hook, it has translation built-in. That said, it is not required to pass t(string) into hook.

PR #40, opened 2022-04-02
paid to date format

notks:


beganovich:
I am not entirely sure if this is how it's meant to be, but here, you are removing formatting. Why?

PR #41, opened 2022-04-02
fixed document delete action and added label

notks:


beganovich:
Thanks for the contribution :+1:

PR #45, opened 2022-04-02
set default remaining cycles to endless

notks:
This would be a lot cleaner if done by the api itself when returning blank invoice.


beganovich:
@turbo124 thoughts on this?

beganovich:
This has changed upstream. Now if you call /create endpoint it will return -1.

PR #48, opened 2022-04-03
hide connect google button on selfhosted

notks:


beganovich:
Thanks for contribution :+1:

PR #51, opened 2022-04-04
Inject created client into invoice

notks:


src/components/tables/Pagination.tsx
```diff
@@ -17,7 +17,7 @@ interface Props extends CommonProps {
   currentPage: number;
   onPageChange: any;
   onRowsChange: (rows: string) => any;
-  totalRecords: number;
+  totalRecords?: number;
```
beganovich, line 20: This has changed upstream. Please merge the latest changes.

beganovich (commented):
Looks good, however, I think it's a wrong approach UX-wise to do it like this. Say I create the client and then select the other one? There's no way for me to go back to a freshly created client, right?

I have a similar task with line items, I suggest waiting on me for this and then changing the implementation :+1:

PR #53, opened 2022-04-08
fixed custom labels submit on enter key

notks:


beganovich:

PR #60, opened 2022-04-16
removed backdating from recuring invoices

notks:


beganovich:
Thanks for the contribution :+1:

PR #61, opened 2022-04-17
disabled auto select on search in combobox

notks:


beganovich:
Hi,
Can you please explain what this feature is & why is it useful?

notks:
Hi,
When trying to search for records if none with exact name are found it used to auto add record with search query as record data . This prevents that from happening so user must select one of options found with search query.

beganovich:
This is not the way to do it. It breaks core functionality where we need shadow records to handle. That's why it's called Combobox, it allows input + searchable options.

PR #64, opened 2022-04-21
disabled new client button when client is selected

notks:


beganovich:
Thank you for the contribution, however, I think it's unfair to leave it like this. There's no way to dismiss the client after we select it? 
The best would be if we could add the "X" button to the right which will show only when the client is selected.

image

beganovich:
Looks good, thanks for the contribution! :)

PR #67, opened 2022-04-22
Set default quanity to 1

notks:


beganovich:
This looks good, but it's incomplete. At the moment we are changing the qty to 1 no matter which case, which means when we edit the invoice/recurring invoice and there's already a line item with quantity, let's say, 15, it will get back to 1 once you change the product.

Please update the quantity only if the current value is less than 1.

beganovich:
Looks good, thanks for the contribution :)

PR #70, opened 2022-04-23
translations of table records count

notks:
I am not sure if this is the way to do it as it may break translations if they are used somewhere else based on previous use case.


beganovich:
This is not the right way to do translations. The backend has no way of knowing about it this way. I suggest talking to team chat about handling translations, as we need to PR this first to the main repo.

PR #74, opened 2022-04-28
Autobill and payment type defaults

notks:


beganovich:
Looks good, one change is needed.

On create page for recurring invoices, please use `company.settings.auto_bill` but on the edit page for the recurring invoice, make use of the `recurring_invoice.auto_bill` value.

beganovich:
Looks good, thanks for the contribution :)

PR #75, opened 2022-04-29
Add auto archive cancelled invoice property

turbo124:


beganovich:
Looks good, thanks for the contribution! :)

PR #79, opened 2022-04-30
Removing selected invoices from combobox

notks:


src/components/forms/DebouncedCombobox.tsx
```diff
@@ -30,6 +30,7 @@ interface Props {
   clearButton?: any;
   onClearButtonClick?: any;
   inputLabel?: string;
+  filterRecords?: any;
```
beganovich, line 33: This is making following & tracking code down super complex. Please annotate and don't use any.

src/components/forms/DebouncedCombobox.tsx
```diff
@@ -184,15 +185,30 @@ export function DebouncedCombobox(props: Props) {
...
+                  <Combobox.Option
+                    className="cursor-pointer select-none relative py-2 px-3 text-gray-900 hover:bg-gray-100"
+                    key={index}
+                    value={record}
+                  >
+                    {record.label}
+                  </Combobox.Option>
+                );
+              }
+            })}
```
beganovich, line 211: This is a bad way of filtering stuff. If I get it correctly, you are performing filtering on each record?

I'd suggest the following way of doing this.

1. Passing filter function w/ callback to DebounceCombobox component that as arguments will return all arguments.
2. Trigger filter function on each change of the DebounceCombobox
3. Watch for change using useEffect and perform filtering top-level.

src/pages/payments/create/Create.tsx
```diff
@@ -128,6 +128,31 @@ export function Create() {
...
+        {
+          _id: v4(),
+          amount:
+            newInvoice.balance > 0 ? newInvoice.balance : newInvoice.amount,
+          credit_id: '',
+          invoice_id: newInvoice.id,
+        }
+      );
+  };
+
```
beganovich, line 155: I assume this is related to the filtering itself. Not sure why we do this, maybe we should communicate about this. It's hard to follow it this way.

src/pages/payments/create/Create.tsx
```diff
@@ -177,7 +202,10 @@ export function Create() {
           {formik.values.client_id && (
             <>
               {formik.values.invoices.map(
-                (record: { _id: string; amount: number }, index) => (
+                (
+                  record: { _id: string; amount: number; invoice_id: string },
+                  index
+                ) => (
```
beganovich, line 208: If possible, extract this to interface so it can be reused,

beganovich (changes requested):
Generally, it works, however, I don't think it's the most performant & clean way to do it. I suggested a way to filter it in another comment.

PR #80, opened 2022-04-30
Show only payable invoices

notks:


beganovich:
Looks good, thanks for the contribution. 

Btw. please make use of `generatePath` instead of template literals for route parameters.

PR #111, opened 2022-05-27
Unset documents array on clone

turbo124:


beganovich:
Looks good, thanks for the contribution! :+1:

PR #170, opened 2022-07-22
Fixes for vendors

turbo124:


beganovich (approved):
Looking good, thanks for contribution!

PR #201, opened 2022-09-08
Fixes for recurring invoice start date

turbo124:
Changing this from `date` to `next_send_date`

One point to note is using native date javascript functions are adjusting the props to the client timezone.. this is wrong, we need to show the date as returned from the API.

I'm am simplying slicing the string if it exists, but this is something we'll want to be careful of elsewhere in the application.


turbo124:
Updated to use day js :rocket:

beganovich:
Awesome, thanks so much!

PR #202, opened 2022-09-11
Invoices without Redux

beganovich:
This moves invoices away from Redux. 

To do: 
- [ ] Cloning features (to recurring invoices, credits, quotes)
- [ ] Remove obsolete files and redux slices

Functionality that is gone (temporarily):
- Cloning invoices to quotes
- Cloning invoices to credits
- Cloning invoices to recurring invoices


beganovich:
Closing in a favor of one big PR that includes complete drop of Redux from resources.

PR #233, opened 2022-10-13
Improve activities

beganovich:
This improves activities, making it more similar to version 4, where each entity is clickable.

image


beganovich:
Rebased of off wrong branch. Reopening new PR.

PR #249, opened 2022-11-09
Improve dashboard cards styles

beganovich:
As per task requirements (RU-298), following had to be implemented.

- [x] Making all boxes same height
- [x] Cards needs to have less padding
- [x] For recent payments, rename "Invoice number" to "Invoice"
- [x] Remove IP address
- [ ] Implement activities on the client page itself

Skipped:
- Card titlebars should use primary color: This is totally off when it comes to rest of the app. Putting accent color on the card makes every other card look incoherent.

- Sub cards need to be grayish color: I think we should keep tables headers accent color. Not change them to gray, because previously mentioned task isn't implemented. Gray on white doesn't create enough contrast for such important element.

Missing:
- [ ] IP address have been removed, as proposed, we should implement this functionality by adding "Activities" tab on the client page itself.

### DIFF
--- a/src/components/cards/Card.tsx
+++ b/src/components/cards/Card.tsx
@@ -51,10 +51,10 @@ export function Card(props: Props) {
       <form onSubmit={props.onFormSubmit}>
         {props.title && (
           <div
-            className={classNames('px-4 sm:px-6 border-b border-gray-200', {
+            className={classNames('border-b border-gray-200', {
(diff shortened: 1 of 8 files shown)


beganovich:
@turbo124 do we want to merge this before creating activities page on the client? With this PR we are loosing ability to see action IP address.

turbo124:
Yes please merge

PR #265, opened 2022-11-17
[Feature]366-Company Details Tabs

Civolilah:
@beganovich @turbo124 PR is updated with implemented custom fields. Here is the screenshot:

customFieldsCompanyDetails


turbo124:
@Civolilah 

This is look good. One small point, on the documents tab, we can remove one level of nesting, so instead of displaying

image

we can show the actual page

image

turbo124:
Forgot to mention, the background color of the TAB bar should match the primary_color of the company.

this property can be found  company?.settings?.primary_color

Civolilah:
> Forgot to mention, the background color of the TAB bar should match the primary_color of the company.
> 
> this property can be found company?.settings?.primary_color

I will replay here for both suggestions with one screenshot. I think that is exactily as you want. Please let me know if it looks good.

companyDetailsNew

turbo124:
@Civolilah this looks good,

@beganovich we should also consider making table headers all primary_color too?

Civolilah:
> @Civolilah this looks good,
> 
> @beganovich we should also consider making table headers all primary_color too?

@turbo124 @beganovich  Okay, thank you for you quick answer. I noticed on the Jira that we have a couple of tasks related to those tabs. Tabs for generate numbers are in progress now, so I will globalize this component on that PR and once we merge it I will apply on this PR and implement for others. While I wait for approval on that PR I will work on different type of tasks from Jira.

beganovich:
I don't think this looks good, few things:

- Tabs are fundamentally wrong here. They're not deep-linked and there's no way to link user to "Company settings > Details" for example?
- I don't think tabs look great, nor they'll look good with primary color applied to them, certainly not with this materialish design applied with deep shadows.
- If we decide to apply primary color to tabs (which I truly hope we don't), all tabs will have to respect this, so base component needs modification.
- We have tab components in the place. Please have a look at Client overview page, there you can see deep linked tabs component which handles active tab, deep linking and all styles you need by wrapping your content in outlet component. There is no need to reinvent. If you need client-sided tabs we have component for them also (in this case you don't).
- For primary color please use `useAccentColor` hook.
- Tables are following accent color, not sure what's happening there.

image

turbo124:
The only issue I have here is there is not a lot of contrast in the tabs, so for a new user, they may not recognize that these are actually Tabs.

@beganovich thoughts?

beganovich:
@turbo124 hm, I mean we do use same component on other pages and it wasn't an issue before? I am open for changes.

turbo124:
Understood, we can iterate on this if needed.

src/pages/settings/company/common/hooks/useCompanyDetailsTabs.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+import { route } from 'common/helpers/route';
+import { Tab } from 'components/Tabs';
+import { useTranslation } from 'react-i18next';
+
+export function useCompanyDetailsTabs() {
+  const { t } = useTranslation();
+
+  const tabs: Tab[] = [
+    { name: t('details'), href: route('/settings/company_details/details') },
```
beganovich, line 19: This should be the default route. So `href` should be`/settings/company_details`. I believe at the moment it won't highlight the first tab cause of a non-existing route, right?

src/pages/settings/routes.tsx
```diff
@@ -17,9 +17,14 @@ import * as Settings from './index';
...
-    <Route path="company_details">
-      <Route path="" element={<Settings.CompanyDetails />} />
+    <Route path="company_details" element={<Settings.CompanyDetails />}>
+      <Route path="" element={<Settings.Details />} />
+      <Route path="details" element={<Settings.Details />} />
+      <Route path="address" element={<Settings.Address />} />
+      <Route path="logo" element={<Settings.Logo />} />
+      <Route path="defaults" element={<Settings.Defaults />} />
       <Route path="documents" element={<Settings.CompanyDocuments />} />
+      <Route path="custom_fields" />
```
beganovich, line 27: @turbo124 do we need custom fields tab here? It wasn't there before.

beganovich (changes requested):
Looks great. Minor changes needed :+1:

src/pages/settings/company/common/hooks/useCompanyDetailsTabs.tsx
Civolilah: > 

Yeah sure, I missed that, now it's changed.

src/pages/settings/routes.tsx
Civolilah: > @turbo124 do we need custom fields tab here? It wasn't there before.

David mentioned me on the Jira that he wants to add this tab same as on the Flutter app. So, I just prepared everything for doing this piece of task. I will ping him to give me more instructions for this tab.

turbo124: @beganovich I can't see where in the app we show company custom fields currently, so i advised @Civolilah to put them here

PR #270, opened 2022-11-19
[Feature] Disable email template

Civolilah:
@beganovich I've been looking for a solution to disable the MDEditor component from the @uiw/react-md-editor npm package, but I haven't found one. Here's a link to the source where I found it's not yet possible for this component: https://github.com/jrm2k6/react-markdown-editor/issues/56. Please let me know if I'm wrong.

@turbo124 Regarding this situation that I don't have permission to disable this component, I decided not to show this component if the user is not allowed to do so. Please let me know if you agree with this kind of solution if Ben confirms that disabling is not possible.

Here is the screenshot of how it looks without body template:

image

### DIFF
--- a/src/pages/settings/templates-and-reminders/TemplatesAndReminders.tsx
+++ b/src/pages/settings/templates-and-reminders/TemplatesAndReminders.tsx
@@ -9,12 +9,13 @@
  */
 
 import { Card, Element } from '@invoiceninja/cards';
-import { InputField, SelectField } from '@invoiceninja/forms';
+import { Button, InputField, Link, SelectField } from '@invoiceninja/forms';
 import { freePlan } from 'common/guards/guards/free-plan';
 import { endpoint, isHosted, isSelfHosted } from 'common/helpers';
 import { generateEmailPreview } from 'common/helpers/emails/generate-email-preview';
 import { request } from 'common/helpers/request';
 import { EmailTemplate } from 'common/hooks/emails/useResolveTemplate';
+import { useCurrentUser } from 'common/hooks/useCurrentUser';
 import { useInjectCompanyChanges } from 'common/hooks/useInjectCompanyChanges';
 import { useTitle } from 'common/hooks/useTitle';
 import { Settings as CompanySettings } from 'common/interfaces/company.interface';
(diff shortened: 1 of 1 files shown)


beganovich:
I don't think we should just hide input. My suggestion would be to replace the markdown component with `InputField` textarea and disable that. 

Nonetheless, I think this needs to be update with the CTA button as we discussed in Slack.

Civolilah:
@beganovich @turbo124 here is the screenshot how it looks with replaced button.

changePlanEmailTemplate

beganovich:
@turbo124 thoughts?

turbo124:
This looks fine

src/pages/settings/templates-and-reminders/TemplatesAndReminders.tsx
```diff
@@ -195,14 +199,30 @@ export function TemplatesAndReminders() {
...
+                  (current) => current && { ...current, body: value }
+                )
+              }
+            />
+          ) : (
+            <div className="flex flex-col items-start">
+              <span className=" text-gray-500" style={{ fontSize: 12 }}>
+                Email body template can be changed on{' '}
+                <strong>Enterprise/Pro</strong> plan.
+              </span>
```
beganovich, line 216: This needs to be replaced with a translatable string.

src/pages/settings/templates-and-reminders/TemplatesAndReminders.tsx
```diff
@@ -195,14 +199,30 @@ export function TemplatesAndReminders() {
...
+              value={templateBody?.body}
+              onChange={(value) =>
+                setTemplateBody(
+                  (current) => current && { ...current, body: value }
+                )
+              }
+            />
+          ) : (
+            <div className="flex flex-col items-start">
+              <span className=" text-gray-500" style={{ fontSize: 12 }}>
```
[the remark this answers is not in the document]
Civolilah: > * Remove blank space before `text-gray-500`
> * Use Tailwind representation for text size (probably you'll need `text-sm`.

Fixed

src/pages/settings/templates-and-reminders/TemplatesAndReminders.tsx
Civolilah: > This needs to be replaced with a translatable string.

Added

beganovich:
Thanks  :+1:

PR #279, opened 2022-11-23
[Feature] Payment Gateways Tabs

Civolilah:
Here is the screenshot of implemented cards for Payment Gateway section:

paymentGatewaysTabs


src/pages/settings/gateways/common/components/Tabs.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+export interface PaymentTabsActivity {
+  overview: boolean;
+  credentials: boolean;
+  settings: boolean;
+  required_fields: boolean;
+  limits_and_fees: boolean;
+}
+
+export function Tabs(props: Props) {
```
Civolilah, line 35: The reason for custom cards for Payment Gateways is the data sharing logic implemented on the Gatways component. The layout of the tabs is exactly the same, but the main difference is that the tabs are based on breadcrumbs. So when you change the card as well, the breadcrumb will change. I think our main Tabs component is not configurable in this case.

beganovich:
Few issues:

- Tabs are blank.

image

- This uses non-standard tab component, also for some reasons, tabs are centered.

src/pages/settings/gateways/common/components/Tabs.tsx
beganovich: Hi, 

I am not sure what you mean. Any chance you can record a quick demo explaining why do we need custom tabs for this?

Civolilah:
>

src/pages/settings/gateways/common/components/Tabs.tsx
Civolilah: > Hi,
> 
> I am not sure what you mean. Any chance you can record a quick demo explaining why do we need custom tabs for this?

Hi Ben, it's pretty simple, I'll use a Company Details cards as an example. There we have several tabs/components. All data in components is managed by Redux (handleChange/useSelector) and that's why we can navigate to a URL and render a component, the data will always be in Redux and available for display or use. Here we need to pass the selected gateway on the Overview tab through all components. If we navigate to a different URL, we will never know which Gateway is selected. The solution to using the already created Tabs component is to switch the logic to Redux. If that's okay for you, I can work on the switch. I'm pretty sure you'll understand what I'm talking about, but if you don't I'll be recording a demo tomorrow, I'm not in a position to record it right now.

image

beganovich:
I think this is fundamentally wrong. An idea of tabs is fine, but there's no need for deep linking on create page. We don't do "drafts" so record isn't created until you hit save. Nobody is expecting to navigate to create page onto specific tab externally.

We have `TabGroup` component in the system which uses transitions for the tabs instead of router outlet. That should do the job just fine.

Civolilah:
> I think this is fundamentally wrong. An idea of tabs is fine, but there's no need for deep linking on create page. We don't do "drafts" so record isn't created until you hit save. Nobody is expecting to navigate to create page onto specific tab externally.
> 
> We have `TabGroup` component in the system which uses transitions for the tabs instead of router outlet. That should do the job just fine.
> 
> Of course, for created record, we can use classic deep-linked style of tabs.

Done, using TabGroup component, and now the logic I have implemented is that if you have not selected any provider you will not be able to see any tab except `Provider`. When you select it, all other tabs will appear. This type of solution turns off all messages.

@turbo124 Please let me know if it is ok from your end. It was the same logic without the tabs, where no section appeared except the provider if we didn't have a provider selected.

turbo124:
@Civolilah 

This looks good.

One important point. I'm still not seeing you watch 4xx errors and displaying the error message back to the user. This is essential for every UI component built.

For exampe, when i attempt to create a payment gateway with no name, i get a generic something went wrong error. Please fetch the 422 error from the response and display this.

beganovich:
Marking as draft until validation part is done.

Civolilah:
> Marking as draft until validation part is done.

@turbo124 @beganovich Validation for all fields regarding Payment Gateway is finished. There are a few things important to mention. 

- Validation for all input fields is adjusted according to the 422 error returned by the API, but I tested almost all available fields and found no case of the API returning a 422, except for one that I will mention in the following text. Of course, the validation is also adjusted if the API does return it or with some change in the future , the validation will be displayed correctly.

- Validation for the provider/gateway name field is a slightly special case. If we don't select any provider, the object we send to the API will be undefined so the API will return a 500 error which is not configurable with our current way of displaying validation errors. So, I decided to use manual validation for this one field and if the object we send to the API (companyGateway object) is undefined, I will display a message like this using two keywords `invalid_data` as message and `invalid_gateway_key` as message under errors for `getway_key` property. Both are not in our resource files, if this is a good enough solution for you, just add them.

invalidGatewayKey

- Another thing to note is that the ``Forte'' gateway is currently unavailable for creation, and I'm not sure why. In this case we'll get a 422 validation error message from the API, but it's important to note that we're actually sending an object that's not undefined so the error isn't a 500 and that's why the message is displayed correctly. It looks like this: 

forteError

Please let me know about your thoughts about the solution.

turbo124:
@Civolilah 

This looks good. A couple of notes.

The forte validation error is actually correct! if the gateway is not defined in the DB, then this message should be returned.

I think if no gateway has been selected, we should disable the save button. Otherwise LTGM @beganovich

Civolilah:
> 

@beganovich @turbo124 I just checked and as I can see we don't have an option to disable the save button on the Settings component. What do you think Ben about adding this prop and turning off this part of the manual check in this case?

Civolilah:
> > 
> 
> @beganovich @turbo124 I just did disabling button if provider is not selected, so in my opinion it works properly!

beganovich:
Looks good, thanks! :+1:

PR #280, opened 2022-11-23
Update Activity Props

turbo124:
Extract VendorContacts
Update missing props from VendorContact
Update missing props from Activity
Update VendorContact form to include send_email prop
Update activity removing user hyperlink


beganovich (approved):
Looks great, thanks  :+1:

PR #286, opened 2022-11-24
[Feature] Transactions matching

Civolilah:
Here is a proposed solution for matching transactions. Here is the screenshot of what it looks like:

transactionMatch

Screenshot 2022-12-04 at 17 19 50

@turbo124 As we discussed in slack, I implemented this feature with a collapsible card from Transactions table. With this setup we missed some keywords in the resource file: `delete_transaction`, `archive_transaction`, `transaction_details`.


turbo124:
@Civolilah 

Matching should be done from the Transaction list page. It looks like it is nested in the settings here? I think the flutter app has the correct UI/UX here for this.

I would suggest refactoring to make it identical to flutter.

I'll make sure we have the correct translation string.

Invoices will always have a number, but vendor should only have the vendor name shown not the vendor number.

Civolilah:
> Here is a proposed solution for matching transactions. Here are two screenshots of what it looks like:

convertToExpense

> @beganovich We talked about filtering invoices/expenses/expense_categories, but I've checked the Flutter app multiple times and we're not hitting the API there every time the search term changes. So, I decided to show you my solution on frontend just like I did for Flutter. Tell me what you think about this proposed solution.
> 
> @turbo124 This feature has a lot of details, I hope I captured everything, but let me know if I missed anything. Also, you can notice in the second screenshot that the expense category search placeholder is incorrect. It refers to the keyword `"search_categories"`. Through the test application I noticed that we can create an invoice but without a number, so if someone associates a transaction with an invoice that does not have a number after matching on the details page it does not look good if I just leave it blank. I decided to put the keyword `"no_number"` in that place if we don't have a number for that invoice.

Civolilah:
> @Civolilah
> 
> Matching should be done from the Transaction list page. It looks like it is nested in the settings here? I think the flutter app has the correct UI/UX here for this.
> 
> I would suggest refactoring to make it identical to flutter.
> 
> I'll make sure we have the correct translation string.
> 
> Invoices will always have a number, but vendor should only have the vendor name shown not the vendor number.

@turbo124 Great catch, we did a first page PR for the transaction details page with the Settings component instead of Default and that's actually the difference where it sits. I just upgraded PR with some additional styles and tested to make sure it works. Please tell me your opinion.

Civolilah:
@turbo124 @beganovich Moved to draft until we adjust styling as we discussed on the Slack!

src/pages/settings/expense-categories/Create.tsx
```diff
@@ -56,6 +60,8 @@ export function Create() {
           toast.dismiss();
           toast.success(t('created_expense_category'));
```
Civolilah, line 61: @beganovich I did not change any part of the previously implemented logic for creating the expense category. We use the formik and the toaster directly from the package. Please let me know if I should include that part of the refactoring in this PR.

src/pages/settings/expense-categories/Edit.tsx
```diff
@@ -72,16 +72,17 @@ export function Edit() {
...
             ? setErrors(error.response?.data)
             : toast.error(t('error_title'));
         })
-        .finally(() => {
-          formik.setSubmitting(false);
-          queryClient.invalidateQueries(
-            route('/api/v1/expense_categories/:id', { id })
-          );
-        });
+        .finally(() => formik.setSubmitting(false));
```
Civolilah, line 85: @beganovich The same situation is here, please let me know if I have to include that part of refactoring.

src/pages/transactions/components/FilterModal.tsx
```diff
@@ -0,0 +1,87 @@
...
+            onValueChange={(value) =>
+              handleChangeSearchParams('maxAmount', value ? Number(value) : 0)
+            }
+          />
+        </div>
+        <div className="flex justify-evenly mt-3 w-full">
+          <div className="flex flex-col items-center w-1/3">
+            <p className="text-sm font-medium text-gray-500 mb-2">{`${t(
+              'start'
+            )} ${t('date')}`}</p>
```
beganovich, line 54: You are working with plain JSX here, no need to encapsulate twice.

```tsx
<p>{t('start')}</p>
```
is just fine.

src/pages/transactions/components/FilterModal.tsx
```diff
@@ -0,0 +1,87 @@
...
+                  'startDate',
+                  event.target.value ? event.target.value : ''
+                )
+              }
+            />
+          </div>
+          <div className="flex flex-col items-center w-1/3">
+            <p className="text-sm font-medium text-gray-500 mb-2">{`${t(
+              'end'
+            )} ${t('date')}`}</p>
```
beganovich, line 70: No need to encapsulate JSX.

src/pages/transactions/components/FilterModal.tsx
```diff
@@ -0,0 +1,87 @@
...
+              className="w-full border-gray-300"
+              type="date"
+              value={props.searchParams.startDate}
+              onChange={(event) =>
+                handleChangeSearchParams(
+                  'startDate',
+                  event.target.value ? event.target.value : ''
+                )
+              }
+            />
```
beganovich, line 65: Why can't we use `<InputField>` component here?

src/pages/transactions/components/FilterModal.tsx
```diff
@@ -0,0 +1,87 @@
...
+              className="w-full border-gray-300"
+              type="date"
+              value={props.searchParams.endDate}
+              onChange={(event) =>
+                handleChangeSearchParams(
+                  'endDate',
+                  event.target.value ? event.target.value : ''
+                )
+              }
+            />
```
beganovich, line 81: Why can't we use `<InputField>` component here?

src/pages/transactions/components/ListBox.tsx
```diff
@@ -0,0 +1,164 @@
...
+export function ListBox(props: Props) {
+  const [searchParams, setSearchParams] = useState<SearchInput>({
+    searchTerm: '',
+    minAmount: 0,
+    maxAmount: 0,
+    startDate: '',
+    endDate: '',
+  });
+
+  const [clientId, setClientId] = useState<string>('');
```
beganovich, line 55: `clientId` should be either string or null.

src/pages/transactions/components/ListBox.tsx
```diff
@@ -0,0 +1,164 @@
...
+    useState<boolean>(false);
+
+  const [isInvoicesDataKey, setIsInvoicesDataKey] = useState<boolean>(false);
+
+  const isItemChecked = (itemId: string) => {
+    return Boolean(props.selectedIds?.find((id) => id === itemId)?.length);
+  };
+
+  const selectItem = (itemId: string, clientId?: string) => {
+    setClientId(clientId || '');
```
beganovich, line 79: This it not good way of handling something "that doesn't exists". Please make sure to use nullabe clientId.

src/pages/transactions/components/ListBox.tsx
```diff
@@ -0,0 +1,164 @@
...
+          isFilterModalOpened={isFilterModalOpened}
+          setSearchParams={setSearchParams}
+          setSelectedIds={props.setSelectedIds}
+        />
+      </div>
+      <ul
+        className="flex flex-col justify-start overflow-y-auto"
+        style={{
+          height: isInvoicesDataKey ? 400 : 200,
+        }}
```
beganovich, line 146: Can we use Tailwind here?

src/pages/transactions/components/SearchArea.tsx
```diff
@@ -0,0 +1,105 @@
...
+  ) => {
+    props.setSearchParams((prevState) => ({
+      ...prevState,
+      [property]: value,
+    }));
+  };
+
+  return (
+    <>
+      <CreateExpenseCategoryModal
```
beganovich, line 54: This should probably be encapsulated within expense category selector.

src/pages/transactions/components/SearchArea.tsx
```diff
@@ -0,0 +1,105 @@
...
+        visible={isCreateExpenseCategoryModalOpen}
+        setVisible={setIsCreateExpenseCategoryModalOpen}
+        setSelectedIds={props.setSelectedIds}
+      />
+
+      <CreateVendorModal
+        visible={isCreateVendorModalOpen}
+        setVisible={setIsCreateVendorModalOpen}
+        setSelectedIds={props.setSelectedIds}
+      />
```
beganovich, line 64: This should be encapsulated into vendors selector.

src/pages/transactions/components/TransactionMatchDetails.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+  const [isFormBusy, setIsFormBusy] = useState<boolean>(false);
+
+  const [invoiceIds, setInvoiceIds] = useState<string[]>();
+
+  const [vendorIds, setVendorIds] = useState<string[]>();
+
+  const [expenseCategoryIds, setExpenseCategoryIds] = useState<string[]>();
+
+  const convertToPayment = (event: FormEvent<HTMLFormElement>) => {
```
beganovich, line 54: We rarely wanna expose direct events. Any chance you can make this function accepts the data instead of global event?

src/pages/transactions/components/TransactionMatchDetails.tsx
```diff
@@ -0,0 +1,191 @@
...
+        })
+        .catch((error: AxiosError) => {
+          console.error(error);
+          toast.error();
+        })
+        .finally(() => setIsFormBusy(false));
+    }
+  };
+
+  const convertToExpense = (event: FormEvent<HTMLFormElement>) => {
```
beganovich, line 91: Same as previous one.

src/pages/transactions/components/TransactionMatchDetails.tsx
```diff
@@ -0,0 +1,191 @@
...
+    if (props.transactionDetails.status_id !== TransactionStatus.Converted) {
+      props.setActionButton(
+        <ConvertButton
+          isFormBusy={isFormBusy}
+          isCreditTransactionType={props.isCreditTransactionType}
+          onClick={
+            props.isCreditTransactionType ? convertToPayment : convertToExpense
+          }
+        />
+      );
```
beganovich, line 145: This could be an anti-pattern. Can you explain why we are carrying whole JSX element to the props setter?

src/pages/transactions/routes.tsx
```diff
@@ -52,15 +51,6 @@ export const transactionRoutes = (
...
     />
-    <Route
-      path=":id"
-      element={
-        <Guard
-          guards={[() => permission('view_transaction')]}
-          component={<Transaction />}
-        />
-      }
-    />
```
beganovich, line 63: I am not sure we wanna nuke whole page. @turbo124 do we want this? We are deleting page to view single transaction and making it visible only if we click on the table row (slideover).

beganovich (changes requested):
Looks good, few changes needed.

src/common/interfaces/transactions.ts
```diff
@@ -53,9 +53,3 @@ export interface TransactionResponse {
   update_at: number;
   vendor_id: string;
 }
-
-export interface TransactionDetails {
-  amount?: number;
-  date?: string;
-  currency_id?: string;
-}
```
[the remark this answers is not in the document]
Civolilah: Actually, I realized that I don't need one more interface if I can you one that you actually made. So, I'm using only Transaction interface that you made earlier.

src/components/cards/CollapseCard.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { Dialog, Transition } from '@headlessui/react';
+import CommonProps from 'common/interfaces/common-props.interface';
+import { MdClose } from 'react-icons/md';
+import classNames from 'classnames';
+
+interface Props extends CommonProps {
+  visible: boolean;
+  setVisible: Dispatch<SetStateAction<boolean>>;
+  title?: string;
+  actionChildren?: ReactNode;
```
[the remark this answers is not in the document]
Civolilah: Yes, I understand your point, but on the example of this component that they kept a place reserved for actions only. So I got the idea to make this option, you can pass a button or any ReactNode element, but of course if it's not good for some solution you can only make it in normal children. I just wanted to make more options. In my opinion in this case it seems much better to pass this element which is at the end of the component. I think in some cases where you can immediately pass the button to this component is a very good option. Additionally, we don't need to worry about centering or any styling of the action elements, it's at the end of the component by default.

src/components/cards/CollapseCard.tsx
```diff
@@ -0,0 +1,93 @@
...
+                      </div>
+                      <div className="flex flex-1 flex-col justify-between items-center">
+                        <div className="divide-y divide-gray-200 w-full">
+                          {props.children}
+                        </div>
+                      </div>
+                    </div>
+                    {props.actionChildren && (
+                      <div className="flex justify-center px-4 py-4">
+                        {props.actionChildren}
```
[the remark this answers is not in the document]
Civolilah: I explained this part in more detail in a previous comment, but the idea is really just the children that are associated with the actions at the bottom of the slider.

src/pages/transactions/common/hooks/useTransactionColumns.tsx
```diff
@@ -68,13 +67,6 @@ export function useTransactionColumns() {
     {
       id: 'description',
       label: t('description'),
-      format: (value, transaction) => {
-        return (
-          <Link to={route('/transactions/:id', { id: transaction.id })}>
-            {value}
-          </Link>
-        );
-      },
```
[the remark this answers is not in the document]
Civolilah: In this implementation, all the details we had on the Details page are in the slider. So we don't have the route details anymore, so it's been removed. If we have a description, we'll just display the plain text in the column.

src/common/queries/expense-categories.ts
```diff
@@ -14,20 +14,29 @@ import { request } from 'common/helpers/request';
...
-      endpoint(
-        '/api/v1/expense_categories?per_page=:perPage&page=:currentPage&sort=:sort',
-        {
-          perPage: params.perPage,
-          currentPage: params.currentPage,
-          sort: params.sort ?? 'id|asc',
-        }
-      )
-    )
+export function useExpenseCategoriesQuery(params?: Params) {
```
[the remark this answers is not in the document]
Civolilah: changed

src/common/queries/vendor.ts
```diff
@@ -35,3 +37,20 @@ export function useBlankVendorQuery() {
     { staleTime: Infinity }
   );
 }
+
+export function useVendorsQuery(params?: Params) {
```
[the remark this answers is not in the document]
Civolilah: removed

src/components/DataTable.tsx
```diff
@@ -259,7 +262,11 @@ export function DataTable(props: Props) {
             data?.data?.data?.map((resource: any, index: number) => (
               <Tr
                 key={index}
-                onClick={() => document.getElementById(resource.id)?.click()}
+                onClick={() =>
+                  props.setResourceId
+                    ? props.setResourceId(resource.id)
+                    : document.getElementById(resource.id)?.click()
+                }
```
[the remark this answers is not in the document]
Civolilah: done

src/components/cards/CollapseCard.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+interface Props extends CommonProps {
+  visible: boolean;
+  setVisible: Dispatch<SetStateAction<boolean>>;
+  title?: string;
+  actionChildren?: ReactNode;
+  size: 'extraSmall' | 'small' | 'regular' | 'large' | 'extraLarge';
+}
+
+export function CollapseCard(props: Props) {
```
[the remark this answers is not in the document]
Civolilah: Changed to Slider, is it okay?

src/pages/invoices/common/queries.ts
```diff
@@ -0,0 +1,45 @@
...
+import { request } from 'common/helpers/request';
+import { GenericSingleResourceResponse } from 'common/interfaces/generic-api-response';
+import { Invoice } from 'common/interfaces/invoice';
+import { Params } from 'common/queries/common/params.interface';
+import { useQuery } from 'react-query';
+
+interface InvoiceParams extends Params {
+  client_status?: string;
+  client_id?: string;
+  without_deleted_clients?: boolean;
```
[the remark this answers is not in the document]
Civolilah: changed

src/pages/invoices/common/queries.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+export function useInvoicesQuery(params?: InvoiceParams) {
```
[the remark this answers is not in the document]
Civolilah: changed

src/pages/settings/expense-categories/Edit.tsx
```diff
@@ -72,16 +72,17 @@ export function Edit() {
         .catch((error: AxiosError) => {
           toast.dismiss();
 
+          queryClient.invalidateQueries('/api/v1/expense_categories');
+
+          queryClient.invalidateQueries(
+            route('/api/v1/expense_categories/:id', { id })
+          );
```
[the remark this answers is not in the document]
Civolilah: No, it's not enough. I've tested it but the data won't be fetched again if I turn off invalidate query for endpoint with id. In my opinion this is fine if it works like this because actually ``/api/v1/expense_categories'' and ``/api/v1/expense_categories/:id'' are two different endpoints.

src/pages/settings/expense-categories/Edit.tsx
Civolilah: done

src/pages/transactions/common/queries.ts
```diff
@@ -28,3 +31,21 @@ export function useTransactionQuery(params: { id: string | undefined }) {
...
+      client_status: 'unpaid',
+      filter,
+      client_id,
+    });
+  } else if (dataKey === 'vendors') {
+    return useVendorsQuery({ filter });
+  }
+
+  return useExpenseCategoriesQuery({ filter });
+}
```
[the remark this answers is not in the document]
Civolilah: Great to know, thanks! Then, I changed it and move all queries to ListBox component without conditions.

src/pages/transactions/components/ConvertButton.tsx
```diff
@@ -0,0 +1,39 @@
...
+    >
+      {<MdContentCopy fontSize={22} />}
+      <span>
+        {props.isCreditTransactionType
+          ? t('convert_to_payment')
+          : t('convert_to_expense')}
+      </span>
+    </Button>
+  );
+}
```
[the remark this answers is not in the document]
Civolilah: Yeah, actually that is just one button with some specfic details button and we don't need to create component because of that. So, this component has been removed.

src/pages/transactions/components/CreateExpenseCategoryModal.tsx
```diff
@@ -0,0 +1,120 @@
+/**
```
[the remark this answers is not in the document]
Civolilah: done

src/pages/transactions/components/CreateVendorModal.tsx
```diff
@@ -0,0 +1,119 @@
+/**
```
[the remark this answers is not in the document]
Civolilah: done

src/pages/transactions/components/FilterModal.tsx
Civolilah: changed

src/pages/transactions/components/FilterModal.tsx
Civolilah: I was not able to resize them at that moment, but now I tried with one more solution and resolve it. So, now we use InputField! Changed!

src/pages/transactions/components/ListBox.tsx
```diff
@@ -0,0 +1,164 @@
...
+import { SearchArea } from './SearchArea';
+import { useTransactionMatchResourceQuery } from '../common/queries';
+import { ListBoxItem } from './ListBoxItem';
+
+export interface ResourceItem {
+  id: string;
+  name: string;
+  number: number;
+  clientName: string;
+  client_id: string;
```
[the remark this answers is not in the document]
Civolilah: sure, changed

src/pages/transactions/components/ListBox.tsx
```diff
@@ -0,0 +1,164 @@
...
+  maxAmount: number;
+  startDate: string;
+  endDate: string;
+}
+
+interface Props extends CommonProps {
+  transactionDetails: TransactionDetails;
+  dataKey: 'invoices' | 'categories' | 'vendors';
+  setSelectedIds: Dispatch<SetStateAction<string[] | undefined>>;
+  selectedIds: string[] | undefined;
```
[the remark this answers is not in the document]
Civolilah: Sure, changed

src/pages/transactions/components/ListBox.tsx
Civolilah: changed

src/pages/transactions/components/ListBox.tsx
Civolilah: Actually, they use the same way of handling this lists with ul element. https://tailwindui.com/components/application-ui/lists/stacked-lists

In our case I must to limit the height because of the screen size and behaviour on the screen if we get overflow.

src/pages/transactions/components/TransactionMatchDetails.tsx
Civolilah: Actually the data is within the component so we don't have to pass it `const [invoiceIds, setInvoiceIds] = useState<string[]>();` and then I use it later within in the function: `invoiceIds.join(','),`. Event here is just to prevent default.

src/pages/transactions/components/TransactionMatchDetails.tsx
Civolilah: I replied above!

src/pages/transactions/components/TransactionMatchDetails.tsx
Civolilah: The logic has been changed, now we use the button in this component so we don't have to pass it to another one.

src/pages/transactions/routes.tsx
Civolilah: Here's minor work to return previous logic. But it was obvious to me not to duplicate places for the same data, so I removed the route and kept it only on the Slider.

src/pages/transactions/components/Details.tsx
```diff
@@ -8,31 +8,215 @@
...
+    id: transaction?.vendor_id || '',
+  });
+
+  const { data: expenseResponse } = useExpenseQuery({
+    id: transaction?.expense_id || '',
+  });
+
+  const { data: expenseCategoryResponse } = useExpenseCategoryQuery({
+    id: transaction?.ninja_category_id || '',
+  });
```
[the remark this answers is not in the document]
Civolilah: done

src/pages/transactions/components/SearchArea.tsx
Civolilah: done

src/pages/transactions/components/SearchArea.tsx
Civolilah: done

turbo124:
@Civolilah 

Looking really good.

One small thing.

Can we please disable the Convert to Expense / Convert to payment button until the user has select an invoice OR a vendor/category please.

At the moment there is no feedback, so disabling the button until we are in a valid state would be the best UX.

Civolilah:
> @Civolilah
> 
> Looking really good.
> 
> One small thing.
> 
> Can we please disable the Convert to Expense / Convert to payment button until the user has select an invoice OR a vendor/category please.
> 
> At the moment there is no feedback, so disabling the button until we are in a valid state would be the best UX.

@turbo124 

Thanks!

Yes, of course, you are right. It's just implemented. So, if there are no selected invoices if we are talking about the CREDIT type or if there is no selected vendors or expense categories if we are talking about the DEPOSIT type, button will be disabled!

turbo124:
@Civolilah 👍

turbo124:
@Civolilah 

Small things, please replace translation streing transaction_details with the description of the transaction

image

We may want to concatenate to ensure no overflow

For the search box for expense categories, please pass a blank string for :count, this will remove that particular place holder :)

beganovich:
Marking this as draft, since Dave requested further changes.

Civolilah:
> @Civolilah
> 
> Small things, please replace translation streing transaction_details with the description of the transaction
> 
> image
> 
> We may want to concatenate to ensure no overflow
> 
> For the search box for expense categories, please pass a blank string for :count, this will remove that particular place holder :)

Sure, it's implemented! This is how it looks when we enter the description (the full description will be displayed if it is shorter than 35 characters, if it is more than that we will only display `...` as the end of the title):

Screenshot 2022-12-07 at 13 21 05

When we don't have any description the date will be displayed instead:

Screenshot 2022-12-07 at 13 16 36

Also, for placeholder, I'm not sure that I'm able to remove `:count`, because I just use keyword `search_categories` for string in the placeholder. I think that we have to correct this keyword to fix this. Of course, please let me know, if I miss something.

beganovich:
Looks good, I think this is last round of changes. Following needed, from UX/UI perspective.

1. Slideover animation is **too slow**. Can you adjust it to be same as modals and dropdown?
2. Top bar of the slideover look *disabled*. I don't think this gray fits well. My suggestion would be to add `bg-white border-b` as this aligns with rest of components.
3. "Search invoices" also looks disabled. Is it possible to use standard input field component? If not, try to copy styles from it, at very least.
4. Action button / action button container doesn't have proper transparency configuration. See the screenshot.
5. Make sure to merge latest `develop` as there were some big changes (upgrade to React 18) and fix merge conflicts.

image

I think after these changes it should be fine to merge. @turbo124 if you have time, glance over this once again to make sure everything works properly.

turbo124:
@Civolilah 

Looks good :+1: 

One small thing. If i add multiple categories from transaction matching page. the list does not update and the first category remains selected. We'll want to ensure the last created category is set as checked

turbo124:
@beganovich this LGTM now :+1:

Civolilah:
> 

@beganovich All things that you mentioned are adjusted.

1. Animation is the same as on modal
2. The top part of the slider has bg white and the border bottom
3. Removed styling so it now looks like a user-usable field, but not a disabled field
4. I got the idea to fit the content better so this problem was removed after that, the button will always be at the bottom, so for this the opacity doesn't matter
5. Any changes you have merged evolve from your branch are also merged into this one
6. Yes, you are right, I had to link it, so now we have the color change on hover, but also the links.

@beganovich @turbo124 I know David replied to the comment that everything looks good for functionality, but just to confirm again that I can't find a single little bug on my end. We talked about it on Slack.

beganovich:
Looking good, but you got the animation part wrong. Animation shouldn't be the same as modals/dropdowns, just the *speed of animation*. We still want to keep the "slide" animation.

Civolilah:
> Looking good, but you got the animation part wrong. Animation itself shouldn't be the same as modals/dropdowns (fade), just the _speed of animation_. We still want to keep the "slide" animation.

Oh, sorry, I thought you wanted the same behavior as for the modal. Now it is adjusted and duration of `enter` is set to 300 milliseconds and `leave` is set to 200 milliseconds same for modals.

beganovich:
Looks great, thanks 👍

Issue #288, opened 2022-11-27
Company documents messed up


hismayilov:
For multiple companies, it shows unrelated documents.

![image](https://user-images.githubusercontent.com/676555/204161869-9766765f-43ea-446a-b2a8-e8a6aed14e94.png)

turbo124:
This is a bug, we are checking in a fix for this, both in the UI and also the API, so the next release will have this issue patched.

beganovich:
Fixed in #289. Thanks for reporting it!

PR #289, opened 2022-11-27
Patch turbo124 company documents

turbo124:
Filters for company documents #288

### DIFF
--- a/src/common/queries/common/params.interface.ts
+++ b/src/common/queries/common/params.interface.ts
@@ -14,5 +14,5 @@ export interface Params {
   filter?: string;
   status?: string[];
   sort?: string;
-  companyDocuments?: string;
+  companyDocuments?: 'true' | 'false';
 }
--- a/src/common/queries/common/params.interface.ts
+++ b/src/common/queries/common/params.interface.ts
@@ -14,4 +14,5 @@ export interface Params {
   filter?: string;
   status?: string[];
   sort?: string;
+  companyDocuments?: string;
 }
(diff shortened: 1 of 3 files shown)


src/pages/settings/company/documents/components/Table.tsx
```diff
@@ -37,11 +37,12 @@ export function Table() {
   const [t] = useTranslation();
   const [currentPage, setCurrentPage] = useState<number>(1);
   const [perPage, setPerPage] = useState<string>('10');
+  const companyDocuments = 'true';
```
beganovich, line 40: Does this need to be external variable? From what I can see, it's never changing?

src/common/queries/common/params.interface.ts
```diff
@@ -14,4 +14,5 @@ export interface Params {
   filter?: string;
   status?: string[];
   sort?: string;
+  companyDocuments?: string;
```
beganovich, line 17: This can be improved. You can pass `"true" | "false"` or `boolean`, but don't forget to stringify it later.

src/pages/settings/company/documents/components/Table.tsx
```diff
@@ -37,11 +37,12 @@ export function Table() {
...
   const { dateFormat } = useCurrentCompanyDateFormats();
   const [document, setDocument] = useState('');
   const [isPasswordConfirmModalOpen, setPasswordConfirmModalOpen] =
     useState(false);
-  const { data, isLoading } = useDocumentsQuery({ perPage, currentPage });
+  const { data, isLoading } = useDocumentsQuery({ perPage, currentPage, companyDocuments });
```
beganovich, line 45: Since the variable never changes, just pass the direct value:

```ts
const { data, isLoading } = useDocumentsQuery({ perPage, currentPage, companyDocuments: 'false' });
```

beganovich (changes requested):
Looks good, minor changes needed :+1:

beganovich:
Thank you so much :+1:

PR #292, opened 2022-12-04
[Feature] Hide sensitive data

Civolilah:
Here is an implemented solution for hiding sensitive data on payment gateway cards and the ability to display them if the user wants when the field type is password!

Screenshot 2022-12-04 at 16 36 53

The solution also applies to all password type fields such as the login/registration form!

Screenshot 2022-12-04 at 16 45 12

@beganovich  I've been trying to find if the DebounceInput component contains a property to add some icons to the field, but I haven't found anything in the documentation and other resources. So I think this component does not contain that and implemented this solution. Some MUI TextField components contain properties like end adornment that we can use to implement this functionality, but we're not using MUI so this isn't available.

@turbo124 This workaround is implemented on all input fields on the Credential tab (the other tabs do not contain any sensitive data) that contain a `key`, `password`, `secret`, or `id` in the property name.

### DIFF
--- a/src/components/forms/InputField.tsx
+++ b/src/components/forms/InputField.tsx
@@ -84,7 +84,7 @@ export function InputField(props: Props) {
         />
 
         {isInitialTypePassword && (
-          <span className="absolute top-1/4 right-5 cursor-pointer">
+          <span className="absolute top-1/4 right-3 cursor-pointer">
             {isEyeOpen ? (
               <AiFillEye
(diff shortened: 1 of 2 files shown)


turbo124:
@beganovich LGTM :+1:

beganovich:
I memoized the input type value and extracted it into computed property.

Otherwise, looks good. Thanks :+1:

PR #299, opened 2022-12-08
Restrict payment updates

turbo124:
PUT route should only update meta data


src/pages/payments/edit/Edit.tsx
```diff
@@ -72,7 +72,7 @@ export function Edit() {
       }}
     >
       {payment?.client && <ClientCard client={payment.client} />}
-      {payment && <PaymentOverview payment={payment} />}
+      {payment && <PaymentOverview payment={data?.data.data} />}
```
beganovich, line 75: If we are pushing empty list of invoices/credits into `payment` object, why don't we use it here? What's purpose of the `payment` then?

beganovich (commented):
Looks good, just to confirm the changes are the ones we wanted.

src/pages/payments/edit/Edit.tsx
turbo124: Payment cannot have invoices or credit array in request.

However we need invoices array to display existing invoices in the payment

beganovich: That's fine, yes. But we are making `payment` variable essentially `data.data.data` just with empty invoices and then completely ignoring that variable? Why use it at all? 

Why was this change from `payment` to `data.data.data`?

beganovich:
Looks good, thanks :+1:

PR #300, opened 2022-12-09
[Feature] Localization tabs

Civolilah:
Here is the screenshot of implemented solution for localization tabs:

Screenshot 2022-12-09 at 17 15 19


beganovich (approved):
Looks great, thanks 👍

PR #301, opened 2022-12-10
[Feature] Invoices additional filter

Civolilah:
Here is the screenshot of implemented solution for additional filters for invoices:

Screenshot 2022-12-10 at 01 55 46

@turbo124 I included in the dropdown menu values that you mentioned in the description on the Jira, `All`, `Paid`, `Unpaid` and `Overdue`. Please let me know if I have to include other filters from Flutter app.


turbo124:
@Civolilah 

this list of filters is fine for now.

src/components/DataTable.tsx
```diff
@@ -77,6 +79,7 @@ export function DataTable(props: Props) {
   const queryClient = useQueryClient();
 
   const [filter, setFilter] = useState('');
+  const [customFilter, setCustomFilter] = useState<string[]>(['']);
```
beganovich, line 82: Does this need to have a default value of one empty string? Why can't we provide it with just `[]`?

src/components/datatables/Actions.tsx
```diff
@@ -14,6 +14,10 @@ import { InputField } from '../forms/InputField';
 import Select, { MultiValue, SingleValue, StylesConfig } from 'react-select';
 import { ReactNode, ChangeEvent } from 'react';
 
+interface Placeholders {
+  client_status: string;
```
beganovich, line 18: This is not following the conventions of the project. Make sure this is `clientStatus`.

src/components/datatables/Actions.tsx
```diff
@@ -29,24 +33,42 @@ interface Props extends CommonProps {
   rightSide?: ReactNode;
   onFilterChange?: any;
   onStatusChange?: any;
+  onCustomFilterChange?: any;
```
beganovich, line 36: Can we typehint this? `any` isn't actually helpful, and I have plans to implement no implicit any rule so this will throw an error in the future.

src/components/datatables/Actions.tsx
```diff
@@ -29,24 +33,42 @@ interface Props extends CommonProps {
...
-  function onStatusChange(
+  const onStatusChange = (
     options:
       | MultiValue<{ value: string; label: string }>
       | SingleValue<{ value: string; label: string }>
-  ) {
+  ) => {
     const values: string[] = [];
 
     (options as any).map((option: { value: string; label: string }) =>
```
beganovich, line 51: Can we possibly provide types for this?

src/components/datatables/Actions.tsx
```diff
@@ -29,24 +33,42 @@ interface Props extends CommonProps {
...
+  };
+
+  const onCustomFilterChange = (
+    options:
+      | MultiValue<{ value: string; label: string }>
+      | SingleValue<{ value: string; label: string }>
+  ) => {
+    const values: string[] = [];
+
+    (options as any).map((option: { value: string; label: string }) =>
```
beganovich, line 65: Same with this.

src/components/datatables/Actions.tsx
```diff
@@ -70,6 +92,10 @@ export function Actions(props: Props) {
     }),
   };
 
+  const customFiltersPlaceholders: Placeholders = {
+    client_status: t('invoice_status'),
```
beganovich, line 96: I assume this will be fixed once interface is changed, so make sure to have this covered as well.

src/pages/invoices/index/Invoices.tsx
```diff
@@ -12,39 +12,28 @@ import { useTitle } from 'common/hooks/useTitle';
...
-          className="w-4 h-4 mr-2"
-          xmlns="http://www.w3.org/2000/svg"
-          viewBox="3 3 20 20"
-        >
-          <Download />
-        </svg>
-        <span>{t('import')}</span>
-      </button>
-    </ReactRouterLink>
-  );
```
beganovich, line 44: I don't know how I feel about this. I think it's overengineering for such a simple button to have the additional file. Do we need to reuse this button? If not, please keep it as is.

Also, I noticed you used the same pattern on the transactions page, so please make sure this is aligned with the rest of the project.

beganovich (changes requested):
Looks good, minor changes needed.

src/components/DataTable.tsx
Civolilah: Sure, removed!

src/components/datatables/Actions.tsx
Civolilah: The logic of getting the placeholder string was via `query_key` and that's why we had `client_status` out of convention. There's no way if I'm right to do that with `clientStatus` because the actual query_key is client_status. So I changed the logic of this part a bit and now I think this part is conventional.

src/components/datatables/Actions.tsx
Civolilah: Yeah, all types are adjusted now, including type for `onFilterChange` prop.

src/components/datatables/Actions.tsx
Civolilah: Yeah, adjusted!

src/components/datatables/Actions.tsx
Civolilah: Yes, adjusted!

src/pages/invoices/index/Invoices.tsx
Civolilah: I agree with you that this is a very simple button, but if we're just changing the route every time I think that's a good reason to globalize it. Also, if we include using the same button in 5 or more places, I think it's good for globalization. I implemented this component in every place where I had some functions or refactoring. So I'll go ahead with that, or if you want, maybe I can do a separate PR to implement this button everywhere.

beganovich (commented):

beganovich:
This is completely not needed. 
image


Can we just pass the filter string as DataTable prop, otherwise just default to "Status"?

Civolilah:
> This is completely not needed. image
> 
> Can we just pass the filter string as DataTable prop, otherwise just default to "Status"?

@beganovich Of course, actually, at the beginning I did it with a prop, and then I just wanted to avoid props, so I did it this way. Also, I decided to use prop instead of default value, since we will have custom filters for almost all other tables, in which case I think it's better to pass placeholder.

beganovich (approved):
Looks good, thanks :+1:

PR #302, opened 2022-12-10
[Bugfix] Highlighting settings menu items

Civolilah:
Here is the screenshot of resolved bug when we switch tabs of menu items in the settings:

Screenshot 2022-12-10 at 15 31 29


src/components/layouts/Settings.tsx
```diff
@@ -39,102 +39,107 @@ interface AdvanceSetting {
 
 export function Settings(props: Props) {
   const [t] = useTranslation();
+
   const location = useLocation();
 
+  const doesPathNameIncludes = (route: string) => {
+    return location.pathname.includes(route);
+  };
```
beganovich, line 47: I don't think we need a separate function for this. See comment below.

src/components/layouts/Settings.tsx
```diff
@@ -39,102 +39,107 @@ interface AdvanceSetting {
...
+
   const basic = [
     {
       name: t('company_details'),
       href: '/settings/company_details',
-      current: location.pathname === '/settings/company_details',
+      current: doesPathNameIncludes('/settings/company_details'),
```
beganovich, line 53: I think this would work way better if it was checking for start of string.

```ts
location.pathname.startsWith('/settings/company_details')
```

beganovich (changes requested):
Looks good, minor changes are needed.

src/components/layouts/Settings.tsx
Civolilah: changed

src/components/layouts/Settings.tsx
Civolilah: removed

beganovich (approved):
Looks good, thanks :+1:

PR #304, opened 2022-12-11
[Feature] Company import

Civolilah:
Here is the screenshot of prepared/finished solution for company import:

Screenshot 2022-12-11 at 14 57 40

@turbo124 I've tried some combinations for the import company, but I'm not sure what the CSV file should look like for the import company (company object really has a lot of properties). Please just give me an example to test, or if you have some free time, test it and let me know if it works or not, so we can move the PR to be ready for review. Also, if this UI/UX is okay, we are missing `import_company` keyword.


turbo124:
@Civolilah 

The company backup is stored in a .zip file. For company imports, you can place some validation on the file being .zip

image

PR #305, opened 2022-12-11
[Cleanup] Tax rate formatting

Civolilah:
Here is the screenshot of reafactored table and columns for tax rates:

Screenshot 2022-12-11 at 15 37 36

@turbo124 @beganovich Also, I noticed that this table is not consistent with other tables and through testing I didn't notice any difference or specific case why we don't use the same table as for other features, so I refactored it. If there is no reason why we used a custom table, there are two new keywords to add, `delete_tax_rate` and `restore_tax_rate`. One small thing about keywords, as you can see in the screenshot `edit_tax_rate` the keyword has a value of `Edit tax rate` so I recommend that you change all the words to uppercase as well as the other tax rate related keywords.

One more thing about the UI for tax rates. On the edit page we have archive and delete functionality, if we have those functionalities in the more-actions dropdown menu in the table, I recommend removing it from the edit page. Here's a screenshot:

Screenshot 2022-12-11 at 15 38 16

### DIFF
--- a/src/pages/settings/tax-rates/TaxRates.tsx
+++ b/src/pages/settings/tax-rates/TaxRates.tsx
@@ -7,112 +7,21 @@
  *
  * @license https://www.elastic.co/licensing/elastic-license
  */
-import { Button, Link } from '@invoiceninja/forms';
-import {
-  Pagination,
-  Table as TableComponent,
-  Tbody,
-  Td,
-  Th,
-  Thead,
-  Tr,
-} from '@invoiceninja/tables';
-import { AxiosError } from 'axios';
-import { TaxRate } from 'common/interfaces/tax-rate';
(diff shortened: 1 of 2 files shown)


turbo124:
@Civolilah 

I'll add those translations in.

unifying the table sounds fine to me, however i'll default to @beganovich

Civolilah:
> Refactoring tables to `<DataTable>` component is good. This page was created before we had that component so it makes sense that is out of date.
> 
> As for the cruddy actions, please keep them. A possibly nice refactor would be dropdown as it is on the products edit page.

@beganovich As we discussed on the Slack, changes about dropdown menu will be done on some other PR.

beganovich:
Looks good, thanks :+1:

PR #308, opened 2022-12-13
[Feature] Language switching

Civolilah:
Here is the screenshot of the implemented solution for switching to available languages:

Screenshot 2022-12-13 at 16 22 13

Screenshot 2022-12-13 at 20 45 08

@turbo124 As an example I will take `Arabic` and `Polish`, and as you can, we are missing some keywords for some languages ​​like `task_settings`... Just to let you know that we have to fix it. Also, I tested all the available languages ​​from the dropdown menu, I noticed that we are missing resource files for `Slovak` and `Serbian`, all other languages ​​have resource files.


turbo124:
@Civolilah 

This looks good. In cases where there is no translation string, the system will default back to english.

Also to note, after a user has updated the language, we need to pull in the statics list again, this will contain the translated list of variables that we use in the application for dates / currencies  / countries etc etc that are localized. So you'll need to hit

```
/api/v1/statics
```

src/App.tsx
```diff
@@ -23,7 +32,22 @@ export function App() {
...
       : document.querySelector('html')?.classList.remove('dark');
-  }, [darkMode]);
+
+    const resolvedLanguage = resolveLanguage(company?.settings.language_id);
+
+    if (resolvedLanguage?.locale) {
+      if (!i18n.hasResourceBundle(resolvedLanguage.locale, 'translation')) {
+        import(
+          `./resources/lang/${resolvedLanguage.locale}/${resolvedLanguage.locale}.json`
+        ).then((response) => {
```
Civolilah, line 42: The reason for dynamically importing json files as translation resources is actually memory. If we create some index file and include them all at once, the index.tsx file will be as heavy as `10MB` and will fail in `npm run build` command because it is over `1500KB`. I did some research on Google and almost all people do it this way. You can check out the first commit on this branch to see how I implemented this using all the resource files immidiatelly.

src/pages/settings/common/hooks/useHandleCompanySave.tsx
```diff
@@ -34,29 +37,21 @@ export function useHandleCompanySave() {
...
+        Statics.reloadQuery();
+
+        toast.success('updated_settings');
       })
       .catch((error: AxiosError<ValidationBag>) => {
-        console.error(error);
-
-        toast.error(t('error_title'), { id: toastId });
+        if (error.response?.status === 422) {
+          toast.error(error.response.data.message);
```
Civolilah, line 48: As we discussed earlier on the Slack, this way of handling 422 errors is not good. So, I changed the package we used for toasting and removed this handling. In my opinion, this part of the refactoring is not related to this PR. We use this `useHandleCompanySave` hook on like 10 different components, so the refactoring will involve many more files. I recommend creating an issue on Jira for this refactoring and then we can end up with 422 errors for all components. But, let me know your thoughts.

Civolilah:
> @Civolilah
> 
> This looks good. In cases where there is no translation string, the system will default back to english.
> 
> Also to note, after a user has updated the language, we need to pull in the statics list again, this will contain the translated list of variables that we use in the application for dates / currencies / countries etc etc that are localized. So you'll need to hit
> 
> ```
> /api/v1/statics
> ```

@turbo124 Sure, good catch! This part is also finished. Please let me know if I have to implement anything additional.

src/pages/settings/common/hooks/useHandleCompanySave.tsx
beganovich: >  In my opinion, this part of the refactoring is not related to this PR.

Exactly. Please move this to another pull request.

beganovich (changes requested):

Civolilah:
> Looks good, minor change needed.

@beganovich I've reverted the old validation error handling system in the `useHandleCompanySave` hooks and created an issue on Jira so we can refactor it in one of the next PRs.

beganovich:
Marking this as draft as build fails critically with memory leaks.

Civolilah:
> Marking this as draft as build fails critically with memory leaks.

@beganovich It's fixed. I found a solution for dynamic import but also to adapt everything to our vita scripts. Sorry for making some extra commits, but I really had to put it on GitHub. When I tried to do the same command locally, I didn't get the same error.

beganovich (approved):
Looks good, thanks 👍

PR #310, opened 2022-12-15
[Feature] Transactions linking

Civolilah:
Here are screenshots of implemented solution for linking transactions to payment or expense regarding to the `base_type`:

Screenshot 2022-12-15 at 16 40 52

Screenshot 2022-12-15 at 16 41 09

@turbo124 In current implementation we are missing a few keywords in translation: `match_payment`, `link_to_payment`, `match_expense`, `link_to_expense`. As we discussed on the Slack, we just need to fix linking transaction to payment. Maybe, there is just some mapping bug on the API.


src/pages/transactions/components/TransactionMatchDetails.tsx
```diff
@@ -131,6 +175,66 @@ export function TransactionMatchDetails(props: Props) {
...
+        return convertToExpense;
+      }
+    } else {
+      if (props.isCreditTransactionType) {
+        return linkToPayment;
+      } else {
+        return linkToExpense;
+      }
+    }
+  };
```
beganovich, line 236: Is it possible to simplify this? Looks like a ton of nesting. Perhaps we can simplify it with early return?

Something like this should work:

```ts
const getMatchingFunction = () => {
  if (isCreate) {
    if (props.isCreditTransactionType) {
      return convertToPayment;
    }

    return convertToExpense;
  }

  if (props.isCreditTransactionType) {
    return linkToPayment;
  }

  return linkToExpense;
};
```

beganovich (changes requested):

beganovich:
Looking good, I wonder why do we reinvent tabs here? Can we just use regular tabs? (TabGroup)

image

src/pages/transactions/components/TransactionMatchDetails.tsx
Civolilah: Yeah, you are right. It's simplified.

Civolilah:
> Looking good, I wonder why do we reinvent tabs here? Can we just use regular tabs? (TabGroup)
> 
> image

@beganovich Yes, I understand you, but in my opinion the custom implementation for those two tabs looks much better for the UI in the slider, and also the `TabGroup` component will not be changed. I know it's better to have all tabs consistent, but I'm not sure TabGroup components provide a good looking tabs for this feature.

Two main reasons that I did not implement this type of tabs are:

- UI as I mentioned
- The list boxes for all the data we display in this slider are a bit specific (especially the height). If we use TabGroup, we won't be able to know which tab we are on (Create or Match), but this data is really needed for some parts of the logic. Additionally, we are unable to set the maximum available height for list boxes because we have some parent elements in the TabGroup component that do not allow us. Setting the maximum height is also very helpful for the user interface.

Actually, I customized those components to use TabGroup to see what we have to customize and how it looks, but it's not worth it in my opinion. But if it's much better for you, I agree to use TabGroup. But, I think the modified code is a bit difficult if we adapt the TabGroup component for this purpose. Please let me know your thoughts.

beganovich:
1. We are not trying to match Flutter UI-wise, we are trying to match it UX-wise. So not each element has to match with Flutter. We have totally different UI language here in React app compared to material UI.

2. I have no idea why you won't be able to do it. It worked just fine for me at #306 .

Civolilah:
> 1. We are not trying to match Flutter UI-wise, we are trying to match it UX-wise. So not each element has to match with Flutter. We have totally different UI language here in React app compared to material UI.
> 2. I have no idea why you won't be able to do it. It worked just fine for me at [Additional kanban features #306](https://github.com/invoiceninja/ui/pull/306) .

@beganovich Okay, understand you. I reused `TabGroup` component with additional adjusting within the component for height. I just saw your PR and as I see that you adjusted width, the same feature will be very helpful for me. In my case, I had to implement a very similar type of setup to achieve full height.

beganovich:
I think the whole point of matching the tabs the same way I did is to make it consistent in the app. I made PR that extracts these changes into the latest develop. #315.

Few more things:

1. Can we use this updated version of the TabGroup component? 
2. I noticed you have `onTabClick` implemented for tabs. IMO, tabs should only do the switching, unless this is absolutely required, I'd look for a workaround. Can you explain a bit of the context behind this?
3. It still looks a bit janky. These margins are weird. Can you glance at the tabs at #306 and how I did it there with a combination of Element and TabGroup?

image

This double border looks weird. The left and right margins are also looking weird. Then we also have a double border before the last input. 

If you are unsure how to deal with this, please let me know. Thanks! :)

Civolilah:
> I think the whole point of matching the tabs the same way I did is to make it consistent in the app. I made PR that extracts these changes into the latest develop. #315.
> 
> Few more things:
> 
> 1. Can we use this updated version of the TabGroup component?
> 2. I noticed you have `onTabClick` implemented for tabs. IMO, tabs should only do the switching, unless this is absolutely required, I'd look for a workaround. Can you explain a bit of the context behind this?
> 3. It still looks a bit janky. These margins are weird. Can you glance at the tabs at [Additional kanban features #306](https://github.com/invoiceninja/ui/pull/306) and how I did it there with a combination of Element and TabGroup?
> 
> image
> 
> This double border looks weird. The left and right margins are also looking weird. Then we also have a double border before the last input.
> 
> If you are unsure how to deal with this, please let me know. Thanks! 👍

@beganovich I agree with you. I implemented all things you mentioned:

1. Used last TabGroup version
2. I adjusted logic and it reflected to removing onTabClick function
3. I take a look into your implementation and adjusted it with this feature
4. Additional styling is also finished

Here is the screenshot of how it looks now:

Screenshot 2022-12-18 at 20 37 39

src/components/TabGroup.tsx
```diff
@@ -51,15 +53,32 @@ export function TabGroup(props: Props) {
...
+                  currentIndex === index
+                    ? `${props.height === 'full' ? 'flex' : 'block'}`
+                    : 'none',
+                ...(props.height === 'full' && {
+                  flexDirection: 'column',
+                }),
+                ...(props.height === 'full' && {
+                  flexGrow: 1,
+                }),
+              },
```
beganovich, line 81: Is it possible to style these with Tailwind? The classes you are looking for are: flex, block, flex-1, flex-column

beganovich (changes requested):
Looks good, small changes needed 👍

src/components/TabGroup.tsx
Civolilah: Sure, by passing the style in the className instead of the style property we can achieve this. So this is refactored!

beganovich:
Looks good, thanks 👍

PR #316, opened 2022-12-18
[Feature] Recurring expenses

Civolilah:
Here are screenshots of implemented feature [Recurring expenses]:

Screenshot 2022-12-20 at 15 08 41

Screenshot 2022-12-20 at 14 54 54

Screenshot 2022-12-20 at 14 56 42

@turbo124 Just a few things about translation files. We are missing two keywords associated with this feature: `started_recurring_expense` and `stopped_recurring_expense`. If you look at the third screenshot, `clone_to_expense` which gives us the result 'Clone to expense', I suggest we update it to 'Clone to Expense' to be consistent with the others. Also, for the number label on the Edit page I used `expense_number` instead of `recurring_expense_number`. IMO it was enough, but let me know what you think.


Civolilah:
> 

@turbo124 I changed the translation keywords for the cases you mentioned. But, IMO, maybe just using Start/Stop is a bit unusual. Rather to use Started/Stopped? In that case, we have already `stopped` in the translation files and just need to add the `started`. When we talk about recurring invoices, this is fully implemented, `started_recurring_invoice` with message 'Succesfully started recurring invoice' for example. But I agree with you that maybe just using Started/Stopped would be enough.

turbo124:
We use start/stop for this in recurring invoices. I think this is the correct word for this setting.

Civolilah:
> We use start/stop for this in recurring invoices. I think this is the correct word for this setting.

@turbo124 Okay, the toaster message has been updated!

turbo124:
@Civolilah for recurring expenses, the only required field is frequency Id, this should automatically default to monthly.

Civolilah:
> @Civolilah for recurring expenses, the only required field is frequency Id, this should automatically default to monthly.

@turbo124 Sure I missed it, now it's adjusted. Just to let you know, I double checked the recurring invoices and saw that blank query returns `frequency_id=5` (Monthly), in the case of recurring expenses we don't have that parameter set to any value. I set this default value manually, but if you think it's better to return through an empty query, I can simply remove this manullay setting.

turbo124:
@Civolilah 

I"ve just pulled this in, i'm still unable to save a blank recurring expense, the save button is blanked out.

Civolilah:
> 

@turbo124 The button is disabled when no client is selected. I saw the same kind of disablement on another feature, so I just replicated it. You will now be able to create blank recurring expense, I have removed this kind of validation.

turbo124:
@Civolilah looks good.

One small change, endless recurring expenses have a frequency of -1, can we please alias this to the 'endless' translation string please.

turbo124:
@Civolilah 

One thing I have just notifced, the recurring expenses do not have the column functionality that the other entities have. Please ensure the columns functionality is also implemented in this PR please.

Civolilah:
> 

done

src/pages/recurring-expenses/common/hooks.tsx
```diff
@@ -0,0 +1,467 @@
...
+  'vendor',
+  'client',
+  'date',
+  'amount',
+  'public_notes',
+  'entity_state',
+  'archived_at',
+  //   'assigned_to', @Todo: Need to resolve relationship
+  //   'category', @Todo: Need to resolve relationship
+  'created_at',
```
Civolilah, line 58: I kept this comments same as for expense entity, please let me know if you have any updates here.

turbo124:
@Civolilah @beganovich LGTM

src/pages/expenses/common/hooks.tsx
```diff
@@ -65,6 +74,47 @@ export const expenseColumns = [
...
 type ExpenseColumns = typeof expenseColumns[number];
 
+export function useActions() {
+  const [t] = useTranslation();
+
+  const navigate = useNavigate();
+
+  const [, setExpense] = useAtom(expenseAtom);
+
+  const [, setRecurringExpense] = useAtom(recurringExpenseAtom);
```
beganovich, line 84: To skip the getter for the state, you can use `useUpdateAtom`.

```ts
const setExpense = useUpdateAtom(expenseAtom);
```

https://jotai.org/docs/utils/use-update-atom

src/pages/expenses/create/Create.tsx
```diff
@@ -26,42 +26,53 @@ import { TaxSettings } from './components/Taxes';
...
-  ) => {
-    setExpense((expense) => expense && { ...expense, [property]: value });
-  };
+  useEffect(() => {
+    if (mounted) {
+      return () => setExpense(undefined);
+    } else {
+      mounted = true;
+    }
+  }, []);
```
beganovich, line 71: Why?

src/pages/expenses/create/Create.tsx
```diff
@@ -26,42 +26,53 @@ import { TaxSettings } from './components/Taxes';
...
 export function Create() {
   const [t] = useTranslation();
 
   const navigate = useNavigate();
 
   const { documentTitle } = useTitle('new_expense');
+
   const { data } = useBlankExpenseQuery();
 
+  let mounted = false;
```
beganovich, line 42: Generally you should avoid this style of variables, as they aren't reactive (won't re-render).

src/pages/recurring-expenses/create/Create.tsx
```diff
@@ -0,0 +1,138 @@
...
+    }
+  }, [data]);
+
+  useEffect(() => {
+    if (mounted) {
+      return () => setRecurringExpense(undefined);
+    } else {
+      mounted = true;
+    }
+  }, []);
```
beganovich, line 71: Please let me know why we do this.

beganovich (changes requested):
Looks good, small changes are needed.

The documents tab is taking full page width (same with expenses). Can we please do 2/3 of the page?

image

src/pages/expenses/common/hooks.tsx
Civolilah: replaced

src/pages/expenses/create/Create.tsx
Civolilah: As we discussed on the Slack this part is removed from technical reasons

src/pages/expenses/create/Create.tsx
Civolilah: Same as above, removed

src/pages/recurring-expenses/create/Create.tsx
Civolilah: removed

Civolilah:
> 

@beganovich Including other small changes in the code the UI for document tab is also slightly changed and it looks like this:

Screenshot 2023-01-02 at 16 25 54

beganovich:
Looks good, thanks 👍

PR #320, opened 2022-12-20
[Feature] Transaction Filters

Civolilah:
Here is the screenshot of implemented solution for custom filters of Bank Transactions:

Screenshot 2022-12-21 at 00 39 24

@turbo124 I set `transaction_status` as a placeholder on the filter dropdown. If you agree with it, just add it to the translation files.


turbo124:
@Civolilah 

Change transaction_status to just status (do this across all status filters in the app)

@beganovich LGTM after the above change.

Civolilah:
> 

@turbo124 Okay, it has been changed!

beganovich (approved):
Looks great, thanks 👍

PR #321, opened 2022-12-21
[Bug Fix] Company Validation Handling

Civolilah:
Here is the screenshot how validation should looks like after getting 422 error:

Screenshot 2022-12-21 at 18 31 52

Note: This validation has mocked and it's just for preview. Just to make sure `Validation message` will come from `errors.message` API's response.


beganovich (approved):
Looks great, thanks 👍

PR #322, opened 2022-12-21
[Feature] Drop down icons

Civolilah:
Here are two screenshots of implemented solution for dropdown icons:

Screenshot 2022-12-28 at 17 50 28

Screenshot 2022-12-28 at 17 50 41

@beganovich @turbo124 Now all dropdown menus have icons implemented. In addition to working on adding icons, I refactored some small parts of the code to follow the current convention and create better formatting (specially related to dropdown menus). Let me know your thoughts.

### DIFF
--- a/src/components/dropdown/Dropdown.tsx
+++ b/src/components/dropdown/Dropdown.tsx
@@ -35,7 +35,8 @@ export function Dropdown(props: Props) {
         interactive={true}
         render={() => (
           <div
-            className={`box w-48 rounded-md shadow-lg py-1 bg-white dark:bg-gray-800 ring-1 ring-black ring-opacity-5 focus:outline-none ${props.className}`}
+            className={`box rounded-md shadow-lg py-1 bg-white dark:bg-gray-800 ring-1 ring-black ring-opacity-5 focus:outline-none ${props.className}`}
+            style={{ minWidth: 192, maxWidth: 235 }}
           >
             {/* eslint-disable-next-line @typescript-eslint/ban-ts-comment */}
             {/* @ts-ignore */}
--- a/src/components/dropdown/Dropdown.tsx
+++ b/src/components/dropdown/Dropdown.tsx
@@ -36,7 +36,7 @@ export function Dropdown(props: Props) {
         render={() => (
           <div
             className={`box rounded-md shadow-lg py-1 bg-white dark:bg-gray-800 ring-1 ring-black ring-opacity-5 focus:outline-none ${props.className}`}
-            style={{ minWidth: 192, maxWidth: 235 }}
+            style={{ minWidth: '12rem', maxWidth: '14.7rem' }}
           >
             {/* eslint-disable-next-line @typescript-eslint/ban-ts-comment */}
             {/* @ts-ignore */}
--- a/src/components/dropdown/DropdownElement.tsx
+++ b/src/components/dropdown/DropdownElement.tsx
@@ -8,12 +8,17 @@
  * @license https://www.elastic.co/licensing/elastic-license
  */
 
+import classNames from 'classnames';
+import { Icon } from 'components/icons/Icon';
+import { ReactNode } from 'react';
 import { Link } from 'react-router-dom';
 import CommonProps from '../../common/interfaces/common-props.interface';
 
 interface Props extends CommonProps {
   to?: string;
   setVisible?: (value: boolean) => any;
+  iconType?: 'view' | 'new' | 'archive' | 'delete' | 'purge';
+  icon?: ReactNode;
 }
 
 export function DropdownElement(props: Props) {
@@ -21,9 +26,22 @@ export function DropdownElement(props: Props) {
     return (
       <Link
         to={props.to}
-        className={`w-full text-left hover:bg-gray-100 z-50 block px-4 py-2 text-sm text-gray-700 ${props.className}`}
+        className={classNames(
+          {
+            'flex items-center': props.iconType,
+          },
+          `w-full text-left hover:bg-gray-100 z-50 block px-4 py-2 text-sm text-gray-700 ${props.className}`
+        )}
       >
-        {props.children}
+        {props.iconType && !props.icon && <Icon type={props.iconType} />}
+        {props.icon}
+        <div
+          className={classNames({
+            'ml-2': props.iconType,
+          })}
+        >
+          {props.children}
+        </div>
       </Link>
     );
   }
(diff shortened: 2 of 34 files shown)


turbo124:
LGTM if it is ok with @beganovich

src/components/icons/Icon.tsx
```diff
@@ -0,0 +1,69 @@
+/**
+ * Invoice Ninja (https://invoiceninja.com).
+ *
+ * @link https://github.com/invoiceninja/invoiceninja source repository
+ *
+ * @copyright Copyright (c) 2022. Invoice Ninja LLC (https://invoiceninja.com)
+ *
+ * @license https://www.elastic.co/licensing/elastic-license
+ */
+
```
beganovich, line 10: I don't think I understand what's the purpose of this component. Why can't we just directly pass the icon reference to the dropdown component instead of having deeply locked Icon component?

src/components/icons/Icon.tsx
```diff
@@ -0,0 +1,69 @@
...
+} from 'react-icons/md';
+
+interface Props {
+  type: 'view' | 'new' | 'archive' | 'delete' | 'purge';
+  fontSize?: number;
+  color?: string;
+}
+
+export function Icon(props: Props) {
+  const fontSize = props.fontSize ? props.fontSize : 18;
```
beganovich, line 26: This is not the correct naming scheme. The correct prop should be `size` since you are in the context of icons.

src/components/icons/Icon.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  switch (props.type) {
+    case 'view':
+      return (
+        <MdPictureAsPdf
+          fontSize={fontSize}
+          color={props.color ? props.color : '#278ADF'}
```
beganovich, line 33: Not following accent color.

src/pages/clients/common/components/CustomResourcefulActions.tsx
```diff
@@ -46,6 +46,7 @@ export function CustomResourcefulActions(props: Props) {
           to={route('/clients/:id/statement', {
             id: props.clientId,
           })}
+          iconType="view"
```
beganovich, line 49: I don't think I get this. What does icon type "new" has to do with "View statement"?

beganovich (changes requested):
Looks good, few changes are needed.

1. Icons in the dropdown are not following the accent color schemes.
2. Colorful buttons for bulk actions may look cool in blue color, but as soon as you switch to something else 🤮

src/components/dropdown/DropdownElement.tsx
```diff
@@ -8,22 +8,40 @@
...
+import classNames from 'classnames';
+import { Icon } from 'components/icons/Icon';
+import { ReactNode } from 'react';
 import { Link } from 'react-router-dom';
 import CommonProps from '../../common/interfaces/common-props.interface';
 
 interface Props extends CommonProps {
   to?: string;
   setVisible?: (value: boolean) => any;
+  iconType?: 'view' | 'new' | 'archive' | 'delete' | 'purge';
```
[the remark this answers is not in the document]
Civolilah: Actually, apart from this one `iconType` we have an `icon` prop, the `iconType` prop is only used for frequent icons (New, Archive...), but another icon prop allows us to pass maybe a specific icon.

src/components/dropdown/DropdownElement.tsx
```diff
@@ -8,22 +8,40 @@
...
+        className={classNames(
+          {
+            'flex items-center': props.iconType,
+          },
+          `w-full text-left hover:bg-gray-100 z-50 block px-4 py-2 text-sm text-gray-700 ${props.className}`
+        )}
       >
-        {props.children}
+        {props.iconType && !props.icon && <Icon type={props.iconType} />}
+        {props.icon}
```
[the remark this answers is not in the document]
Civolilah: `{props.iconType && !props.icon && <Icon type={props.iconType} />}`

Icons from the `Icon` component will only appear if the `icon` prop doesn't exist, so we won't have a case where we have two icons.

src/components/dropdown/DropdownElement.tsx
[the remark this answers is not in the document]
Civolilah: > Actually, apart from this one `iconType` we have an `icon` prop, the `iconType` prop is only used for frequent icons (New, Archive...), but another icon prop allows us to pass maybe a specific icon. But, as we discussed on the Slack the logic will be slightly changed.

src/components/icons/Icon.tsx
Civolilah: As we discussed on the Slack this component will just get new shape of purpose.

src/pages/clients/common/components/CustomResourcefulActions.tsx
Civolilah: Hmm, I think that you missed it. It's actually, 'view'.

src/components/icons/Icon.tsx
Civolilah: Changed

src/components/icons/Icon.tsx
Civolilah: Now, we are using accentColor

src/components/dropdown/Dropdown.tsx
```diff
@@ -35,7 +35,8 @@ export function Dropdown(props: Props) {
         interactive={true}
         render={() => (
           <div
-            className={`box w-48 rounded-md shadow-lg py-1 bg-white dark:bg-gray-800 ring-1 ring-black ring-opacity-5 focus:outline-none ${props.className}`}
+            className={`box rounded-md shadow-lg py-1 bg-white dark:bg-gray-800 ring-1 ring-black ring-opacity-5 focus:outline-none ${props.className}`}
+            style={{ minWidth: 192, maxWidth: 235 }}
           >
```
Civolilah, line 40: The reason for the style change for this component is the wider width of all the dropdown menus when we included the icons at the beginning. Tailwind doesn't give us a `min-w-48` (for example) class to set, so I had to set it through the style property.

Civolilah:
@beganovich @turbo124 The icons are assigned to all the drop-down menus I know. Please let me know if I've missed any dropdowns in the app.

turbo124:
@Civolilah 

Looks good, however we need to make the edit action the first option across the application.

Civolilah:
> @Civolilah
> 
> Looks good, however we need to make the edit action the first option across the application.

@turbo124 It's implemented and as we discussed on Slack, I've removed the entity name from the dropdowns where appropriate as it's obvious that editing that entity. (`edit_product` -> `edit`) Also, description of this PR is updated with screenshots of how pages look now.

turbo124:
@Civolilah @beganovich 

LGTm

src/components/dropdown/Dropdown.tsx
beganovich: This is okay, but Tailwind min/max are defined in REMs. Can we do the same? Like this (I think) it's in pixels which means it's not fully reactive to different screen sizes?

beganovich:
Looks good, just to resolve that issue with defining the width and height in rems.

src/components/dropdown/Dropdown.tsx
Civolilah: Changed

Civolilah:
> Looks good, just to resolve that issue with defining the width and height in rems.

@beganovich done

beganovich:
Hey, @Civolilah looks like we have some merge conflicts. Can you fix them so I can merge?

Civolilah:
> 

@beganovich Of course, I'll move this PR to draft until I resolve the minor merge conflicts.

Civolilah:
@beganovich Merge conflicts resolved.

beganovich:
Looks good, thanks :+1:

PR #323, opened 2022-12-21
[Bug Fix] Payments sorting

Civolilah:
Here is the screenshot of resolved bug with payment sorting:

Screenshot 2022-12-21 at 23 57 27

@turbo124 We also had a bug with the perPage parameter. We sent 50 through the endpoint but only got 10 in the result (the default value). So this PR also includes a fix for this also.


beganovich (approved):
Looks good, thanks 👍

PR #325, opened 2022-12-23
[Feature] Email Verification Banner

Civolilah:
Here is the screenshot of implemented solution for showing banner when user does not have verified email:

Screenshot 2022-12-23 at 00 50 51

@turbo124 The email verification flow is:

1. When the user enters the application, we will immediately check the email verification.
2. If the user does not have a verified email, the banner will be displayed and they will not be able to use anything behind the banner (any app feature).
3. Also, there is a resend email button, which works perfectly and returns a status of 200, but as we know we can't send emails on the staging environment.
4. After verification, the banner will not be shown anymore.
5. For phone verification, I think that we will have to allow the user to use only one part of the application related to phone verification.

I included only `HOSTED` here same as for phone verification. Is that correct?

Let me know your thoughts.


turbo124:
@Civolilah 

This should be a red alert banner inthe header please.

Civolilah:
> @Civolilah
> 
> This should be a red alert banner inthe header please.

@turbo124 Here is the screeshot of restyled banner for email verification:

Screenshot 2022-12-28 at 00 37 27

Let me know your thoughts.

src/pages/settings/expense-categories/components/CreateExpenseCategoryModal.tsx
```diff
@@ -27,7 +27,6 @@ export function CreateExpenseCategoryModal(props: Props) {
       title={t('create_expense_category')}
       visible={props.visible}
       onClose={() => props.setVisible(false)}
-      size="small"
     >
```
Civolilah, line 30: The reason for removing this size is because we now have an `extraSmall` and `small` size. Now the small is a bit wider than before, so I had to remove it to set the `extraSmall' size as the default.

src/components/Modal.tsx
```diff
@@ -100,17 +103,22 @@ export function Modal(props: Props) {
                   </div>
                 </div>
 
-                <X
-                  className="cursor-pointer"
-                  onClick={() => props.onClose(false)}
-                />
+                {!props.disableClosing && (
+                  <X
```
Civolilah, line 107: What do you think about changing this react-icons instead of react-feather?

Civolilah:
@beganovich Just to post a screenshot for you too so you can see the latest version of the modal. David and I have been discussing the UI for this feature on Slack. So this is the final version that David agrees with:

Screenshot 2022-12-31 at 02 41 09

src/components/Modal.tsx
beganovich: I am not sure I follow. React Icons is a meta package and standalone doesn't provide icons. It's just bundle of different vendors for icons, including Feather.

src/App.tsx
```diff
@@ -30,6 +40,8 @@ export function App() {
     ? resolveLanguage(company.settings.language_id)
     : undefined;
 
+  const isLoginPage = location.pathname.startsWith('/login');
```
beganovich, line 43: Can we inline this if we don't reuse this variable?

src/App.tsx
```diff
@@ -56,8 +68,18 @@ export function App() {
     }
   }, [darkMode, resolvedLanguage]);
 
+  useEffect(() => {
+    if (user) {
+      setIsEmailVerified(!!user.email_verified_at);
```
beganovich, line 73: Please don't use double !!, kittens will die. 

Stick to Boolean as it's more readable.

src/App.tsx
```diff
@@ -56,8 +68,18 @@ export function App() {
...
+    }
+  }, [user]);
+
   return (
     <div className="App">
+      <VerifyModal
+        visible={!!user && !isLoginPage && !isEmailVerified && isHosted()}
```
beganovich, line 80: !! -> Boolean()

beganovich (changes requested):
Looks good, few changes are needed. 

I also think we should simplify the modal to make it look like the rest of te app.
"Refresh data" should also go away as we're going to implement this UX differently.

image

src/App.tsx
Civolilah: sure, made inline

src/App.tsx
Civolilah: changed

src/App.tsx
Civolilah: changed

Civolilah:
> 

@beganovich Including other small changes in the code logic, style is also changed to recommended and it looks like this:

Screenshot 2023-01-02 at 21 34 38

PR #336, opened 2022-12-30
[Bugfix/Cleanup] Transaction amount

Civolilah:
@turbo124 We only had a small bug with setting the default value for the currency separators when creating the transaction, so the amount field actually works like this. To manage this field, we use the `DecimalNumberInput` input component and set the currency separators from the currently selected `currency` in the drop-down menu in the field below. When we change the currency, the separators in the amount field will change automatically. The implementation is the same as for invoices.

@beganovich Also, I saw a chance for a good refactoring and renaming of some things. I hope you are okay with this type of refactoring. Now the code is easier to maintain and more readable.


beganovich (approved):
Looks good, thanks 👍

PR #337, opened 2022-12-31
[Feature] Quick create entity

Civolilah:
Here are the screenshots of created popover with button for creating some of available entities in the app:

Screenshot 2023-01-05 at 00 21 04

Screenshot 2023-01-05 at 00 22 34

@beganovich @turbo124 In the first screenshot, I unconditionally turned on all the listed options for the Settings section. Conditions are applyed in the second. So the functionality is based on navigating to the `Create` page for a specific entity or some settings pages. `Settings` section is only available for `HOSTED` users.


turbo124:
@Civolilah lets discuss UI for this on slack

Civolilah:
> Here is the screenshot of created dropdown with button for creating all available entities in the app:
> 
> Screenshot 2022-12-31 at 02 20 54
> 
> @turbo124 As you mentioned in the Jira ticket this is just a prototype and we will discuss the functionality later. I assume you envision the flow something like this:
> 
> 1. Clicking on a specific dropdown will immediately call the specific API route to create that entity
> 2. After creating the entity, we will navigate user to the edit page of the entity
> 
> Let me know your thoughts.

s

beganovich:
Looks good, thanks :+1:

PR #338, opened 2022-12-31
[Feature] User details notification options

Civolilah:
Here is the screenshot where we have added missing options for user details:

Screenshot 2022-12-31 at 03 25 44

@turbo124 As you can notice on the second option we don't have a proper label, there we used the `notification_invoice_sent` keyword, to fix this label bug I replaced it with `invoice_sent`, but this one contains the `:count` parameter. Also, this one parameter has been replaced with '', but now our label is `invoice sent` instead of `Invoice sent`. If you agree with this type of translation implementation, just update the `invoice_sent` keyword.

Here is the screenshot with updated translation keyword:

Screenshot 2022-12-31 at 03 38 28


beganovich:
Looks good, thanks 👍

PR #339, opened 2022-12-31
[Bug fix] Custom filter placeholders

Civolilah:
@turbo124 The custom placeholder filters have been changed to `'status'` for Invoices and Recurring invoices and there are only two places where we missed it.


beganovich:
Looks good, thanks 👍

PR #352, opened 2023-01-05
[Bugfix] Quote error status badge

Civolilah:
@turbo124 @beganovich Me and David were talking on Slack today about the `Sent` and `Expired` status code and we found a solution that the -1 status code is correct for the `Expired` Quote status.


beganovich (approved):
Looks good, thanks :+1:

PR #359, opened 2023-01-06
[Bugfix] Confirm email modal showing

Civolilah:
@beganovich @turbo124 This PR only slightly changes the modal displaying logic. Currently in production, we show the modal by default, but if the user has verified their email, we hide it. On this PR we just reversed the logic to avoid the current behavior to avoid unnecessary rendering of this modal.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -30,7 +30,7 @@ export function App() {
 
   const location = useLocation();
 
-  const [isEmailVerified, setIsEmailVerified] = useState(false);
+  const [isEmailVerified, setIsEmailVerified] = useState<boolean>(true);
 
   const resolveLanguage = useResolveLanguage();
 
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -67,7 +67,7 @@ export function App() {
   }, [darkMode, resolvedLanguage]);
 
   useEffect(() => {
-    if (user) {
+    if (user && Object.keys(user).length) {
```
beganovich, line 70: What exactly are we checking here for?

AFAIK `user` is `User | undefined` the first part of the check should be enough.

Civolilah: I agree with your point, but actually when we enter the application the initial value of the user is an empty object {} but it is not undefined. An empty object is also a truthy value, so just putting that part in the if statement isn't enough. The second part is just checking if the object is empty, if it is, we won't do anything. This was one of the most important parts for making bug on the production.

beganovich: Got it, I thought it was undefined or populated object. Thanks!

beganovich (approved):
Looks good, thanks 👍

PR #360, opened 2023-01-06
[Feature] Company activity modal

Civolilah:
Here is the screenshot of implemented solution for displaying modal when company is not activated yet:

Screenshot 2023-01-09 at 03 51 06

@beganovich @turbo124 Flow of this feature looks like this:

- When the user enters each new session, we will double check that the company is not yet activated
- If the company is not activated, we will show this modal and never again in that session
- The Settings button will navigate the user to the `Account Management` section of the app where they can activate it
- Or he can just close the modal
- The easiest way to check if the modal is already displayed is to transfer the data from the session storage, when the user closes, the application session storage data will be deleted

@turbo124 For the title of the screenshot, I set the plain text Company Activity, replaced with the keyword `company_activity`. So, just add it to the translation resource files if it's ok on your end too.


Civolilah:
> Here is the screenshot of implemented solution for displaying modal when company is not activated yet:
> 
> Screenshot 2023-01-09 at 03 51 06
> 
> @beganovich @turbo124 Flow of this feature looks like this:
> 
> * When the user enters each new session, we will double check that the company is not yet activated
> * If the company is not activated, we will show this modal and never again in that session
> * The Settings button will navigate the user to the `Account Management` section of the app where they can activate it
> * Or he can just close the modal
> * The easiest way to check if the modal is already displayed is to transfer the data from the session storage, when the user closes, the application session storage data will be deleted

ss

turbo124:
@Civolilah feedback on slack for this

Civolilah:
> @Civolilah feedback on slack for this

@turbo124 Fixed a bug we discussed on Slack so it wasn't appearing modal on the company switch and also removed the restriction to only show to `HOSTED` users. Should this modal be displayed for both types of users?

turbo124:
@Civolilah yes, activate modal is for all users.

turbo124:
@Civolilah 

Please change company_activity label to activate_company

Civolilah:
> @Civolilah yes, activate modal is for all users.

@turbo124 In that case, it's adjusted for both types of users.

Civolilah:
> 

@turbo124 Changed!

beganovich:
The code looks good, however, I think the style of the modal is unnecessarily complicated. 

image

beganovich (approved):
Looks good, thanks 👍

PR #364, opened 2023-01-07
[Feature] Product sales report

Civolilah:
Here is the screenshot of implemented solution for product sales report:

Screenshot 2023-01-09 at 02 45 56

@turbo124 `client_id` is sent through the request body. Let me know what you think about the functionality.


turbo124:
@Civolilah 

Looks good, but once a client has been set, i am unable to clear it.

Also, if you can set the placeholder for the Client to "All" (pass a null value for this) just to indicate to the user that it is possible to generate a report for a single client, or all.

Civolilah:
> 

@turbo124 The bug with no close option has been fixed on the `Subscriptions` PR so when we merge it, it will apply to all other selected components in the app.

For the additional `all` option you mentioned here, I changed the client selection component. I realized that we really don't need a `New Client` option in this part of the app. So now we have the `all` option and if it is selected we will send a `null` value in the request body. Otherwise, the `client_id` value will be sent.

turbo124:
@Civolilah 

The problem here is we lose the ability to search using the combo box, I think we can rewind this, if no client is set, we can assume the user wants to export all product sales.

Civolilah:
> 

@turbo124 As we discussed in Slack, the best solution for us is to exclude the `New Client` action option and if there is no `client_id` selected, we will send `null` to the request body. Otherwise, we will send the `client_id` as a string. The description has been updated with a new screenshot.

src/pages/reports/index/Reports.tsx
```diff
@@ -324,9 +341,21 @@ export function Reports() {
...
+
+    request(
+      'POST',
+      endpoint(report.endpoint),
+      {
+        ...report.payload,
+        ...(report.identifier === 'product_sales' && {
+          client_id: !client_id ? null : client_id,
+        }),
+      },
```
beganovich, line 354: Can we move this level above with a clear `if`? This way it's a bit unreadable and I can imagine someone crying trying to read this one :laughing:

src/pages/reports/index/Reports.tsx
```diff
@@ -470,6 +499,18 @@ export function Reports() {
...
+          {report.identifier === 'product_sales' && (
+            <Element leftSide={t('client')}>
+              <ClientSelector
+                value={report.payload.client_id}
+                onChange={(client) =>
+                  handlePayloadChange('client_id', client.id)
+                }
+                withoutAction={true}
+              />
+            </Element>
```
beganovich, line 512: It looks like we are missing a clear button.

beganovich (changes requested):
Looks good, small changes are needed.

src/pages/reports/index/Reports.tsx
Civolilah: 😄 , The point of those conditions is to not send `client_id` property to the API if it is not needed. I believe that I made more clear code now. Changed!

src/pages/reports/index/Reports.tsx
Civolilah: Good catch, it's added!

beganovich (approved):
Looks good, thanks 👍

PR #366, opened 2023-01-08
[Cleanup] Danger zone CTA

Civolilah:
Here is a screenshot of the danger zone when we only have one company created on the account:

Screenshot 2023-01-08 at 01 23 33

Here is a screenshot of the danger zone when we have multiple companies created on the account:

Screenshot 2023-01-08 at 01 25 07

@turbo124 Just one thing to mention as my suggestion. We already have the same help text on the modal after clicking one of those actions, it might be better to avoid one place to put the help text. IMO, it's good to have right below the action buttons or just on the modal. The text messages on modals are also adjusted. Here are screenshots of the modal:

Screenshot 2023-01-08 at 01 42 45

Screenshot 2023-01-08 at 01 41 56

Screenshot 2023-01-08 at 01 41 18

### DIFF
--- a/src/components/Modal.tsx
+++ b/src/components/Modal.tsx
@@ -88,27 +88,29 @@ export function Modal(props: Props) {
                 }
               )}
             >
-              <div className="flex justify-between items-start">
-                <div>
+              <div className="flex flex-col justify-between items-start">
+                <div className="flex w-full justify-between">
                   <Dialog.Title
                     as="h3"
                     className="text-lg leading-6 font-medium text-gray-900"
                   >
                     {props.title}
                   </Dialog.Title>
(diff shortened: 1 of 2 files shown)


turbo124:
@Civolilah thans, ok, we can remove the additional help text as it appears redundant.

Civolilah:
> 

@turbo124 Unnecessary helper text has been removed, so we now have help text displayed correctly on modals. Also, the delete company/account button has appropriate text. We had the `Danger Zone` title set as plain text so I replaced it with the `danger_zone` keyword, so just add it to the resource files.

beganovich (approved):
Looks good, thanks :+1:

PR #369, opened 2023-01-10
[Feature] Sms verify modal

Civolilah:
Here is the screenshot of implemented solution for sms verify modal:

Screenshot 2023-01-13 at 00 17 35

@turbo124 Behaviour should be like this:

- Once user enters the app we will check if `account_sms_verified` is false
- If this property is false we will show the modal and never again in that user's session
- This is only available for `HOSTED` users


Civolilah:
@beganovich In the last commit I just reused your style and created a component that will handle both types of warning (for company activity and sms phone verification). I think it's better than create one more unnecessary component, but style is the same. 

@turbo124 Just please add `go_to_settings` keyword to translation files.

turbo124:
@Civolilah 
Please use the translations key `verify_phone_number` for this.

Civolilah:
> @Civolilah Please use the translations key `verify_phone_number` for this.

@turbo124 Changed! Screenshot in the description updated!

turbo124:
@Civolilah small update, the padding between the entities are out of proportion, the functionality here is correct, but i'll default to @beganovich and his opinion on layout.

beganovich (approved):
Looks good, thanks 👍

PR #370, opened 2023-01-10
[Cleanup] Connect accounts

Civolilah:
Here is the screenshot of implemented solution when there is no provider connected (default):

Screenshot 2023-01-10 at 17 23 24

This is behaviour when `Google` provider is connected to account:

Screenshot 2023-01-10 at 17 23 45

This is behaviour when `Microsoft` provider is connected to account:

Screenshot 2023-01-10 at 17 24 07

@beganovich  @turbo124 As we discussed on the Slack, the current focus of this PR is to get the buttons displayed correctly, the Connect Email functionality mentioned in the Jira ticket is also implemented here. I checked the Flutter app about this feature and noticed some really weird behavior. If the account is connected, for example to Google provider, disconnection will be available by clicking the same button called Connect Gmail. IMO, this UX is a bit confusing, I'm not sure how this part will work in the future when we implement all those parts, but I think it's better when we connect Gmail to show the Disconnect Gmail button. Literally, to make the same behavior as for Connect Google -> Disconnect Google. Let me know your thoughts.


beganovich (approved):
Looks good, thanks 👍

PR #374, opened 2023-01-11
[Feature] Task create option

Civolilah:
Here is the screenshot of implemented solution for 'Creating' task instead of just 'Saving':

Screenshot 2023-01-12 at 01 14 00

@turbo124 Important note: The solution is globalized so there will be less work if we want to implement the same functionality for any other entity where we use this `Card` component.

@beganovich Apart from the function, I refactored the code to be consistent with the rest of the project.

### DIFF
--- a/src/components/cards/Card.tsx
+++ b/src/components/cards/Card.tsx
@@ -10,7 +10,7 @@
 
 import classNames from 'classnames';
 import { Spinner } from 'components/Spinner';
-import { CSSProperties, FormEvent, ReactNode, useState } from 'react';
+import { CSSProperties, FormEvent, ReactNode } from 'react';
 import { useTranslation } from 'react-i18next';
 import { CardContainer } from '.';
 import { Button } from '../forms';
@@ -18,7 +18,7 @@ import { Element } from 'components/cards/Element';
 import { Dropdown } from 'components/dropdown/Dropdown';
 import { DropdownElement } from 'components/dropdown/DropdownElement';
 import { Icon } from 'components/icons/Icon';
-import { BiPlusCircle, BiSave } from 'react-icons/bi';
+import { BiPlusCircle } from 'react-icons/bi';
 
 interface Props {
   children: ReactNode;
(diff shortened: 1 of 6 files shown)


turbo124:
@Civolilah 

Looks good. i think an important change here is to make the save button a split dropdown. this will prevent the user needing to perform two actions to save the task status. The primary button will be the regular Save action, and the secondary dropdown will just save Save and Create.

image

I don't think there is a native split dropdown for tailwind, however it should not be too difficult to craft one.

Civolilah:
> 

@turbo124 Just changed UI for this functionality, let me know your thoughts.

turbo124:
@Civolilah 

Looks good, small changes required.

Please give the left margin of the dropdown a darker border than the button itself to show that it is a split button.

Also the functionality of the buttons are not quite right.

The action here should not switch the buttons around at all. The save button should just save. And the Save/Create New should save the status and then display the blank form again for a new status.

Civolilah:
> 

@turbo124 Ah sorry, I didn't catch you the first time. Now, I believe everything is implemented as you mentioned. Just one thing, I think it's better to separate the dropdown icon from the button with a lighter border than a darker one. Let me know your thoughts.

src/components/cards/Card.tsx
```diff
@@ -10,25 +10,32 @@
...
+import { BiPlusCircle } from 'react-icons/bi';
 
 interface Props {
   children: ReactNode;
   title?: string | null;
   description?: string;
   withSaveButton?: boolean;
-  onFormSubmit?: (event: React.FormEvent<HTMLFormElement>) => any;
-  onSaveClick?: any;
+  withCreateOption?: boolean;
```
beganovich, line 28: I don't think this is good naming. 

If we gonna support multiple options for card button it shouldn't be named withCreateOption. Can we abstract this and generalize the dropdown?

beganovich (changes requested):
Looks good, few changes needed 👍

src/components/cards/Card.tsx
Civolilah: I hope that I got your idea about generalizing drop down options for save button on this component. So, I made logic something similar to `more_actions` options. Let me know your thoughts.

beganovich:
Looks good, thanks 👍

beganovich (approved):
Looks good, thanks 👍

PR #376, opened 2023-01-11
[Feature] Kanban add task status

Civolilah:
Here are screenshots of implemented solution for adding task status via modal on Kanban page:

Screenshot 2023-01-11 at 18 57 23

Screenshot 2023-01-11 at 18 57 34

@turbo124 IMO, it's a better solution to have this button sticky on the right side instead of at the end after scrolling through all the displayed task statuses. Let me know your thoughts.


src/pages/settings/task-statuses/components/CreateTaskStatusModal.tsx
```diff
@@ -0,0 +1,124 @@
...
+  return (
+    <Modal
+      title={t('new_task_status')}
+      visible={props.visible}
+      onClose={() => {
+        props.setVisible(false);
+        setTaskStatus(blankTaskStatus);
+      }}
+    >
+      <div className="space-y-6 px-4">
```
beganovich, line 103: There's no need for this spacing. Please keep the one that's default to all modals.

src/pages/tasks/kanban/Kanban.tsx
```diff
@@ -324,97 +332,117 @@ export function Kanban() {
...
-                  <div
-                    className="bg-white rounded shadow select-none h-max"
-                    style={{ minWidth: 360 }}
-                  >
-                    <div className="border-b border-gray-200 px-4 py-5">
-                      <h3 className="leading-6 font-medium text-gray-900">
-                        {board.title}
-                      </h3>
-                    </div>
-
```
beganovich, line 344: Can you explain why we are wrapping in double flex?

beganovich (changes requested):
I wouldn't say I like the fixed/static + button. 
I suggest replicating what Jira did (non-static).

@turbo124 thoughts?

src/pages/settings/task-statuses/components/CreateTaskStatusModal.tsx
Civolilah: Sure, removed!

src/pages/tasks/kanban/Kanban.tsx
Civolilah: If I just put a new icon div in a div with classes `flex pb-6 space-x-4 overflow-x-auto mt-4` and if I say ok be sticky and `right :0` it will look like this:

Screenshot 2023-01-17 at 14 42 58

The icon div doesn't know that we have empty space on the right because of the padding. So we need one parent to run this part. In that case the div with the mentioned classes will be styled with `pr-6` and the icon will fit without any problems. If I exclude the flex class from the div I just added, the icon will be at the bottom (direction will be column by default).

I've only tried some other solutions, but I haven't found any simpler than this one.

turbo124: As long as the icon appears next to the last status, i'm fine.

Civolilah:
@turbo124 @beganovich The UI has been changed and now the button to add a new task status is at the end of the horizontal scroll.

Screenshot 2023-01-19 at 13 43 45

turbo124:
LGTM

beganovich (approved):
Looks good, thanks 👍

PR #378, opened 2023-01-12
[Feature] Add task kanban option

Civolilah:
Here are screenshots of UI for add task from Kanban page option:

Screenshot 2023-01-12 at 17 22 28

Screenshot 2023-01-12 at 17 22 42

@turbo124 It is important to note that the user will not be able to change the task status if they try to add a task from the Kanban page, but the project will always be changeable. IMO, this behavior makes more sense, but let me know what you think.


src/pages/tasks/edit/Edit.tsx
```diff
@@ -89,7 +96,9 @@ export function Edit() {
       navigationTopRight={task && <Actions task={task} />}
       onSaveClick={() => task && handleSave(task)}
     >
-      {task && <TaskDetails task={task} handleChange={handleChange} />}
+      {task && (
+        <TaskDetails task={task} handleChange={handleChange} errors={errors} />
```
Civolilah, line 100: With this change, I actually included errors as response to my part of the function and the part of the function you implemented earlier. If I include errors as an prop to `TaskDetails`, I have to implement it in every component where we use this component. So that's the reason for the code change here.

turbo124:
@Civolilah looks great!!

src/pages/tasks/kanban/components/CreateTaskModal.tsx
```diff
@@ -0,0 +1,147 @@
+/**
```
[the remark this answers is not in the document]
Civolilah: Yeah, you are right, it's moved!

src/pages/tasks/kanban/Kanban.tsx
```diff
@@ -237,11 +249,14 @@ export function Kanban() {
...
-          })
+          route(
+            '/api/v1/tasks?project_tasks=:projectId&limit=1000&per_page=500',
+            {
+              projectId,
+            }
+          )
         )
-      : setApiEndpoint('/api/v1/tasks?limit=1000');
+      : setApiEndpoint('/api/v1/tasks?limit=1000&per_page=500');
```
beganovich, line 259: May I know why this was added? We don't have pagination on the kanban page, thus limit & per_page make no sense.

Civolilah: Sure, I have an explanation for the `per_page` meta data, but the `limit` was already there so I don't know what impact it has. So even if we don't have any pagination the API will return paginable data, by default I think it will be 15/20 results per page and when we fetch the data from the API only the first page will be set to the state that will be used to display data on the UI. If I exclude meta data `per_page` I always get 20 results even if the total result is 121 in my case. I've been researching about react queries and I think it's not about the settings in the react query options, it's about how our logic works on the API. Do you know of any way to remove the pagable logic in the sepcific request?

For better understanding here is the screenshot:

Screenshot 2023-01-16 at 14 46 36

Here is also GenericManyResponse interface which includes this part for meta data:

```
export interface GenericManyResponse<T> {
  data: T[];
  meta: {
    pagination: {
      total: number;
      count: number;
      per_page: number;
      current_page: number;
      total_pages: number;
      links: string[];
    };
  };
}
```

Now, I just set `per_page` query parameter same as limit to 1000. Let me know your thoughts.

beganovich:
Looks good, thanks 👍

PR #380, opened 2023-01-13
[Feature] Task timer updating

Civolilah:
Here is the screenshot of implemented solution for updating task timer if task is running:

Screenshot 2023-01-13 at 03 54 56

@turbo124 @beganovich Let me know your thoughts.

### DIFF
--- a/src/common/queries/tasks.ts
+++ b/src/common/queries/tasks.ts
@@ -48,17 +48,18 @@ interface TasksParams {
 
 export function useTasksQuery(params: TasksParams) {
   return useQuery<GenericManyResponse<Task>>(
-    route(':endpoint?limit=:limit', {
+    route(':endpoint', {
       endpoint: params.endpoint || '/api/v1/tasks',
       limit: params.options?.limit.toString() || 10,
     }),
     () =>
       request(
         'GET',
-        endpoint(':endpoint?limit=:limit', {
+        endpoint(':endpoint', {
           endpoint: params.endpoint || '/api/v1/tasks',
           limit: params.options?.limit.toString() || 10,
         })
-      ).then((response) => response.data)
+      ).then((response) => response.data),
+    { staleTime: Infinity }
   );
 }
--- a/src/common/queries/tasks.ts
+++ b/src/common/queries/tasks.ts
@@ -40,9 +40,6 @@ export function useBlankTaskQuery() {
 }
 
 interface TasksParams {
-  options?: {
-    limit: number;
-  };
   endpoint?: string;
 }
 
@@ -50,14 +47,12 @@ export function useTasksQuery(params: TasksParams) {
   return useQuery<GenericManyResponse<Task>>(
     route(':endpoint', {
       endpoint: params.endpoint || '/api/v1/tasks',
-      limit: params.options?.limit.toString() || 10,
     }),
     () =>
       request(
         'GET',
         endpoint(':endpoint', {
           endpoint: params.endpoint || '/api/v1/tasks',
-          limit: params.options?.limit.toString() || 10,
         })
       ).then((response) => response.data),
     { staleTime: Infinity }
--- a/src/pages/tasks/kanban/Kanban.tsx
+++ b/src/pages/tasks/kanban/Kanban.tsx
@@ -83,7 +83,7 @@ export function Kanban() {
 
   const queryClient = useQueryClient();
 
-  const [apiEndpoint, setApiEndpoint] = useState('/api/v1/tasks?per_page=500');
+  const [apiEndpoint, setApiEndpoint] = useState('/api/v1/tasks?per_page=1000');
   const [projectId, setProjectId] = useState<string>();
 
   const { data: taskStatuses } = useTaskStatusesQuery();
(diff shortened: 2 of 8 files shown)


turbo124:
I don't see the timers updating in real time, they have the static times, is it possible to have the timer running?

Civolilah:
> 

@turbo124 Sorry, I missed managing timers on the main Kanban page. I think I've covered all the points now. Let me know your thoughts.

turbo124:
@Civolilah looks great, thanks.

beganovich:
While this generally works, I am trying to understand why such overcomplication. I was able to come up with component for this in few lines, here's the code.

1. You modify `calculateTime` so it can return in seconds:

```ts
interface CalculateTimeOptions {
  inSeconds?: boolean;
}

export function calculateTime(log: string, options?: CalculateTimeOptions) {
  const times = parseTimeLog(log);

  let seconds = 0;

  times.map(([start, stop]) => {
    const startTime = start ? dayjs.unix(start) : dayjs();
    const stopTime = stop ? dayjs.unix(stop) : dayjs();

    seconds += stopTime.diff(startTime, 'seconds');
  });

  if (options?.inSeconds) {
    return seconds;
  }

  return new Date(seconds * 1000).toISOString().slice(11, 19);
}
```

2. TaskClock component

```tsx
interface Props {
  task: Task;
}

export function TaskClock(props: Props) {
  const [seconds, setSeconds] = useState<number>();

  useEffect(() => {
    const calculation = calculateTime(props.task.time_log, { inSeconds: true });

    if (isTaskRunning(props.task)) {
      setSeconds(calculation as number);

      const interval = setInterval(() => {
        setSeconds((current) => current ? current + 1 : current);
      }, 1000);

      return () => clearInterval(interval);
    }
  }, []);

  return (
    <div>
      {isTaskRunning(props.task) &&
        seconds &&
        new Date(seconds * 1000).toISOString().slice(11, 19)}
    </div>
  );
}
```

3. Final step is just using that component if task is running (perhaps you can put the fallback logic so you don't have to check if task is running everywhere and instead show static if not):

```tsx
<p>{card.title}</p>
                             
<small>
  {isTaskRunning(card.task) ? (
    <TaskClock task={card.task} />
  ) : (
    card.description
  )}
</small>
```

I tested this in kanban and it works just fine. Please let me know if you find any flaws in this.

src/pages/tasks/kanban/Kanban.tsx
```diff
@@ -50,8 +50,9 @@ import { Link as ReactRouterLink } from 'react-router-dom';
 import { Card, Element } from '@invoiceninja/cards';
 import { ProjectSelector } from 'components/projects/ProjectSelector';
 import { Inline } from 'components/Inline';
+import { TaskClock } from './components/TaskClock';
 
-interface Card {
+interface CardItem {
```
Civolilah, line 55: This change is for security reasons. We have an imported `Card` component and a `Card interface` so just to make a difference I renamed the interface.

src/common/queries/tasks.ts
```diff
@@ -40,25 +40,21 @@ export function useBlankTaskQuery() {
 }
 
 interface TasksParams {
-  options?: {
-    limit: number;
```
Civolilah, line 44: Me and @turbo124 discussed on the slack and realized that the `limit` option had no use here so it was removed. Instead, `per_page` will handle this part about limiting the results.

Civolilah:
> While this generally works, I am trying to understand why such overcomplication. I was able to come up with component for this in few lines, here's the code.
> 
> 1. You modify `calculateTime` so it can return in seconds:
> 
> ```ts
> interface CalculateTimeOptions {
>   inSeconds?: boolean;
> }
> 
> export function calculateTime(log: string, options?: CalculateTimeOptions) {
>   const times = parseTimeLog(log);
> 
>   let seconds = 0;
> 
>   times.map(([start, stop]) => {
>     const startTime = start ? dayjs.unix(start) : dayjs();
>     const stopTime = stop ? dayjs.unix(stop) : dayjs();
> 
>     seconds += stopTime.diff(startTime, 'seconds');
>   });
> 
>   if (options?.inSeconds) {
>     return seconds;
>   }
> 
>   return new Date(seconds * 1000).toISOString().slice(11, 19);
> }
> ```
> 
> 2. TaskClock component
> 
> ```tsx
> interface Props {
>   task: Task;
> }
> 
> export function TaskClock(props: Props) {
>   const [seconds, setSeconds] = useState<number>();
> 
>   useEffect(() => {
>     const calculation = calculateTime(props.task.time_log, { inSeconds: true });
> 
>     if (isTaskRunning(props.task)) {
>       setSeconds(calculation as number);
> 
>       const interval = setInterval(() => {
>         setSeconds((current) => current ? current + 1 : current);
>       }, 1000);
> 
>       return () => clearInterval(interval);
>     }
>   }, []);
> 
>   return (
>     <div>
>       {isTaskRunning(props.task) &&
>         seconds &&
>         new Date(seconds * 1000).toISOString().slice(11, 19)}
>     </div>
>   );
> }
> ```
> 
> 3. Final step is just using that component if task is running (perhaps you can put the fallback logic so you don't have to check if task is running everywhere and instead show static if not):
> 
> ```tsx
> <p>{card.title}</p>
>                              
> <small>
>   {isTaskRunning(card.task) ? (
>     <TaskClock task={card.task} />
>   ) : (
>     card.description
>   )}
> </small>
> ```
> 
> I tested this in kanban and it works just fine. Please let me know if you find any flaws in this.

@beganovich Sure, thank you for logic suggestion. It has been changed, the PR is ready for review.

beganovich (approved):
Looks good, thanks 👍

PR #384, opened 2023-01-15
[Feature] Task table - client overview page

Civolilah:
Here is the screenshot of implemented solution for task table on the client overview page:

Screenshot 2023-01-15 at 15 13 53

@turbo124 A `status` filter is also implemented here, I thought if we can offer why not add to this `Task` table. Also, some other tables don't have the status filter implemented on the Client Overview page, but that's a minor change to add. Let me know your thoughts.


beganovich (approved):
Looks good, thanks 👍

PR #385, opened 2023-01-15
[Cleanup] Statics query enabling 

Civolilah:
@turbo124 As I can see, `X-NINJA-TOKEN` is stored in local storage after login to the app. When the user logs out it will be removed. So I used it as a condition to enable the `/api/v1/static query`. If we have that query in local storage, we will enable this query, if we don't have it, it will be disabled.


turbo124:
Looks good, thanks!

beganovich (approved):
Looks good, thanks 👍

PR #392, opened 2023-01-15
[Cleanup] Optimize API calls - Expense create page

Civolilah:
@turbo124 @beganovich This PR includes optimizing API endpoint calls on the `Expense` create page, ie. putting some missing queries in the cache. This includes a few parts of our app where we missed put queries into cache, so I fixed it in this PR.


Civolilah:
> It'd be awesome if you can include edge cases that you ran into and explanation so I can understand why specific changes have been made. Thanks.

Sure, @beganovich. I've actually made a few changes and most of them are only related to the `staletime` parameter. So we didn't have `staletime` set for some queries and our API was called every time when some components were rendered. These are the cases:

- The `ClientSelector` component is one of the most used components of all selectors, and there we have the `ClientCreate` modal. In the `ClientCreate` modal we had non-convection code according to our current convention. So I just refactored this component a bit and created `useBlankClientQuery`. The refactoring is about clearing the cache and removing staletime={500}.
- `useQuickCreateActions` is a hook used to conditionally display content in a quick entity create popup in the navbar. As we know, this popup is displayed on every page because it is part of the `Default` component. In this hook, we have a few more queries for those conditions, and one of them is `useCompanyGatewaysQuery` because we need to check the number of gateways. We didn't set staletime here and the API was called every time the URL changed. So the staletime is added.
- `useTaxRatesQuery` is used on the Expense create page in some child components and has not set an staletime. So I just added it so we don't call the API on every render of the create Expense page, it should be cached.

beganovich:
I do understand every change but I am totally not sure about removing the stale time for ClientSelector. Why do we do that?

Civolilah:
> I do understand every change but I am totally not sure about removing the stale time for ClientSelector. Why do we do that?

@beganovich We use this `ClientSelector` component on many other components that include the `Expense` creation component. If we keep 500 as `staletime`, every time we enter the Expense creation page, the `ClientSelector` will call the API directly, even if it was called only a few seconds ago. We invalidate the query whenever we change the list of clients on the API. I think it's completely unnecessary to call the API every time if we can cache it. But, since staletime is only held for 0.5 seconds instead of `Infinitiy` time, it will be called very often. But, let me know if there is something I do not know.

beganovich:
If this is the case only on the expense create page, I suggest passing custom staleTime: Infinity to ClientSelector instead of disabling it for the whole app.

Civolilah:
> If this is the case only on the expense create page, I suggest passing custom staleTime: Infinity to ClientSelector instead of disabling it for the whole app.

@beganovich The logic is just updated, so ClientSelector has Infinity staleTime only for expense pages.

beganovich (approved):
Looks good, thanks 👍

PR #393, opened 2023-01-15
[Feature] View delivery note action

Civolilah:
Here is the screenshot of implemented view delivery action for invoice:

Screenshot 2023-01-15 at 23 04 15

@turbo124 @beganovich The action is performed as easily as possible. For me the best way to do this was something like creating an entity from the client overview page, with search parameters, so in this case the URL would look like `http://localhost:3000/#/invoices/M7e5yXYe2v/pdf?delivery_note=true`. Everything after that is handled in code. I've included two keywords in the dropdown menu for this action: `view_delivery_note` and `pdf`. So the action would be like this `View Delivery Note (PDF)`, because it's also in PDF as a regular `view_pdf` action. I hope you agree with me. I can not find the `view_delivery_note` keyword in the translation files so please just add it.

The same action we have from the table as well.

### DIFF
--- a/src/pages/invoices/edit/components/Actions.tsx
+++ b/src/pages/invoices/edit/components/Actions.tsx
@@ -129,7 +129,7 @@ export function useActions() {
         to={route('/invoices/:id/pdf?delivery_note=true', { id: invoice.id })}
         icon={<Icon element={MdPictureAsPdf} />}
       >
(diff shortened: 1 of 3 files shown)


beganovich (approved):
Looks good, thanks 👍

PR #396, opened 2023-01-17
[Cleanup] Missing fields in total section

Civolilah:
Here is the screenshot of implemented solution for additonal fields in the total section:

Screenshot 2023-01-17 at 22 31 32

@turbo124 I was a bit confused about what this task was about. But I hope I understood correctly what we need to finish here https://invoiceninja.atlassian.net/browse/RU-451. Totals section has been updated with new fields you mentioned in the Jira's ticket. Let me know your thoughts.

### DIFF
--- a/tests/unit/helpers/invoice-sum.test.ts
+++ b/tests/unit/helpers/invoice-sum.test.ts
@@ -28,8 +28,8 @@ describe('InvoiceSum test invoice calculation', () => {
   it('Line Item Calc', () => {
     const invoiceSum = new InvoiceSum(invoice, USD).build();
 
-    expect(invoiceSum.invoice.amount).toEqual(3723);
-    expect(invoiceSum.invoice.balance).toEqual(3723);
+    expect(invoiceSum.invoice.amount).toEqual(3963.9);
+    expect(invoiceSum.invoice.balance).toEqual(3963.9);
   });
 
   it('Single line item', () => {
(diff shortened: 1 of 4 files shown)


tests/unit/helpers/invoice-sum.test.ts
```diff
@@ -28,8 +28,8 @@ describe('InvoiceSum test invoice calculation', () => {
   it('Line Item Calc', () => {
     const invoiceSum = new InvoiceSum(invoice, USD).build();
 
-    expect(invoiceSum.invoice.amount).toEqual(3723);
-    expect(invoiceSum.invoice.balance).toEqual(3723);
+    expect(invoiceSum.invoice.amount).toEqual(3963.9);
```
Civolilah, line 31: After changing the code in the invoice-sum class, the total number changed. As I can see, you have set a fixed number here. So I just set the one shown in the terminal as `received`. Please let me know if this number must be 3723.

turbo124: No, this is a bug. you will need to find out why your branch has introduced this bug, if you run the test in develop, you'll see the test pass correctly with the invoice amount/balance set to 3723

I've also double checked the calculations and 3723 is correct.

Civolilah:
@turbo124 @beganovich I think I figured out why we got a bug with those calculations, so it's fixed now, and I also hope the UI looks as expected.

turbo124:
Thanks!

beganovich (approved):
Looks good, thanks :+1:

PR #402, opened 2023-01-18
[Cleanup] Company name displaying in dropdown

Civolilah:
Here is the screenshot of how it looks when we do not have any name set for specific company:

Screenshot 2023-01-18 at 18 54 49

@turbo124 You mentioned in the ticket that maybe we can use `company.settings.name`, but I'm not sure that's possible and that's good logic. Here we only have a list of all available companies and that company has no name set now we set `untitled_company`. If I set `company.settings.name` we will get what we have in the settings of the currently selected company, regardless of whether other companies may not have that name. I hope that you agree with me.

Before opening dropdown the logic is to display `untitled_company` if there is no entered name:

Screenshot 2023-01-18 at 19 01 36

### DIFF
--- a/src/components/CompanySwitcher.tsx
+++ b/src/components/CompanySwitcher.tsx
@@ -149,7 +149,8 @@ export function CompanySwitcher() {
                     <DropdownElement onClick={() => switchCompany(index)}>
                       <div className="flex items-center space-x-3">
                         <span>
-                          {record.company.settings.name || t('new_company')}
+                          {record.company.settings.name ||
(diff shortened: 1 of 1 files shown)


turbo124:
@Civolilah if company.settings.name is blank, then as long as the fall back is Untitled Company we are good.

Civolilah:
> 

@turbo124 Yeah, that is the case.

beganovich (approved):
Looks good, thanks :+1:

PR #408, opened 2023-01-19
[Bugfix] Marking invoice sent success message

Civolilah:
@turbo124 We had the `notification_invoice_sent` keyword when we marked the invoice as sent, I didn't find a similar keyword for your recommendation on the Jira ticket (:invoice marked sent...). I search everything with the word :invoice, but there is no good keyword for our purpose. But I found `marked_invoice_as_sent` which I believe gives us exactly what we need, message: `Successfully marked invoice as sent`.


turbo124:
looks good!

beganovich (approved):
Looks good, thanks 👍

PR #410, opened 2023-01-19
[Cleanup] Edit project client field added

Civolilah:
Here is the screenshot how edit Project page looks like with the Client field (it is disabled always):

Screenshot 2023-01-20 at 00 16 58


src/pages/projects/edit/Edit.tsx
```diff
@@ -65,6 +65,20 @@ export function Edit() {
...
+          endpoint="/api/v1/clients"
+          label="display_name"
+          onChange={(value) => handleChange('client_id', value.value)}
+          clearButton={Boolean(project?.client_id)}
+          onClearButtonClick={() => handleChange('client_id', '')}
+          errorMessage={errors?.errors.client_id}
+          queryAdditional
+          disabled
+        />
+      </Element>
```
beganovich, line 80: We have `ClientSelector` component in the app. Please make use of it.

beganovich (changes requested):

src/pages/projects/edit/Edit.tsx
Civolilah: It's replaced!

beganovich (approved):
Looks good, thanks 👍

PR #411, opened 2023-01-20
[Cleanup] Email Back/Send Email actions refactoring

Civolilah:
Here are screenshots of refactor UI when Send emails for Quote, Credit, Invoice and Purchase Order components:

Screenshot 2023-01-20 at 01 09 03

Screenshot 2023-01-20 at 01 09 23

Screenshot 2023-01-20 at 01 09 45

Screenshot 2023-01-20 at 01 10 45

@turbo124 The send email button has been moved to the top (in the navbar), we always have the option back in the navbar so the duplicate is simply removed.


beganovich:
If I am getting this correctly, the whole idea of this PR was to move send/back buttons to the top. If so, this is not a good approach. We've moved way too much logic into the parent component (from Mailer) that it's a repetition race now.

The correct way would be to keep sending logic inside of Mailer and use something like an imperative handle hook from React (https://beta.reactjs.org/reference/react/useImperativeHandle) to create and forward refs. 

Finally we should be able to do something like this:

```tsx
  return (
    <Default
      title={documentTitle}
      breadcrumbs={pages}
      onSaveClick={ref.current.send()}
      saveButtonLabel={t('send_email')}
    >
      {invoice && (
        <Mailer
          ref={ref}
          resource={invoice}
          resourceType="invoice"
          list={list}
          defaultEmail="email_template_invoice"
          redirectUrl="/invoices"
        />
      )}
    </Default>
  );
```

Civolilah:
> If I am getting this correctly, the whole idea of this PR was to move send/back buttons to the top. If so, this is not a good approach. We've moved way too much logic into the parent component (from Mailer) that it's a repetition race now.
> 
> The correct way would be to keep sending logic inside of Mailer and use something like an imperative handle hook from React (https://beta.reactjs.org/reference/react/useImperativeHandle) to create and forward refs.
> 
> Finally we should be able to do something like this:
> 
> ```tsx
>   return (
>     <Default
>       title={documentTitle}
>       breadcrumbs={pages}
>       onSaveClick={ref.current.send()}
>       saveButtonLabel={t('send_email')}
>     >
>       {invoice && (
>         <Mailer
>           ref={ref}
>           resource={invoice}
>           resourceType="invoice"
>           list={list}
>           defaultEmail="email_template_invoice"
>           redirectUrl="/invoices"
>         />
>       )}
>     </Default>
>   );
> ```

@beganovich Yes, you're right, we just have to replace the buttons. Thanks for the awesome advise, I didn't know about this hook. Now I learned it and implemented using `useImperativeHandle` hook. Let me know your thoughts.

src/pages/invoices/email/components/Mailer.tsx
```diff
@@ -40,14 +40,15 @@ export type MailerResource = Invoice | RecurringInvoice | Quote | PurchaseOrder;
...
+  ref: RefObject<HTMLInputElement | undefined>;
   resource: MailerResource;
   resourceType: MailerResourceType;
   list: Record<string, string>;
   defaultEmail: string;
   redirectUrl: string;
 }
 
-export function Mailer(props: Props) {
+export const Mailer = forwardRef<MailerComponent, Props>((props, ref) => {
```
Civolilah, line 51: I haven't found a way to implement it with function based components. All the examples I've found are with arrow-based components.

beganovich (approved):
Looks good, thanks 👍

PR #412, opened 2023-01-20
[Bugfix] Recurring expense query include (vendor and client)

Civolilah:
@turbo124 On the Client overview page, in the query for Recurring expense table, the objects Vendor and Client are also included.


beganovich (approved):
Looks good, thanks 👍

PR #414, opened 2023-01-21
[Bugfix] Clients query caching - Recurring expense pages

Civolilah:
@turbo124 @beganovich As with `Expenses`, the query from the `ClientSelector` component is cached on the Create/Edit `Recurring Expenses` pages.


beganovich (approved):
Looks good, thanks 👍

PR #416, opened 2023-01-22
[Cleanup] Translation keywords changing

Civolilah:
@turbo124 @beganovich Only one mistyped translation keyword of `view_delivery_note` to `delivery_note` has been changed here.

I additionally noticed that we don't have the `clone_to_recurring_invoice` keyword in the translation files. This is used for `Quotes` actions. Do we need to add this one or change to something else?


turbo124:
@Civolilah 

I'll add clone_to_recurring_invoice to the lang files

beganovich (approved):
Looks good 👍

PR #418, opened 2023-01-22
Project create modal

Civolilah:
Here is the screenshot of implemented solution for adding projects using modal:

Screenshot 2023-01-22 at 15 48 05


src/pages/projects/common/components/CreateProjectModal.tsx
```diff
@@ -0,0 +1,229 @@
...
+    >
+      {project ? (
+        <>
+          <Element leftSide={t('project_name')}>
+            <InputField
+              value={project.name}
+              onValueChange={(value) => handleChange('name', value)}
+              errorMessage={errors?.errors.name}
+            />
+          </Element>
```
beganovich, line 141: `<Element>`  is meant for usage in cards, not modals. This way we end up with double padding, causing elements to not align.

beganovich (changes requested):
Looks good, just drop the element and use the modal's native wrapping.

src/pages/projects/common/components/CreateProjectModal.tsx
Civolilah: In that case, I removed the `Element` component and the name of the filed is managed with the `label` and `inputLabel` props. But the whole UI doesn't look good with only one column IMO. So, I separated it into two columns of 4 fields. The description has been updated with a screenshot of the new UI. Let me know your thoughts.

beganovich:
@turbo124 thoughts?

turbo124:
@beganovich LGTM!

beganovich (approved):
Looks good, thanks 👍

PR #422, opened 2023-01-22
[Cleanup] Download PDF - File name

Civolilah:
@beganovich @turbo124 The file name for the downloaded PDF file has been changed to the file name we get through the headers instead of just using the entity name.


beganovich (approved):
Looks good, thanks 👍

PR #427, opened 2023-01-23
[Cleanup] Bank account selector

Civolilah:
@turbo124 The bank account selector has been improved and now, after creation via the modal, the newly created bank account will be immediately selected in the selector. 

One important thing to mention. I refactored the code a bit and used `bank_integrations/create` end point for the first time. When we try to create a bank account (POST method `bank_integrations`) we get an error because of the `nickname` property in this object. If I remove it everything works perfectly. So to make everything work I removed this property before saving the new bank account. But is there any way to fix this on the API side and make it work with the `nickname` property, same as for the PUT method?

@beganovich I've included a few more files in this PR, not just the two main files. The reason for this is the slightly unconventional interface of `BankAccount`. To make it conventional with the rest of the app, I decided to make changes to this PR. I think this is a specific case because if we want to split it into two PRs, we need to wait for the first one to merge. I hope you agree with me.

Let me know your thoughts.

### DIFF
--- a/src/common/interfaces/bank-accounts.ts
+++ b/src/common/interfaces/bank-accounts.ts
@@ -10,30 +10,20 @@
 
 export interface BankAccount {
   id: string;
-  bank_account_name: string;
-  bank_account_type: string;
-  balance: number;
-}
-
-export interface BankAccountInput {
-  bank_account_name: string;
-}
-
-export interface BankAccountDetails {
-  id: string;
   archived_at: number;
   auto_sync: boolean;
-  bank_account_name: string;
-  bank_account_type: string;
   balance: number;
   bank_account_id: number;
+  bank_account_name: string;
+  bank_account_number: string;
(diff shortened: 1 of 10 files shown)


turbo124:
@Civolilah looks good. I've fixed the API issue with nickname :+1:

Civolilah:
> @Civolilah looks good. I've fixed the API issue with nickname 👍

@turbo124 I just tried to create bank account without deleting `nickname` property, but it does not work for me, API returned 500 error. Can you please just double check it one more time?

turbo124:
I've fixed it in the repo, but not pushing to production yet.

Civolilah:
> 

@turbo124 Ah, okay, in that case, I've removed unnecessary parts of code for deleting `nickname` property.

beganovich (approved):
Looks good, thanks 👍

PR #428, opened 2023-01-24
[Feature] Merge client

Civolilah:
Here is the screenshot of implemented solution for merging Clients:

Screenshot 2023-01-24 at 14 59 48


src/components/DataTable.tsx
```diff
@@ -354,9 +346,7 @@ export function DataTable<T extends object>(props: Props<T>) {
                       {props.customActions &&
                         props.customActions?.map(
                           (action: any, index: number) => (
-                            <React.Fragment key={index}>
-                              {action(resource)}
-                            </React.Fragment>
+                            <div key={index}>{action(resource)}</div>
```
Civolilah, line 349: The reason for changing React.Fragment to a div HTML element is that React.Fragment doesn't have any events like onClick, but we really need it here. We need to close the dropdown after clicking on a particular option. In our case, if we don't do this, the modal and dropdown will open at the same time, which is very bad behavior IMO.

src/pages/clients/edit/components/Contacts.tsx
```diff
@@ -73,7 +73,7 @@ export function Contacts(props: Props) {
         <div key={index} className="pb-4 mb-4 border-b">
           <Element leftSide={t('first_name')}>
             <InputField
-              id="first_name"
+              id={`first_nam_${index}`}
```
Civolilah, line 76: If we have more than one contact assigned to a particular client, we will get an error in the console that our elements do not have unique IDs in the DOM. It first appeared when I opened the PassowordConfirmation modal.

turbo124:
looks really good :+1:

src/components/dropdown/Dropdown.tsx
```diff
@@ -59,8 +59,17 @@ export function Dropdown(props: Props) {
...
-                ? cloneElement(child, { setVisible, key: index })
+              child &&
+              (child['type'] == DropdownElement || child['type'] == 'div')
+                ? cloneElement(child, {
+                    ...(child['type'] == DropdownElement && {
+                      setVisible,
+                    }),
+                    ...(child['type'] == 'div' && {
+                      onClick: () => setVisible(false),
+                    }),
```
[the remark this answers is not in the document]
beganovich: Is it possible to extract this into separate components or something where we can do clean conditionals? These inline things look really bad IMO and we sacrifice a lot of readability for the sake of another component.

src/pages/clients/edit/components/Contacts.tsx
beganovich: A subtle change, but `first_nam` is a typo.

beganovich (changes requested):
Looks good, few changes needed 👍

src/pages/clients/edit/components/Contacts.tsx
Civolilah: Resolved!

src/components/dropdown/Dropdown.tsx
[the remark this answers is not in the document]
Civolilah: Tbh, I've seen many different examples where people conditionally assign properties like this and so I'm doing it like this.

For your recommendation of separating the components I think it is really unnecessary. I just created a function that will return those properties cleaner and that's what we need here. Let me know what you think about this kind of implementation.

beganovich (approved):
Looks good, thanks 👍

PR #429, opened 2023-01-24
[Cleanup] Api Token Pages - Refactoring

Civolilah:
@beganovich @turbo124 This PR including refactoring for all `Api token` pages, including the table page. We had an old version of the table without some functionality, so I refactored it, but some things were additionally connected, so this page needed to be refactored. Except that, Create/Edit pages are refactored with the current project convention which includes replacing for action buttons and a more-actions drop-down menu. Let me know your thoughts.

Here are screenshots:

Screenshot 2023-01-25 at 00 26 13

Screenshot 2023-01-25 at 00 26 23

Screenshot 2023-01-25 at 00 26 40


turbo124:
@Civolilah 

Look good, could we please make the API token itself hidden with ********* and just use the copytoclipboard functionality please

Civolilah:
> @Civolilah
> 
> Look good, could we please make the API token itself hidden with ********* and just use the copytoclipboard functionality please

@turbo124 Sure, here is the screenshot how it looks now:

Screenshot 2023-01-26 at 02 22 13

So the user always sees only the `*` characters, but when he clicks the copy to clipboard function, the real value will be copied to the board.

turbo124:
LGTM!

beganovich (approved):
Looks good, thanks 👍

PR #431, opened 2023-01-25
[Cleanup] Commented Decimal Comma Option - Localization

Civolilah:
@turbo124 @beganovich As David mentioned in the Jira ticket, I had to remove the `decimal_comma` on the localization settings page. That's why I just commented out this part of the code, so that if we decide to bring it back, we can just uncomment it.


beganovich (approved):
Looks good, thanks 👍

PR #434, opened 2023-01-26
[Feature] Convert currency option - Expense pages

Civolilah:
@turbo124 Currently, PR contains all parts of the functionality except the comparison of two selected currencies. We just need to talk about this part further and include the logarithm to calculate it.


turbo124:
@Civolilah the logic already exists in the application for the comparison of the currencies, however the rounding needs to be fixed.

turbo124:
When the company option to convert currency is not selected, when i manually toggle the Convert Currency toggle in an expense, the toggle never persists afterwards.

you may need to add a check to see if a converted amount is set on the expense. If yes, then it means the toggle should be on for that record so that the user can see the amount.

Civolilah:
> 

@turbo124 The logic has been improved so that now if we have any `foreign_amount` on a particular expense, the `convert_currency` fields will show if they are turned off in the settings.

turbo124:
@Civolilah 

Sorry, this still is not working for me. It looksl ike we set the foreign currency on save regardless. It appears that we monitor invoice_currency_id if a conversion has been set by the user.

Civolilah:
> 

@turbo124 As we discussed on the Slack this behaviour is also improved and on my end it looks good. Let me know about your thoughts.

turbo124:
LGTM!

beganovich (approved):
Looks good, thanks 👍

PR #435, opened 2023-01-27
[Feature] Unlock pro button

Civolilah:
Here is a screenshot of the implemented solution for the unlock_pro/upgrade button:

Screenshot 2023-01-27 at 08 06 56

@turbo124 So the button will look like this on different plans and for different types of users:

- The button will be displayed if the user is `Hosted` and if their plan is `freePlan`, in this case the navigation will be to `user?.company_user?.ninja_portal_url` and the button text will be `unlock_pro`. We don't have this keyword in the translation files, so if you agree with this one, just add it to the translation files.

- The button will be displayed if the user is `SelfHosted`, in this case the navigation will be to the `whiteLabelLink` from the `env` files. The text for this one case will be the `upgrade` keyword found in the screenshot. But I have one case that we haven't specified yet. If a `self-hosted` user buys a `whiteLabel`, IMO we need to hide this button. I'm not sure I know how to check if the `whiteLabel` has been purchased. So let me know that way if you agree with me that we don't need to show the button if it's already purchased.


turbo124:
@Civolilah I've added the translation

Civolilah:
@turbo124 So, as we discussed on the Slack the logic of showing this button is a bit changed. It looks like this:

- If the user is `HOSTED` and has a `free plan`, the button will be displayed and the navigation after clicking the button will be to `user?.company_user?.ninja_portal_url`
- If the user is `SELF-HOSTED` and does not have a `white_label plan` `OR` has a `white_label plan` but it has `expired`, the button will be displayed and the navigation after clicking will be to `import.meta.env.VITE_WHITELABEL_INVOICE_URL`

Let me know your thoughts.

turbo124:
@Civolilah this is correct, however for hosted you also need to inspect the plan_expires, if expired, we need  to display the CTA.

Civolilah:
> @Civolilah this is correct, however for hosted you also need to inspect the plan_expires, if expired, we need to display the CTA.

@turbo124 Yeh, that part is also updated. So now for `HOSTED` users. If he has a `freePlan` `OR` if he has another plan but `has expired`, a button will be shown.

turbo124:
LGTM

src/components/layouts/Default.tsx
```diff
@@ -58,11 +62,25 @@ interface Props extends CommonProps {
...
+  const account = useCurrentAccount();
+
+  const shouldShowUnlockButton =
+    (isHosted() &&
+      (freePlan() ||
+        (account?.plan && new Date(account?.plan_expires) < new Date()))) ||
+    (isSelfHosted() &&
+      ((account?.plan && new Date(account?.plan_expires) < new Date()) ||
+        !account?.plan));
+
```
beganovich, line 77: Is it possible to clean up this a bit? Perhaps extract it into a hook or something with simple ifs, it's really unreadable this way.

beganovich (changes requested):

src/components/layouts/Default.tsx
Civolilah: The logic is separated into hooks.

beganovich (approved):
Looks good, thanks 👍

PR #436, opened 2023-01-27
[Bugfix] Reactivity tables

Civolilah:
@turbo124 @beganovich Here we only had an error with invalidating queries from the tables and that's why we didn't have the tables populated after using some actions from the dropdown menu. That's fixed now and I think I've covered the whole app with this one, but let me know if I've missed anything.


turbo124:
@Civolilah 

still not fixed for me, when i perform actions from the tables, such as archive, the rows persist on screen.

turbo124:
@Civolilah looks really good!

beganovich (approved):
Really great solution, thanks 👍

PR #437, opened 2023-01-28
[Feature] Lock Invoiced Tasks - Field Implementation

Civolilah:
Here's a screenshot of the Lock Invoiced Tasks implementation implemented under task settings:

Screenshot 2023-01-28 at 01 39 45

@turbo124 This field is based on the `off/on` string. Handling this field like this requires additional adjusting on the front end. I guess we don't have any special states of this field, so if that's correct, I suggest you cast this to a boolean value. Also, the only property I found in the company settings based on the lock keyword is `lock_invoices`, I'm not 100% sure I'm on the right track, so please let me know if I'm wrong. Also, tell me what you think.

### DIFF
--- a/src/pages/settings/task-settings/TaskSettings.tsx
+++ b/src/pages/settings/task-settings/TaskSettings.tsx
@@ -47,7 +47,7 @@ export function TaskSettings() {
       })
     );
 
-  const handleToggleChange = (id: string, value: boolean) =>
+  const handleToggleChange = (id: string, value: boolean | string) =>
     dispatch(
       updateChanges({
         object: 'company',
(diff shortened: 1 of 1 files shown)


turbo124:
@Civolilah 

The settings property for this is:

company.invoice_task_lock

type = bool, you may need to update the interfaces.

Civolilah:
> @Civolilah
> 
> The settings property for this is:
> 
> company.invoice_task_lock
> 
> type = bool, you may need to update the interfaces.

@turbo124 Of course this part has been changed and now I use that property you mentioned. Also, as we discussed the Slack I hide Edit action in the Task Table dropdown for invoiced tasks when this property is turned on.

turbo124:
@Civolilah 

Please ensure the start option is not shown when locked

image

also whilst the task can be viewed, the save button / action items must be disabled

image

Civolilah:
> @Civolilah
> 
> Please ensure the start option is not shown when locked
> 
> image
> 
> also whilst the task can be viewed, the save button / action items must be disabled
> 
> image

@turbo124 Those Start/Stop functionalities are removed from dropdown menus if the task is invoiced.

beganovich (approved):
Looks good, thanks 👍

PR #438, opened 2023-01-30
[Feature] Auto start task

Civolilah:
@turbo124 As we know we already have `auto_start_tasks` field under `Task settings` but we didn't have that feature. So, I implemented it here. If this setting is on, after creating a task we will call the API to run this task and after that the user will be navigated to the edit page of that created task. With this setting the task will be created without running. Let me know your thoughts.


turbo124:
LGTM!

beganovich (approved):
Looks good, thanks 👍

PR #439, opened 2023-01-30
[Feature] Emailing Statement

Civolilah:
Here is the screenshot of implemented solution for emailing statement for Client entity:

Screenshot 2023-01-30 at 19 24 11

@turbo124 I refactored the UI a bit to match the rest of the project, for the others we have a drop down menu in the top right corner so I've done the same here. Let me know if you disagree with me for refactoring. Apart from the staging endpoint we are getting a 500 error on this page, I guess endpoints for this page are not available on this staging API, but I have everything set up using the `main Flutter app https://invoicing.co/`. So I hit this endpoint `'/api/v1/client_statement?send_email=true'`, with a statement object in the body. If we get a status code of 200, the toaster will be displayed with the message from the response, `response.data.message`. If client does not have any contacts with the email we will show toaster error message using `client_email_not_set` keyword same as for Flutter app.

I used `email_statement` keyword for label, we do not have this one in translation files. If you agree with this one, please add it to the translation files.

Let me know your thoughts.


turbo124:
looks good, can we please change the menu text to just email please

Civolilah:
> looks good, can we please change the menu text to just email please

@turbo124 Sure, it has been changed!

turbo124:
LGTM!

src/pages/clients/statement/Statement.tsx
```diff
@@ -102,6 +114,11 @@ export function Statement() {
   const iframeRef = useRef<HTMLIFrameElement>(null);
   const [selectRange, setSelectRange] = useState<string>('last_7_days');
 
+  const [client, setClient] = useState<Client>();
+
+  const [shouldEmailStatement, setShouldEmailStatement] =
+    useState<boolean>(false);
```
beganovich, line 120: There's no need for this, you can just break the send action when sending and return early with toast.

src/pages/clients/statement/Statement.tsx
```diff
@@ -146,6 +163,35 @@ export function Statement() {
     toast.dismiss();
   };
 
+  const handleSendEmail = () => {
+    toast.processing();
+
```
beganovich, line 168:
```tsx
const send =  client.contacts?.some((contact) => contact.email) || false;

if (!send) {
    return toast.error('client_email_not_set');
}

// ..
```

beganovich (changes requested):

src/pages/clients/statement/Statement.tsx
Civolilah: Yeah, agree with you. The logic has been changed!

src/pages/clients/statement/Statement.tsx
Civolilah: Yeah, agree with you. The logic has been changed!

beganovich (approved):
Looks good, thanks 👍

PR #441, opened 2023-01-30
[Cleanup] Google analytics / Matomo

Civolilah:
Here is the screenshot of `Analytics ` page after updates:

Screenshot 2023-01-31 at 00 17 33

@turbo124 @beganovich Regarding the request in the Jira ticket to change `google_analytics` to `analytics`, I changed it for all things like urls etc. Except those changes I had to add `Matomo` fields so it was added. Now, we just missed `analytics` as keyword from translation files so please just add it. Let me know your thoughts.


turbo124:
@Civolilah I'll add in the translation keywork `analytics` if you can update the pr to match that string please :)

beganovich (approved):
Looks good, thanks 👍

PR #444, opened 2023-02-01
[Feature] Invoice task functionality

Civolilah:
@turbo124 Actually, we already implemented this functionality, but we just missed some parts about calculating `hours, rates and line_total`. I hope I've covered everything we need for this functionality, but let me know if I missed any part.


turbo124:
Looks good, however i don't think the hours are calculating correctly

image


image

It looks like if the last end date does not have a value, the system cannot correctly calculate the duration.

Also i see rounding issues

image

Civolilah:
> Looks good, however i don't think the hours are calculating correctly
> 
> image
> 
> image
> 
> It looks like if the last end date does not have a value, the system cannot correctly calculate the duration.
> 
> Also i see rounding issues
> 
> image

@beganovich @turbo124 We had a few bugs here, we didn't have something that we have on the Flutter app so they were implemented. So all the weird behaviors are fixed. I tried my best to make the code as readable as possible, but there are so many scenarios.

turbo124:
@Civolilah 

I'm seeing issues still.

For example, create a new task, and then click start.

The start time does not populate.

Then add new item and enter in a duration.

End time populates, but start time does not.

If i exit and then from the table view go into the task, the durations are not populated.

Also, if a task has been invoiced once, it cannot be reinvoiced. So if task.invoice_id != null then don't show that option.

turbo124:
LGTM!

src/pages/invoices/common/hooks/useFormatMoney.ts
```diff
@@ -91,6 +91,6 @@ export function useFormatMoney(props: Props) {
       return Number.formatMoney(value, currency, country);
     }
 
-    return value;
+    return parseFloat(value.toString()).toFixed(2);
```
beganovich, line 94: @turbo124 is this okay? We are hardcoding the decimal rounding to 2nd **all the time**?

src/pages/invoices/create/Create.tsx
```diff
@@ -163,7 +163,7 @@ export function Create() {
           onChange={(id) => handleChange('client_id', id)}
           onClearButtonClick={() => handleChange('client_id', '')}
           onContactCheckboxChange={handleInvitationChange}
-          readonly={searchParams.get('table') === 'tasks'}
+          readonly={searchParams.get('project') === 'true'}
```
beganovich, line 166: Why?

src/pages/tasks/common/helpers.ts
```diff
@@ -24,17 +24,101 @@ export function parseTime(timestamp: number) {
...
+
+  const updatedHours = Number(hours) === 12 ? '00' : hours;
+
+  return (
+    updatedHours.toString() +
+    ':' +
+    minutes.toString() +
+    ':' +
+    seconds.toString()
+  );
```
beganovich, line 54: Can you elaborate on this? It's a kinda big change so I'd like to know the circumstances which lead to this change.

src/pages/tasks/common/helpers.ts
```diff
@@ -24,17 +24,101 @@ export function parseTime(timestamp: number) {
...
+  if (Number(seconds) < 10) {
+    seconds = '0' + seconds.toString();
+  }
+
+  if (stop && startDateValue !== '1970-01-01') {
+    return hours + ':' + minutes + ':' + seconds;
+  } else {
+    return '00:00:00';
+  }
 }
```
beganovich, line 122: Same for this.

src/pages/tasks/common/hooks.tsx
```diff
@@ -367,7 +367,7 @@ export function useActions() {
       !task.invoice_id && (
         <DropdownElement
           onClick={() => invoiceTask([task])}
-          icon={<Icon element={MdTextSnippet} />}
+          icon={<Icon element={BiPlusCircle} />}
```
beganovich, line 370: Why? We are trying to stick either to Feather Icons or Material for the UI.

src/pages/tasks/common/hooks/useStop.ts
```diff
@@ -24,6 +27,18 @@ export function useStop() {
   return (task: Task) => {
     toast.processing();
 
+    const logs = parseTimeLog(task.time_log);
+
+    if (logs[logs.length - 1][0] && logs[logs.length - 1][0] > dayjs().unix()) {
+      logs[logs.length - 1][1] = logs[logs.length - 1][0] + 1;
+
+      task.time_log = JSON.stringify(logs);
```
beganovich, line 35: This really looks like 🍝. Is it possible to simplify & perhaps update naming?

src/pages/tasks/edit/components/Actions.tsx
```diff
@@ -78,7 +78,7 @@ export function Actions(props: Props) {
         {!isTaskRunning(task) && !task.invoice_id && (
           <DropdownElement
             onClick={() => invoiceTask([task])}
-            icon={<Icon element={MdTextSnippet} />}
+            icon={<Icon element={BiPlusCircle} />}
```
beganovich, line 81: Why?

src/pages/invoices/common/hooks/useFormatMoney.ts
Civolilah: @turbo124 @beganovich Just to mention from my end. This is the case when we don't have currency and country, so we don't have separators and similar things to round the value. I'm not sure we have any additional information here to round it dynamically. So that's the reason for hard coding, otherwise without a round UI it really doesn't look good.

src/pages/invoices/create/Create.tsx
Civolilah: David and I had a lot of discussions on Slack about this functionality, so we realized that we always made this selector read-only if we invoiced a task. So David recommended that I make this selector read only when the task has a project selected. It should override all client_id values. So I pass it as a URL parameter and check it here.

src/pages/tasks/common/helpers.ts
Civolilah: The reason for changing this is actually the time format we need on our time pickers. We need hours in the format 00-23, but we had 00-12. So that's the reason for the change. But, I went to the package documentation and found that for this change we only need to change the string from `hh` to `HH`. So now that's the only change needed.

src/pages/tasks/common/hooks.tsx
Civolilah: Actually, I thought the new icon was better for UX, but as we discussed today on Slack for type of icons that we have to use and also after reviewing I think `MdTextSnippet` could be even better for UX. So I went back to the old version. Let me know your thoughts.

src/pages/tasks/common/hooks/useStop.ts
Civolilah: It's actually pretty simple, but I'm sorry I missed naming. It is now aptly named and I think there is no more confusion.

src/pages/tasks/common/helpers.ts
Civolilah: We had a few bugs with duration calculation, especially when we turn on the `end date` and then turn it off in the settings. So, I had to adjust the entire checkout process. I researched the `dayjs` package deeply and found a simpler code solution for this purpose. Now I did the same thing, but with clearer code. Let me know your thoughts.

src/pages/invoices/common/hooks/useFormatMoney.ts
turbo124: @beganovich @Civolilah no, we should never hardcode the rounding, it shuld always use the formatMoney() method.

src/pages/invoices/create/Create.tsx
turbo124: @beganovich this is correct. Selecting the project will always override any other selector that contains a reference to a client_id.

src/pages/invoices/common/hooks/useFormatMoney.ts
Civolilah: @turbo124 @beganovich We talked on Slack about this scenario and I couldn't replicate it. So I just changed the code to the old version where we don't have this manual fix. If we catch the scenario through testing in the future, it will be a minor job to fix it.

beganovich (approved):
Looks good to me :+1:

PR #448, opened 2023-02-03
[Cleanup] Invoice design page - tabs

Civolilah:
Here are the screenshots of implemented solutions for Invoice Design tabs:

Screenshot 2023-02-03 at 10 16 26

Screenshot 2023-02-03 at 10 30 33

@beganovich @turbo124 The card description states that the tabs are FROM the `Client details` card to the `last one`. I did both solutions, to make all the cards into tabs and to do the `General settings card` at the top and just change the tabs below that. IMO, as we've done for the rest of the app and the Flutter app, it's better for the UX to turn all cards into tabs. The second screenshot refers to the solution when we have a tab navigation bar above the `General Settings card` which is fixed. Below this fixed card will be card selected from the navigation tab bar. Currently on the PR is implemented solution of making all cards into tabs. Let me know your thoughts.


beganovich (approved):
Looks good, thanks 👍

PR #450, opened 2023-02-03
[Cleanup] Currency registration field and fixing bug on registration card

Civolilah:
Here is the screenshot of UI with added `currency` field:

Screenshot 2023-02-03 at 18 41 15

@beganovich @turbo124 Besides just adding this field, I noticed that updating the company field `client_registration_fields` doesn't work. So this PR includes that part as well, I think it works correctly now and the API will get an array with objects that include the `field key` and the `required (boolean)`. Let me know your thoughts.


turbo124:
Thanks!

beganovich (approved):
Looks good, thanks👍

PR #451, opened 2023-02-03
[Cleanup] Hiding Custom CSS and JS fields

Civolilah:
@beganovich @turbo124 I've added condition for showing Custom CSS and JS fields. So, if user is SELF-HOSTED it will be shown otherwise it will not be available.


beganovich (approved):
Looks good, thanks 👍

PR #452, opened 2023-02-03
[Cleanup] Send time helper text

Civolilah:
@turbo124 @beganovich Here is the screenshot of UI with removed helper text for `send_time` field:

Screenshot 2023-02-04 at 00 31 17


beganovich (approved):
Looks good, thanks 👍

PR #457, opened 2023-02-06
[Cleanup] Payment template variables

Civolilah:
@turbo124 @beganovich For some templates we have different variables. One of them is `Payment` template so I changed variables when this one is selected:

Screenshot 2023-02-07 at 00 35 06

@turbo124 I copied variables from the Flutter app. Except this one I noticed that we maybe have wrong variables for some other templates. Should I resolve it on this PR or create another one?


turbo124:
@Civolilah not sure I follow, if you are happy that this PR covers the Jira ticket, lets merge it and then we can discuss and open a new issue with the wrong variables.

beganovich (approved):
Looks good, thanks 👍

PR #468, opened 2023-02-13
[Cleanup] Client pages - More actions bugs - Status badge adding

Civolilah:
Here's a screenshot of the UI when the status badge is added and the actions are moved to the navigation bar as for the rest of the project:

Screenshot 2023-02-18 at 15 48 55

@turbo124 @beganovich PR includes adding a `status badge` to the details tab on the client's overview page. In addition, any bugs that were found related to `more actions` for client entity have been resolved. Also, I moved more actions to the navigation bar to the right of the `Back` button.

Also, my recommendation is to remove `Edit Client` button and move it to the more actions dropdown. IMO, the UI will be more clear. What do you think?

Let me know your thoughts.


src/pages/clients/create/Create.tsx
```diff
@@ -59,8 +59,8 @@ export function Create() {
   });
 
   useEffect(() => {
-    if (blankClient?.data.data) {
-      setClient(blankClient.data.data);
+    if (blankClient) {
+      setClient(blankClient.data?.data);
     }
```
Civolilah, line 64: When I go to the client creation page, the whole application crashes. The reason for this is reading data from an undefined value. I recommend that you redo this part in the same way as for the rest of the project using `useBlankClientQuery`. Let me know your thoughts.

turbo124:
@Civolilah I think for client, the edit button is a high traffic item and probably can be left stand alone. It also matches the UX of v4.

Civolilah:
> @Civolilah I think for client, the edit button is a high traffic item and probably can be left stand alone. It also matches the UX of v4.
> 
> Also, we'll want to only allow Restore/Purge more actions when a client is deleted. No other actions should be possible.

@beganovich @turbo124 If Ben can leave some advice for this. Actually, our component is ready to add items to the right side. David, do you mean to put the Edit Client button to the right or below the breadcrumbs? If you want to move the Edit Client button to the left, I'll probably have to change the main component a bit. That is the real reason for seeking advice.

Also, I changed the conditions for displaying actions in the dropdown menu. Now, if the client is deleted, the only actions available will be Restore and Purge.

Let me know your thoughts.

turbo124:
@Civolilah if the actions are being moved into the nav bar, the we can nest the edit button in there, so no need for the additional edit button.

Civolilah:
> @Civolilah if the actions are being moved into the nav bar, the we can nest the edit button in there, so no need for the additional edit button.

@beganovich @turbo124 Me and David were talking on Slack about changing the position for the Edit Client button. So the final version of the current code implementation is this:

Screenshot 2023-02-18 at 15 54 57

The button has been moved to the navbar. Let me know your thoughts.

src/components/DataTable.tsx
```diff
@@ -386,6 +387,14 @@ export function DataTable<T extends object>(props: Props<T>) {
                           {t('delete')}
                         </DropdownElement>
                       )}
+
+                      {props.customActions &&
+                        props.customActions.map(
+                          (action: any, index: number) =>
```
beganovich, line 393: Can we try to typehint this? Code without `any` would be nice.

src/pages/clients/create/Create.tsx
beganovich: This page doesn't use /client/create query?

beganovich (approved):
Looks good, thanks :+1:

src/components/DataTable.tsx
Civolilah: I tried to make it as close as possible. Now, type is created. Let me know if you agree with this.

beganovich (approved):
Awesome, thanks :+1:

PR #469, opened 2023-02-13
[Bugfix] Create Bank Account - Permission handling - Invoice Design Optional Chaining

Civolilah:
@beganovich @turbo124 PR includes a bug fix on the invoice design page, just added optional chaining so they don't break the app if we don't have that property. The second part of the PR is to add a restriction to disable the create bank account button. So, if the user has a pro or enterprise plan, he will be able to create a bank account, otherwise the save button will be disabled.


beganovich (approved):
Looks good, thanks 👍

PR #470, opened 2023-02-14
[Cleanup] Empty columns - Helper text

Civolilah:
Here is the screenshot with added helper text for `hide_empty_columns_on_pdf` company settings property:

Screenshot 2023-02-14 at 00 59 32

@beganovich @turbo124 When property has set to true text should be Hide but when it is set to false, text should be Show. Let me know your thoughts.

### DIFF
--- a/src/pages/settings/invoice-design/components/GeneralSettings.tsx
+++ b/src/pages/settings/invoice-design/components/GeneralSettings.tsx
@@ -20,7 +20,7 @@ import { ChangeEvent } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useDispatch } from 'react-redux';
 import { Card, ClickableElement, Element } from '../../../../components/cards';
-import { SelectField } from '../../../../components/forms';
+import { Radio, SelectField } from '../../../../components/forms';
 import Toggle from '../../../../components/forms/Toggle';
 
 export function GeneralSettings() {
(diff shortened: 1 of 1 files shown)


beganovich:
This is a mistake and confusing. We are controlling if we want to hide empty columns on PDF.

1. When the switch is off, it shows SHOW which implies we want to turn on showing, however, once we turn it on it shows that we turned on hiding.
2. Dynamic label with only hide/show is a pitfall and not a pattern for the app.

Proposed solutions:
1. Update labels: Hide empty columns on PDF: Toggle
2. Keep current label: Empty columns with a select box showing Show, Hide as in Flutter.

I am more of a #2 guy. Let me know your thoughts. @turbo124 @Civolilah

Civolilah:
> 

@beganovich @turbo124 Yes, absolutely. Same IMO, looks a bit confusing to use. My recommendation is also to change the text on the left side, instead of `Empty Columns` to show something that is more understandable with a real purpose. For example, for `task end date` we have `Show Task End Date with help text`. An actual text change can help us keep this toggle solution. For example, if we say `Exclude Empty Columns on PDF` on the field label, this will allow us to keep the exact same logic. But I absolutely agree with Ben's second solution. Let me know your thoughts.

turbo124:
#2 works for me

beganovich:
2-1 for option 2. Sorry, @Civolilah, democracy 🤣

Civolilah:
> 

@beganovich I tried 🤣. It will be implemented with option #2.

Civolilah:
@beganovich @turbo124 The UI/UX for handling `hide_empty_columns_on_pdf` property has been changed and now it looks like this:

Screenshot 2023-02-17 at 18 37 38

Let me know your thoughts.

turbo124:
lgtm!

beganovich (approved):
Looks good, thanks :+1:

PR #471, opened 2023-02-14
[Bugfix] Translating custom label keyword

Civolilah:
Here is the screenshot of updated UI for Localization page:

Screenshot 2023-02-14 at 02 03 07

@turbo124 @beganovich Now, the left side (keyword for custom field) will be translated instead of just showing keyword on the UI.


beganovich (approved):
Looks good, thanks 👍

PR #475, opened 2023-02-17
[Cleanup] Link To GitHub - Variables

Civolilah:
Here's a screenshot of the current UI for the Generated Numbers tabs:

Screenshot 2023-02-17 at 15 35 42

@turbo124 The plain text on the screenshot has been replaced with `click_to_variables` translation keyword. If you agree with me, please just add it to translation files.

@beganovich Just to mention, we had one unnecessary card component nested on all the cards. IMO it made the bottom of each card look a little weird. So, I removed them and set Dividers. IMO, it looks better now.

Let me your thoughts.


turbo124:
@Civolilah 
looks good!

beganovich (approved):
Looks good, thanks 👍

PR #476, opened 2023-02-18
[Feature] Emailing statement - action

Civolilah:
Here is the screenshot of impelemented solution for emailing statement with default settings:

Screenshot 2023-02-18 at 15 36 28

@turbo124 @beganovich PR involves sending statements via email with default settings. This means the date range is last 7 days, show_aging_table is true, show_payments_table is true and status is 'all'. This PR will be in draft until we merge some PRs where I have already worked on actions for the Client entity. So, to avoid mistakes, we'll merge PRs that have already passed QA. Also, we don't have the `email_statement` translation keyword in the files, if you agree with this, just add it to the files. Let me know your thoughts.

### DIFF
--- a/src/pages/clients/common/hooks/useActions.tsx
+++ b/src/pages/clients/common/hooks/useActions.tsx
@@ -8,15 +8,20 @@
  * @license https://www.elastic.co/licensing/elastic-license
  */
 
+import { AxiosError } from 'axios';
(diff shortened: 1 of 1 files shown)


turbo124:
@Civolilah the email statement action needs to be from the Statement view context. From there we will have all of the data that we need (date ranges and other options)

PR #481, opened 2023-02-20
[Feature] Postmark / Mailgun - Email providers

Civolilah:
Here is the screenshot of implemented solution for email providers (Postmark and Mailgun):

Screenshot 2023-02-20 at 17 34 13

@beganovich @turbo124 I just noticed one thing that we don't have `gmail` as a translated keyword, but `microsoft` was in the translation files. Should we have `gmail` and `microsoft` as translated keywords or can we put it as `plain text`. 

Also, David mentioned in the ticket that this should only be possible for Pro/Enterprise plans. The entire page (Email Settings) is available for saving only if the user has one of those plans. That covers that part.

Let me know your thoughts.

### DIFF
--- a/src/pages/settings/email-settings/EmailSettings.tsx
+++ b/src/pages/settings/email-settings/EmailSettings.tsx
@@ -104,10 +104,10 @@ export function EmailSettings() {
             <option defaultChecked value="default">
               {t('default')}
             </option>
-            <option value="gmail">{t('gmail')}</option>
-            <option value="microsoft">{t('microsoft')}</option>
-            <option value="client_postmark">{t('postmark')}</option>
-            <option value="client_mailgun">{t('mailgun')}</option>
+            <option value="gmail">Gmail</option>
+            <option value="microsoft">Microsoft</option>
+            <option value="client_postmark">Postmark</option>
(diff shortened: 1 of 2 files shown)


src/pages/settings/email-settings/EmailSettings.tsx
```diff
@@ -94,6 +94,56 @@ export function EmailSettings() {
...
+              handleChange('settings.email_sending_method', value)
+            }
+          >
+            <option defaultChecked value="default">
+              {t('default')}
+            </option>
+            <option value="gmail">{t('gmail')}</option>
+            <option value="microsoft">{t('microsoft')}</option>
+            <option value="client_postmark">{t('postmark')}</option>
+            <option value="client_mailgun">{t('mailgun')}</option>
```
beganovich, line 110: No need for translations for company names. They are same on all languges.

beganovich (changes requested):

src/pages/settings/email-settings/EmailSettings.tsx
Civolilah: Changed

beganovich (approved):
Looks good, thanks :+1:

PR #483, opened 2023-02-20
[Cleanup] Project pages 422 errors / Selector components error message

Civolilah:
Here is the screenshot of updated UI for project pages when error message is correctly mapped:

Screenshot 2023-02-20 at 20 08 16

@turbo124 @beganovich Actually, we have correctly added the error messages to the components, but only some custom selector components do not have error message prop mapped. I just fixed it and it works as it should. In one of the previous tickets, I made sure that all components add error message props. Let me know your thoughts.

### DIFF
--- a/src/components/products/ProductSelector.tsx
+++ b/src/components/products/ProductSelector.tsx
@@ -22,6 +22,7 @@ interface Props {
   onClearButtonClick?: () => unknown;
   onProductCreated?: (product: Product) => unknown;
   onInputFocus?: () => unknown;
+  errorMessage?: string | string[];
 }
 
 export function ProductSelector(props: Props) {
@@ -45,6 +46,7 @@ export function ProductSelector(props: Props) {
         onInputFocus={props.onInputFocus}
         sortBy="product_key|asc"
         withShadowRecord
+        errorMessage={props.errorMessage}
       />
 
       <ProductCreate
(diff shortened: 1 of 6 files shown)


src/components/products/ProductSelector.tsx
```diff
@@ -22,6 +22,7 @@ interface Props {
   onClearButtonClick?: () => unknown;
   onProductCreated?: (product: Product) => unknown;
   onInputFocus?: () => unknown;
+  errorMessage?: string | string[];
```
beganovich, line 25: Is this correct? I thought it's always string[]?

beganovich (commented):
Looks good, just need confirmation.

src/components/products/ProductSelector.tsx
Civolilah: Hmm, I agree with you, there should always be an array of strings. But I just checked, wherever we have `errorMessage` as a prop, we have marked it as a `string | string[]`. I think now, should we do it as the rest of the app or can we do it here because we know it's the right way? The actual reason, I marked it like this, are other `errorMessage` props (on other custom components).

beganovich: Ah okies then, I though this was the case for this PR only. 

All good.

beganovich (approved):
Looks good, thanks :+1:

PR #484, opened 2023-02-20
[Bugfix] Project query invalidation

Civolilah:
@beganovich @turbo124 PR include query invalidation for project endpoints after creation and editing. We didn't have it before and that's the reason why we didn't have updated selectors after creating/updating the projects. Let me know your thoughts.


beganovich (approved):
Looks good, thanks 👍

PR #488, opened 2023-02-22
[Cleanup] UI After Restored Company

Civolilah:
Here is the screenshot of UI when the company is successfully restored:

Screenshot 2023-02-22 at 20 50 43

@turbo124 @beganovich When the company is successfully restored, this card will be displayed instead of the default one. I used plain text just to show how it looks. The `Successfully started import` is replaced with `restore_started` and the `The process of importing data has started. It will be finished soon!` is replaced with `importing_process_started`. Let me know your thoughts.


beganovich (approved):
Looks good, thanks :+1:

PR #489, opened 2023-02-22
[Bugfix] Microsoft button centering

Civolilah:
@beganovich @turbo124 This PR only includes simply centering signup/signin buttons on Register page.


beganovich (approved):
Looks great, thanks :+1:

PR #496, opened 2023-02-25
[Cleanup] Users endpoint - without query parameter

Civolilah:
Here is the screenshot of UI when after including without query parameters to users endpoint:

Screenshot 2023-02-24 at 11 19 02

@turbo124 @beganovich As you know, after logging into the account, we must have at least one user to display in this table. But, after including this query parameter, the currently logged in user will not be listed. Let me know your thoughts.


beganovich (approved):
Looks great, thanks :+1:

PR #498, opened 2023-02-25
[BugFix] PDF Preview - Credits, Invoices, Quotes and PurchaseOrders

Civolilah:
@turbo124 @beganovich PR includes a bug fix for the incorrect display of PDF previews on edit pages. The actual bug was using `atoms` on edit pages instead of simply `useState`. Because of that mistake, we passed the wrong entity object to the API. Now, from my side, everything looks good now. The bug was reported for `invoices`, but I realized we had the same error on the `Quotes`, `Credits` and `PurchaseOrders` tables, so it is resolved on all pages. Let me know your thoughts.


turbo124:
LGTM!

PR #514, opened 2023-03-02
[Feature] Invoice project

Civolilah:
Here is the screenshot of implemented solution for invoicing project:

Screenshot 2023-03-02 at 15 55 52

@turbo124 @beganovich Before I knew we were going to have test tasks, I was almost done with this one. I'm just doing PR now. The specific things I want to mention here is that I created two possible messages while invoicing the project:

- When a user tries to invoice a project that has no assigned tasks on that project, in that case we will display an error message with the translation `no_assigned_tasks`. We can set the message to, for example, `No tasks assigned to this project.`

- When a user tries to invoice a project that has some invoices assigned, but all those tasks are already invoiced. In this case, we will display an error message with the translation of the `no_uninvoiced_tasks` keyword. We can set the message to, for example, `All tasks have already been invoiced for this project.`

I think those messages will improve UX for this functionality. 

Let me know your thoughts.


turbo124:
@Civolilah 

I tested this, the functionality for no tasks works ok, but after i invoice a project. I am still able to create more invoices for the same project. I think the best way to handle this, is to still push the user to the invoice creation page, but if there are not tasks available, then just leave the line items empty.

Civolilah:
> @Civolilah
> 
> I tested this, the functionality for no tasks works ok, but after i invoice a project. I am still able to create more invoices for the same project. I think the best way to handle this, is to still push the user to the invoice creation page, but if there are not tasks available, then just leave the line items empty.

@turbo124 As we discussed on Slack, the behavior has been improved. Now, only when there are no tasks assigned to the project, we will display a toaster error message with the translation keyword `no_assigned_tasks`. If you agree with me, just add this keyword to the translation files. Otherwise, let me know your thoughts.

turbo124:
@Civolilah I've added the translation into the system: no_assigned_tasks

src/pages/invoices/create/Create.tsx
```diff
@@ -95,7 +95,10 @@ export function Create() {
     setInvoice((current) => {
       let value = current;
 
-      if (searchParams.get('action') !== 'clone') {
+      if (
+        searchParams.get('action') !== 'clone' &&
+        searchParams.get('action') !== 'invoice'
+      ) {
```
beganovich, line 101: Naming this is wrong. The invoice page should never have the action `invoice` because that's what the default way of doing something is. 

Can you elaborate on this, please?

src/pages/projects/common/hooks/useInvoiceProject.tsx
```diff
@@ -0,0 +1,123 @@
...
+          .join('\n')
+          .trim();
+
+        invoice.line_items = [item];
+      });
+
+      setInvoice(invoice);
+
+      navigate(
+        route('/invoices/create?table=tasks&project=true&action=invoice')
```
beganovich, line 119: `action=invoice` feels so wrong. See the comment above.

Civolilah: Action name changed. I explained in the comment above.

src/pages/invoices/create/Create.tsx
Civolilah: @beganovich So the point is the same as for the `clone` action. Through `project invoicing`, we set the invoice atom to some value calculated in the `useInvoiceProject` hook. If we do not add this part of the condition we will always have an `undefined` invoice value. But I absolutely agree with the naming, I didn't get it, I renamed the action to `project_invoicing`, I hope you agree with this naming too. Let me know your thoughts.

src/pages/projects/common/hooks/useInvoiceProject.tsx
turbo124: invoice_project?

Civolilah:
@turbo124 @beganovich Just to confirm, per David's suggestion, the action has been renamed to `invoice_project`.

src/pages/projects/common/hooks.tsx
```diff
@@ -246,7 +256,49 @@ export function useActions() {
...
   };
 
+  const handleInvoiceProject = (project: Project) => {
+    toast.processing();
+
+    queryClient.fetchQuery({
+      queryKey: route('/api/v1/tasks?project_tasks=:projectId&per_page=100', {
+        projectId: project.id,
+      }),
+      queryFn: () =>
```
beganovich, line 266: Is there any reason we went with this instead of keeping consistency with rest of API calls?

beganovich (commented):
Small clarification needed before merge.

src/pages/projects/common/hooks.tsx
Civolilah: @beganovich Tbh, I research a bit about `fetchQuery` function and saw this convention on their web page. Now, I removed unnecessary parts and made convention same as the rest of the project. Changed!

beganovich (approved):
Looks good, thanks :+1:

PR #518, opened 2023-03-06
[Bugifix] Invoice client name - Transaction match slider

Civolilah:
@beganovich @turbo124 The PR fixes a bug showing `display_name` instead of the classic `Client name`. The `display_name` property will display the name of some of the contacts assigned to the `Client` if there is no entered classic `name` property. Let me know your thoughts.


beganovich (changes requested):
This needs refactoring using useMemo. Looping thru clients each time can be an expensive operation and if we can memoize it, we should.

PR #519, opened 2023-03-06
[Bugfix] Subject property - Disabling - Emailing Invoice

Civolilah:
@beganovich @turbo124 PR includes changes to the condition of disabling the `subject` property on the emailing invoice page. So far we've disabled it if the user only has `freePlan()`, but we haven't covered that this field should be enabled for `SELF-HOSTED` users. So I just added that part of the condition. Let me know your thoughts.


turbo124:
Looks good, we'll want to merge this to main once this has been merged so that we can tag and release.

beganovich (approved):
Looks good, thanks :+1:

PR #521, opened 2023-03-08
Playwright - Latest

Civolilah:
@beganovich @turbo124 The PR is based on Ben's PR where he replaced the Cypress package with Playwright and wrote several main tests such as Invoices and Login. So, I pulled back his changes and created the rest of the basic tests for the main entities. What I noticed is the `vendor creation test`. When we add the `create_vendor` permission to the `permissions@example.com` account, this user still cannot create a vendor (API returns 401). I also tested manually. All other entities are also tested on the same principles and work perfectly. So David, please just double check this out and let me know if you need anything from me? Let me know your thoughts.

I assume that Ben will close his PR and this one will be ready for the merge only.


beganovich:
Looks good, thanks :+1:

PR #526, opened 2023-03-13
[Bugfix] Cloning Invoice - Line items

Civolilah:
@beganovich @turbo124 David noticed an bug in misplacing line items when cloning an invoice from a table. The reason for this is because we don't have an `_id` property for each line item in the table response. On the edit page in the response we have this property for each line item and we haven't had any bugs there. Looking deep into the code, the line item index is calculated using the `_id` property with this line of code:

`const index = resource.line_items.findIndex((item) => item._id === _id);`

Then I realized that we actually get the same result if we just pass the `lineItemIndex` from the line item table map. I've tested all the cases I know about this and I haven't seen any bad impact just this and also the code is formatted a bit (unnecessary code removed). Ben, if you know any part of the logic that this change will make an bug, please let us know. Otherwise, I think this is a solution for bug. Let me know your thoughts.

### DIFF
--- a/src/pages/invoices/common/components/ProductsTable.tsx
+++ b/src/pages/invoices/common/components/ProductsTable.tsx
@@ -93,17 +93,11 @@ export function ProductsTable(props: Props) {
                             key={columnIndex}
                           >
                             {length - 1 !== columnIndex &&
-                              resolveInputField(
-                                column,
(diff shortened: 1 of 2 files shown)


turbo124:
lgtm @beganovich can we get this merged to main please, this is an important fix.

PR #529, opened 2023-03-14
[Bugfix] Formatting money on cloning

Civolilah:
@beganovich @turbo124 We had a bug with incorrect formatting of money values ​​when cloning entities (invoice, quote, etc.). The actual bug was missing UI rerendering when we tried to clone entities with the same `client_id/vendor_id` twice in a row. On the second attempt, we will not have the proper UI displayed (with formatted values). The `atom` and `resolveClient` parts of the logic we actually use on creation pages (especially atom) have the biggest impact to resolve it like this. This change will resolve the bug in all entities where we had this bug. Let me know your thoughts.

### DIFF
--- a/src/common/helpers/number.ts
+++ b/src/common/helpers/number.ts
@@ -58,7 +58,7 @@ export class Number {
     } else if (swapSymbol) {
       return `${formatted} ${symbol.trim()}`;
     } else if (!options?.showCurrencyCode) {
-      return `${symbol}${formatted}`;
+      return `${symbol} ${formatted}`;
     }
 
     return this.formatValue(formatted, currency);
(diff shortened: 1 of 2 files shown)


turbo124:
Cannot properly QA this as the issue is intermittent in UI. I wasn't able to recreate the issue however with this PR.

Civolilah:
> Cannot properly QA this as the issue is intermittent in UI. I wasn't able to recreate the issue however with this PR.

@turbo124 @beganovich I did a demo for David on Slack so we're on the same line with the bug. Additionally, David noticed that sometimes there is no space between the currency code and the value. So I just added space where we were missing it. Let us know what you think.

beganovich (approved):
Looks good, thanks :+1:

PR #548, opened 2023-03-18
[Cleanup] Adding sort=id|desc for table views

Civolilah:
@beganovich @turbo124 PR includes adding sort query parameter for all queries on the table views where we did not have it earlier (sort=id|desc). Let me know your thoughts.


turbo124:
looks good!

PR #550, opened 2023-03-19
[Cleanup] Upgrade CTA Label - Self Hosted

Civolilah:
@beganovich @turbo124 PR includes changing the label in the upgrade CTA in the top navigation bar to translate the `purchase_white_label` keyword. Here's a screenshot:

Screenshot 2023-03-19 at 16 17 12

We do not have this keyword in the translation files, so if it is okay from your end also, please just add it. Let me know your thoughts.


beganovich (approved):
Looks good, thanks :+1:

PR #551, opened 2023-03-19
[Cleanup] Default checking | Purchase Order Vendor Contacts

Civolilah:
@beganovich @turbo124 PR includes adding condition for default checking vendor contacts on purchase order creation. So, if `send_email` property is `true` on specific contact, that contact will be checked by default. Let me know your thoughts.

### DIFF
--- a/src/pages/purchase-orders/edit/components/VendorSelector.tsx
+++ b/src/pages/purchase-orders/edit/components/VendorSelector.tsx
@@ -45,7 +45,9 @@ export function VendorSelector(props: Props) {
       (i) => i.vendor_contact_id === id
     );
 
-    return Boolean(potential);
+    const contact = vendor?.contacts.find((contact) => contact.id === id);
+
+    return Boolean(potential) || Boolean(contact?.send_email);
   };
 
   return (


beganovich (changes requested):
Even though this "works", it can lead to an invalid state in the app due Checkbox not being synced with the record. You (almost) always want UI to be representation of the record, not opposite.

The recommended approach to fix this is the one we use in other resources, for example, invoices. Check the following code:

https://github.com/invoiceninja/ui/blob/main/src/pages/invoices/create/Create.tsx#L164

Civolilah:
> Even though this "works", it can lead to an invalid state in the app due Checkbox not being synced with the record. You (almost) always want UI to be representation of the record, not opposite.
> 
> The recommended approach to fix this is the one we use in other resources, for example, invoices. Check the following code:
> 
> https://github.com/invoiceninja/ui/blob/main/src/pages/invoices/create/Create.tsx#L164

@beganovich Ah okay, I didn't know this part of logic solved it. The logic has been changed and the bug has been fixed the same as for the invoice creation page. Let me know your thoughts.

beganovich (approved):
Awesome, thanks :+1:

PR #552, opened 2023-03-19
[Cleanup] Taxes displaying refactor

Civolilah:
@beganovich @turbo124 PR includes changing the way taxes are displayed in the totals section. We had grouped taxes without separating the label and value, now we have the label on the left and the value on the right. Here's a screenshot:

Screenshot 2023-03-20 at 00 10 30

Let me know your thoughts.


beganovich (approved):
Looks good, thanks :+1:

PR #553, opened 2023-03-19
[Bugfix] Displaying bulk actions | Product Edit Page

Civolilah:
@beganovich @turbo124 We had a bug with displaying bulk actions when switching to the `Documents/Product Fields` tabs. It's resolved now and looks good from my end. Here's a screenshot:

Screenshot 2023-03-20 at 00 40 06

Let me know your thoughts.


beganovich:
Better to be explicit about the check. See the updated code.

Thanks :+1:

PR #556, opened 2023-03-20
[Feature] Report permissions

Civolilah:
@beganovich @turbo124 PR includes adding the `view_reports` permission to show/hide `Reports` option in the left bar. I'm not sure we have any other place to use this permission yet, other than the show/hide Reports menu option. Let me know your thoughts.


turbo124:
@Civolilah thats correct, this should show / hide the reports section, and also prevent users from accessing the /reports URL without this permission or is_admin/is_owner

Civolilah:
> @Civolilah thats correct, this should show / hide the reports section, and also prevent users from accessing the /reports URL without this permission or is_admin/is_owner

@turbo124 Thanks for explanation. It is actualy finished already on this PR. So, it is ready for @beganovich review.

PR #557, opened 2023-03-21
[Bugfix] Bulk Routes

Civolilah:
@beganovich @turbo124 We had a bug with bulk route for some entities. The reason for this is that we didn't specify `bulkRoute` before, so after adding `?sort=id|desc` the route was broken. Now that's fixed. Let me know your thoughts.


turbo124:
Thanks this looks good, @beganovich when satisified, we'll need to get htis into Main as soon as possible please.

beganovich (approved):
Looks good, thanks :+1:

PR #565, opened 2023-03-24
[Feature] Time Log Description

Civolilah:
Here is the screenshot of updated UI when Description column is added:

Screenshot 2023-03-24 at 19 37 14

@beganovich @turbo124 PR includes adding a `Description` column to the task time log table and managing the `Description` field. This column will be shown if `company.settings.show_task_item_description is true`, otherwise it will be hidden. I researched the `Task Settings` page and see that we don't have a Toggle implemented to handle this property. I didn't implement it in this PR because it wasn't mentioned in the ticket. But please let me know if I need to include adding a Toggle to manage the `show_task_item_description` property. Let me know your thoughts.

### DIFF
--- a/src/common/interfaces/company.interface.ts
+++ b/src/common/interfaces/company.interface.ts
@@ -62,6 +62,7 @@ export interface Settings {
   military_time: boolean;
   language_id: string;
   show_currency_code: boolean;
+  show_task_item_description: boolean;
   show_email_footer: boolean;
   company_gateway_ids: string;
   currency_id: string;
--- a/src/pages/tasks/common/components/TaskTable.tsx
+++ b/src/pages/tasks/common/components/TaskTable.tsx
@@ -24,7 +24,7 @@ import {
   parseTime,
   parseTimeToDate,
 } from '../helpers';
-import { parseTimeLog } from '../helpers/calculate-time';
+import { parseTimeLog, TimeLogsType } from '../helpers/calculate-time';
 
 interface Props {
   task: Task;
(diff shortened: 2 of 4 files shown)


turbo124:
@Civolilah 

Can we please get the description on a new line, we'll also need to add a checkbox for the billable flag

We use company.settings.show_task_item_description as the flag to determine if we show the description dialog

We use company.settings.allow_billable_task_items to determine if we show the billable checkbox.

Civolilah:
> @Civolilah
> 
> Can we please get the description on a new line, we'll also need to add a checkbox for the billable flag
> 
> We use company.settings.show_task_item_description as the flag to determine if we show the description dialog
> 
> We use company.settings.allow_billable_task_items to determine if we show the billable checkbox.

@turbo124 As we discussed in Slack, I moved the `Description` field to a new row so we can put the `allow_billable_task_items` property in the column. Additionally, I also implemented a toggle to handle the `show_task_item_description` property within the task settings. Screenshots:

Screenshot 2023-03-25 at 15 48 51

Screenshot 2023-03-25 at 16 01 51

I assume you want to add some help text for this field as well as the others. That's why I added a keyword for it. We don't have `show_task_item_description` and `show_task_item_description_help` in the translation files. If you agree with me, please just add. Let me know your thoughts.

turbo124:
@Civolilah 

looks good, can we make the description box have less height by default please, it should only be 1-2 lines of height.

Civolilah:
> @Civolilah
> 
> looks good, can we make the description box have less height by default please, it should only be 1-2 lines of height.

@turbo124 Nicee, here is the screenshot when the number of rows for `Description` field is 2:

Screenshot 2023-03-26 at 01 43 27

IMO, it looks good, let me know your thoughts.

beganovich (approved):
Looks good, thanks :+1:

PR #567, opened 2023-03-25
[Bugfix] created_by column | Expenses

Civolilah:
@beganovich @turbo124 While working on the columns task, I saw that we have the `column_by` column on Expenses page active and available for use. This is not the case for other entities, that part of the code is commented out on every other entity with the message that we need to resolve the relationship. I just wanted to remove this column from the dropdown list until we resolve the relationship so that it is available to use this column properly. Screenshot:

Screenshot 2023-03-24 at 20 27 10

Of course, let me know if we already have this relationship resolved. Let me know your thoughts.


beganovich (approved):
Looks good, thanks :+1:

PR #569, opened 2023-03-26
[Bugfix] UX - Edit User Page - Closing Password Confirmation Modal

Civolilah:
@beganovich @turbo124 PR includes improving the UX when closing the password confirmation modal on the `Edit User` page. Previously, when we close the password confirmation modal, we got this weird UI where we don't have any tab because the user didn't enter the correct password and from this page the user can't open the modal again. Screenshot:

Screenshot 2023-03-26 at 03 19 20

Now, instead of just hiding the modal when the user just closes the password confirmation modal, we navigate the user to the `/settings/users` page where he is available again to open the same `Edit User` page and enter the correct password. IMO, this behavior improves the UX for this case. But, let me know your thoughts.

### DIFF
--- a/src/components/PasswordConfirmation.tsx
+++ b/src/components/PasswordConfirmation.tsx
@@ -9,6 +9,7 @@
  */
 import { ChangeEvent, FormEvent, useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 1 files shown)


beganovich (approved):
LGTM :+1:

PR #572, opened 2023-03-27
[Feature] Additional e2e tests - Invoice Entity

Civolilah:
@beganovich @turbo124 PR includes:

- Improving the `can view invoice` test with additional checks
- Creating an invoice deletion test
- Creating test for archiving invoice
- Creating test for cloning invoice
- Creating test for previewing documents tab
- Creating test for uploading documents

Let me know your thoughts.


beganovich:
Tests are failing on my side. Multiple issues:

- Table is not showing permissions@example because it is buried on the second page (implement user search).
- Not using scoped user (invoices@example.com)
- Tests will fail even with correct data because ClientSelector will return an empty dropdown when `view_client` is not present. Make sure to include that permission on create pages as well.

Civolilah:
> 

@beganovich I just updated updated PR with additional changes per latest seeder. All the above issues have been solved. I just did it on this entity and when we merge it, I will apply it to the other entities because we have some shared changes. Let me know your thoughts.

tests/e2e/helpers.ts
```diff
@@ -21,10 +21,17 @@ export async function login(
 }
 
 export function permissions(page: Page) {
-  const clear = async () => {
+  const clear = async (email = "permissions@example.com") => {
```
beganovich, line 24: Never have null variables as the first / or only one. Rather pass the default value.

beganovich (approved):

PR #578, opened 2023-03-29
[Bugfix] Password Confirmation Navigation/Closing

Civolilah:
@beganovich @turbo124 The last time when I was working on improving the behavior of the `/settings/users/:id/edit` page when the user closes the `Password Confirmation` modal, I didn't notice that we were using the same component in the rest of the app. So when the user clicks the `X` on any Password Confirmation modal, the user will be navigated to the `/settings/users` page, even though it's a completely different page.

So I fixed that and just added a condition to navigate the user to the `/settings/users` page only if the user closes the modal from the `/settings/users/:id/edit` page. 

Note: Sorry for this little mistake on the last PR.

Let me know your thoughts.

### DIFF
--- a/src/components/PasswordConfirmation.tsx
+++ b/src/components/PasswordConfirmation.tsx
@@ -9,7 +9,7 @@
  */
 import { ChangeEvent, FormEvent, useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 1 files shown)


beganovich (approved):
LGTM :+1:

PR #581, opened 2023-03-29
[Bugfix] More Actions Dropdown | Payment Edit Page

Civolilah:
@beganovich @turbo124 We had a bug with hidden `More Actions` dropdown if we change the tab from `Edit` to `Documents/Custom Fields`. I fixed it and improved the UX for bulk actions here. For this entity, if we delete it, we will not be able to get it back. So now, if the user deletes a payment from the edit page, there is no bulk action under the dropdown that will be available to call. The same behavior will be implemented on the table, but on a different PR. Let me know your thoughts.

### DIFF
--- a/src/pages/payments/routes.tsx
+++ b/src/pages/payments/routes.tsx
@@ -60,27 +60,11 @@ export const paymentRoutes = (
         />
       }
     >
+      <Route path="edit" element={<Edit />} />
       <Route path="documents" element={<Documents />} />
       <Route path="payment_fields" element={<PaymentFields />} />
       <Route path="apply" element={<Apply />} />
       <Route path="refund" element={<Refund />} />
-    </Route>
-    <Route path=":id/edit" element={<Payment />}>
-      <Route
-        path=""
-        element={
-          <Guard
-            guards={[
-              or(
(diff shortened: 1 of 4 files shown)


src/pages/payments/routes.tsx
```diff
@@ -60,27 +60,11 @@ export const paymentRoutes = (
         />
       }
     >
+      <Route path="edit" element={<Edit />} />
       <Route path="documents" element={<Documents />} />
       <Route path="payment_fields" element={<PaymentFields />} />
       <Route path="apply" element={<Apply />} />
       <Route path="refund" element={<Refund />} />
     </Route>
-    <Route path=":id/edit" element={<Payment />}>
```
Civolilah, line 68: @beganovich Is there any particular reason why we separated this route from the others? 

It looks like we have a free space to improve code clarity here. We had the same `guards` and the same `path` for both of them. With the older code we had unnecessary rendering every time we changed the tab to `Edit`. That's the actual reason for deleting this.

beganovich: Before it was, I think that case is gone.

beganovich (approved):
LGTM :+1:

PR #582, opened 2023-03-29
[Bugfix] Restore Option | Payment Tables

Civolilah:
@beganovich @turbo124 As we know, the only entity that is not available for restoring is `Payment`. So, I made this PR to hide this option on `Payment tables` when a particular payment is deleted. For other entities, the behavior will be the same as earlier. Screenshots:

Screenshot 2023-03-29 at 16 33 29

Screenshot 2023-03-29 at 16 33 47

Let me know your thoughts.


src/components/DataTable.tsx
```diff
@@ -160,6 +160,13 @@ export function DataTable<T extends object>(props: Props<T>) {
     },
   ];
 
+  const isRestoreAvailable = (isDeleted: boolean) => {
+    return (
+      (props.resource === 'payment' && !isDeleted) ||
+      props.resource !== 'payment'
+    );
+  };
+
```
beganovich, line 169: This is not good. DataTable should know nothing about any specific resource. This way we are coupling logic where it shouldn't be.

Just like we have `withResourcefulActions`, I suggest making `withoutResourcefulActions` prop that accept array of possible actions to exclude.

This way we don't couple logic with resource specifically and it's expandable.

beganovich (changes requested):

src/components/DataTable.tsx
Civolilah: Yeah, I agree with you.

I change it now. But, I have to mention one thing.

This is very specific case, maybe we will not have any other case where we will have to exclude one of Delete/Archive/Restore actions. So, I created not too much complex logic.

I'm not sure about the `withoutResourcefulActions` solution, because we need a condition to exclude a specific action. The restore option is not always hidden, only if the payment has been deleted. After the payment is archived, it should work normally.

Let me know your thoughts about solution that I implemented now.

beganovich (approved):
Looks good, thanks :+1:

PR #585, opened 2023-03-30
[Cleanup] Task Invoicing | Billable Time Logs

Civolilah:
@beganovich @turbo124 When we did the invoicing task, we had a situation where we invoiced all time logs, regardless of whether that time log had a `billable` value of true or false. Now I have just added a condition to invoice only time logs those have billing equal true. Let me know your thoughts.


turbo124:
@Civolilah we need to adjust the functionality here.

If the billable tasks option is OFF, we need to bill all tasks.

When the billable tasks option is ON we then selectively bill all tasks. Note, by default all tasks should be marked as billable, and the user can deselect if they do not wish for a particular task to be billable.

Civolilah:
> @Civolilah we need to adjust the functionality here.
> 
> If the billable tasks option is OFF, we need to bill all tasks.
> 
> When the billable tasks option is ON we then selectively bill all tasks. Note, by default all tasks should be marked as billable, and the user can deselect if they do not wish for a particular task to be billable.

@turbo124 Okay, those things you mentioned are adjusted and on the PR. Let me know your thoughts.

turbo124:
@Civolilah I've messaged you a screencast of the remaining issues, please update here when resolved.

Civolilah:
> @Civolilah I've messaged you a screencast of the remaining issues, please update here when resolved.

@turbo124 I just updated PR with changes for those we talked about. When we do not have billable value in the time log array... Let me know your thoughts.

beganovich:
Has this passed final QA? @turbo124

turbo124:
@beganovich yes it has :)

beganovich (approved):

PR #586, opened 2023-03-30
[Feature] Invoice Task Hours And Invoice Task Project |  New Task Settings

Civolilah:
@beganovich @turbo124 This PR includes the addition of two new task settings. We use those settings to customize the `description` line item field. If we turned on `invoice_task_hours`, we will include the hours of each time_log in the description field. If we turned on `invoice_task_project`, we will include the project name at the top of the description. 

Screenshot of updated task settings:

Screenshot 2023-03-30 at 18 00 24

Let me know your thoughts.


src/pages/tasks/common/hooks/useInvoiceTask.ts
```diff
@@ -55,7 +64,30 @@ export function useInvoiceTask() {
...
+      .then((response: GenericSingleResourceResponse<Project>) => {
+        project = response.data.data;
+      })
+      .catch((error: AxiosError) => {
+        toast.error();
+        console.error(error);
+      });
+
+    return project;
+  };
```
beganovich, line 88: Shouldn't we extract this to hook as other queries?

Civolilah: Hmm, tbh I've looked at other cases where we use `queryClient.fetchQuery` and nowhere do we separate that in the hook. We just created a function and in that function we called `fetchQuery` (or we call it in the useEffect). We have that case on the `invoicing project`, there we only have the function `handleInvoiceProject` where we call fetchQuery. IMO, it's good to keep it like this.

But, of course, if you think it's much better to extract it, I'll do it. Let me know your thoughts.

Civolilah: @beganovich Aham, okay. I extracted it into hook. I hope that you agree with this solution. Let me know your thoughts.

beganovich (approved):

PR #594, opened 2023-04-03
[Bugfix] Transaction Table | Query Invalidation

Civolilah:
@beganovich @turbo124 PR includes a fix for the `Transaction` table not updating when `converting` a transaction. On my end it works perfectly now. Let me know your thoughts.


Civolilah:
> 

@beganovich Hmm, the point wasn't to fix the functionality. Functionality (linking, converting) works fine. We just missed a properly update/invalidation query on the table. In this atom we have the correct table endpoint value and with the correct invalidation we will get refreshed table results. Actually, the status of transaction was not updated when we link/convert transaction. Let me know your thoughts.

beganovich (approved):

PR #596, opened 2023-04-03
Live preview for customize and preview

beganovich:
This improves the live preview page & merges it with old settings for the invoice design.


turbo124:
@beganovich is this for review?

beganovich:
Implemented a few missing settings. Should be good for review @turbo124 :+1:

turbo124:
image

discount and tax are missing from the quote product columns options.

also, it would be good to show the label for the custom values across the entities. in the options rather than First Custom...

turbo124:
I see these columns in flutter?

image

PR #600, opened 2023-04-04
Fixes for organising results (clearing)

beganovich:
This implements a fallback to empty records when organizing (fixes for clearing the results).


turbo124:
Thanks, looks good, can we get this into main please.

PR #611, opened 2023-04-08
[Feature] Not Found Page

Civolilah:
@beganovich @turbo124 PR includes creating a `Not-Found` page for all routes that we don't have in our application, but the user somehow enters them. I tried to make the page as simple and clear as possible without any background images and stuff like that. Here is a screenshot:

Screenshot 2023-04-08 at 16 46 10

`Note`: I used plain text with no translation keywords to show how the UI might end up looking once we get the translation keywords. If you agree with this text concept, just give me suggestions for translation keywords.

Let me know your thoughs.

### DIFF
--- a/src/common/routes.tsx
+++ b/src/common/routes.tsx
@@ -30,6 +30,7 @@ import { purchaseOrderRoutes } from '$app/pages/purchase-orders/routes';
 import { reportRoutes } from '$app/pages/reports/routes';
 import { transactionRoutes } from '$app/pages/transactions/routes';
 import { recurringExpenseRoutes } from '$app/pages/recurring-expenses/routes';
+import { NotFound } from '$app/components/NotFound';
 
 export const routes = (
   <Routes>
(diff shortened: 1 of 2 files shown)


turbo124:
@Civolilah 


Can we change the bottom logo to either invoiceninja-white-logo.png or invoiceninja-black-logo-2.png which should be available in public/images.

We'll also want to remove the CTA, also for now lets just  use the translation `api_404` and no other text for this page.

Civolilah:
> @Civolilah
> 
> Can we change the bottom logo to either invoiceninja-white-logo.png or invoiceninja-black-logo-2.png which should be available in public/images.
> 
> We'll also want to remove the CTA, also for now lets just use the translation `api_404` and no other text for this page.

@turbo124 Sure, here is the updated UI for this page with implemented all of your suggestions:

Screenshot 2023-04-09 at 04 59 33

Let me know your thoughts.

Civolilah:
> 

a

turbo124:
Looks good.

beganovich (approved):

PR #646, opened 2023-04-21
[Cleanup] Date | Time Formatting

Civolilah:
@beganovich @turbo124  Initially, I believed that the changes required for this PR would be minor. However, it turned out to be the most challenging task I have faced since joining the company. This PR involves several modifications, including:

- Replacing our conventional date picker with the one from the react-datepicker npm package with adjustments for UI, which we already had installed.
- Adapting the value-changing approach to work with the new date picker package.
- Improved logic where we missed date formatting (such as Transaction pages).
- Developing a time picker that fulfills our requirements for entering and displaying time values in the app's format (12/24 hours format).
- Changing the input fields for time value entry (it was just four places).
- Formatting the time value in the line item description to match the company's format.

I believe that we have achieved exactly what we need, and everything is now perfectly formatted. Please let me know your thoughts on this.


turbo124:
@Civolilah 

Does this datepicker support localization ?

Also, i note there is no easy way using the date picker to select a month or year. 

we may need to look into a different picker?

Civolilah:
> 

@turbo124 

The package absolutely support localization, should I include its implementation in this PR? 

Also, the package is highly customizable, and I have already started working on making it easier to change months and years. The customization feature will be added soon.

IMO, this package is very good for our purpose.

src/common/hooks/useResolveDayJSLocale.ts
```diff
@@ -0,0 +1,303 @@
...
+ * Invoice Ninja (https://invoiceninja.com).
+ *
+ * @link https://github.com/invoiceninja/invoiceninja source repository
+ *
+ * @copyright Copyright (c) 2022. Invoice Ninja LLC (https://invoiceninja.com)
+ *
+ * @license https://www.elastic.co/licensing/elastic-license
+ */
+
+export function useResolveDayJSLocale() {
```
Civolilah, line 11: @beganovich, this is the only reliable way I found to import the correct locale from dayjs/locale after a lot of research. The problem is that if we ignore Vite for dynamic imports, we'll run into issues with webpack not allowing it. The issue is that we need to put variables into the path from which we import something. (That is actually something is not allowed by webpack)

Here's a reference for this:

https://webpack.js.org/api/module-methods/#dynamic-expressions-in-import

I know that hard-coding isn't ideal, and I've talked to David about it, but we must not change this even if we add more languages to our app because it contains all available languages from dayjs. So, our app can't have more languages than what dayjs offers. The `fetch` method won't help here also. I'm open to your suggestions, but I haven't found a better solution for this issue. Let me know what you think.

Civolilah:
@turbo124 Just confirmation, everything is prepared for review here. Localization and all other things regarding that are finished.

turbo124:
@Civolilah 

looks good! Thanks!!

src/common/hooks/useResolveDayJSLocale.ts
beganovich: This looks fine. I am not sure what it has to do with Webpack, we don't use it. Vite uses Rollup to bundle assets, but ok.

Nonetheless, it's an ok approach :+1:

src/components/forms/Datepicker.tsx
```diff
@@ -8,22 +8,277 @@
...
+interface Month {
+  value: number;
+  label: string;
+}
+
+const MONTHS: Month[] = [
+  {
+    value: 0,
+    label: 'january',
+  },
```
beganovich, line 53: Are we sure we want to index months by 0?

Does this affect the output (on change handler)?

Civolilah: @beganovich No worries, I tested it and after your comment I tested it again, it works as expected. Actually, I was also surprised to see that months in Javascript start from 0, but not from 1. So it's just adapted to the JS logic.

PR #659, opened 2023-04-26
Rework combobox component

beganovich:
This rewrites combobox component & decouples the logic from the fetching aspect. Additionally, it improves accessibility by allowing usage via a keyboard.

For now, it is only implemented in ClientSelector as it's possible to incrementally adopt this component.


turbo124:
@beganovich was this ready for review?

PR #662, opened 2023-04-30
[Cleanup] Improve Gateway Refund Logic

Civolilah:
@beganovich @turbo124 The PR includes improving the logic for the `Gateway Refund` option. Now we will first take the `gateway_type_id` from the `Payment` object and using this `gateway.options[gatewayTypeId].refund` logic we will take the right `optionObject` from the `gateway.options` and check the refund value. If the refund value is true, we will display a toggle. Let me know your thoughts.

### DIFF
--- a/src/pages/payments/refund/Refund.tsx
+++ b/src/pages/payments/refund/Refund.tsx
@@ -131,7 +131,7 @@ export default function Refund() {
       const gateway: Gateway = companyGateway.data.data.gateway;
       const gatewayTypeId = payment.data.data.gateway_type_id;
 
-      const showGatewayRefund = gateway.options[gatewayTypeId].refund;
+      const showGatewayRefund = Boolean(gateway.options[gatewayTypeId]?.refund);
 
       setShouldShowGatewayRefund(showGatewayRefund);
     }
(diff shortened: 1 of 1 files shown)


turbo124:
lgtm

src/pages/payments/refund/Refund.tsx
```diff
@@ -127,12 +127,11 @@ export default function Refund() {
...
   useEffect(() => {
-    if (companyGateway) {
+    if (companyGateway && payment) {
       const gateway: Gateway = companyGateway.data.data.gateway;
+      const gatewayTypeId = payment.data.data.gateway_type_id;
 
-      const showGatewayRefund = Object.values(gateway.options).some(
-        (option) => option.refund
-      );
+      const showGatewayRefund = gateway.options[gatewayTypeId].refund;
```
beganovich, line 134: Won't this crash if we are trying to access w/the index missing? Shouldn't TS complain here?

Civolilah: I don't think we need to worry about the app crashing in this case. If the `Payment` object doesn't have a `company_gateway_id`, we won't even get to this part of the code. And if we do have it, we'll always be able to access the correct property under the `gateway.options` object. Still, I added an extra constraint to prevent any possibility of the app crashing.

beganovich (approved):
Looks good, thanks :+1:

PR #664, opened 2023-04-30
[Feature] Refresh | Bank Accounts

Civolilah:
@beganovich @turbo124 The PR includes adding a `Refresh` functionality to the main page of bank accounts. It is a `POST` request to the `'/api/v1/bank_integrations/refresh_accounts'` endpoint with an `empty object payload`. Screenshot:

Screenshot 2023-05-02 at 15 41 08

This is only available for `Enterprise` plan on the `Hosted` platform. Let me know your thoughts.

### DIFF
--- a/src/pages/settings/bank-accounts/index/BankAccounts.tsx
+++ b/src/pages/settings/bank-accounts/index/BankAccounts.tsx
@@ -24,6 +24,7 @@ import { useNavigate } from 'react-router-dom';
 import { AxiosError } from 'axios';
 import { ValidationBag } from '$app/common/interfaces/validation-bag';
 import { toast } from '$app/common/helpers/toast/toast';
+import { Icon } from '$app/components/icons/Icon';
 import { proPlan } from '$app/common/guards/guards/pro-plan';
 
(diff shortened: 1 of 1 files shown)


beganovich:
I am unsure how we ended up having so many `primary` buttons there, but it looks awful. Are they all meant to be primary actions? @turbo124 Should we move buttons to the secondary style and leave only the primary action blue (New bank account)?

turbo124:
@beganovich I think @Civolilah was just following UX from flutter app, but if you think secondary works better, happy to go in that direction.

Civolilah:
> @beganovich I think @Civolilah was just following UX from flutter app, but if you think secondary works better, happy to go in that direction.

@beganovich @turbo124 I absolutely agree with your advice. I had forgotten about this aspect a little. There should be only one primary action, which is to create a new bank account. I updated UI and here is the screenshot:

Screenshot 2023-05-02 at 15 41 08

Let me know your thoughts.

beganovich (approved):
Looks good, thanks :+1:

PR #667, opened 2023-04-30
[Feature] Expense Payment Type Field

Civolilah:
@beganovich @turbo124 PR includes implementation of `Expense Payment Type` field under `Defaults`. Screenshot:

Screenshot 2023-04-30 at 19 14 00

We miss `expense_payment_type` translation keyword from the translation files, so if you agree please just add it. Let me know your thoughts.

### DIFF
--- a/src/pages/settings/company/components/Defaults.tsx
+++ b/src/pages/settings/company/components/Defaults.tsx
@@ -83,6 +83,23 @@ export function Defaults() {
             </Element>
           )}
 
+          <Element leftSide={t('expense_payment_type')}>
+            <SelectField
+              value={companyChanges?.settings?.default_expense_payment_type_id}
+              onChange={handleChange}
+              id="settings.default_expense_payment_type_id"
+            >
+              <option value="0"></option>
+              {statics?.payment_types.map(
+                (type: { id: string; name: string }) => (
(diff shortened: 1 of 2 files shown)


src/pages/settings/company/components/Defaults.tsx
```diff
@@ -83,6 +83,23 @@ export function Defaults() {
             </Element>
           )}
 
+          <Element leftSide={t('expense_payment_type')}>
+            <SelectField
+              value={companyChanges?.settings?.default_expense_payment_type_id}
+              onChange={handleChange}
+              id="settings.default_expense_payment_type_id"
+            >
+              <option value="0"></option>
```
beganovich, line 92: SelectField comes with a `withBlank` prop that injects the empty field. Does that not fit this case?

Civolilah: We couldn't use it because we needed a blank option value of `'0'`, but the blank option only had an empty string value. So, I made improvements to the `SelectField` component by adding a `blankOptionValue` and improved the implementation of SelectFields on the `Defaults` component. I hope you find my solution agreeable.

beganovich (approved):
Looks good, thanks :+1:

PR #669, opened 2023-05-08
[Cleanup] Dashboard Refactoring

Civolilah:
@beganovich @turbo124 PR includes Dashboard refactoring using mockup example. After discussing on the Slack we made some things differently and we excluded some functionalities for now, so the current UI looks like this:

Screenshot 2023-05-15 at 15 59 18

Missing keywords from translation files are: `welcome`, `recent_transactions`, `total_invoices_outstanding` and `recent_activity`.

Let me know your thoughts.


turbo124:
Once we resolve the double queries for totals / chart_summary this is good to merge.

src/components/DataTable.tsx
```diff
@@ -107,6 +123,10 @@ export function DataTable<T extends object>(props: Props<T>) {
 
   const mainCheckbox = useRef<HTMLInputElement>(null);
 
+  const handleVerticalOverflowChange = (hasOverflow: boolean) => {
+    setHasVerticalOverflow(hasOverflow);
+  };
+
```
beganovich, line 129: This doesn't need to be a method. I checked and we call it only once. Even if it had multiple references, it'd need better co-location or useCallback.

Nonetheless, we can just directly call `setHasVerticalOverflow` unless I'm missing something?

src/components/tables/Table.tsx
```diff
@@ -9,25 +9,124 @@
...
+        } else {
+          setNumberOfCells(
+            element.querySelectorAll('tbody > tr > td').length ?? 0
+          );
+        }
+      }
+
+      setTableHeight(element.clientHeight);
+    }
+  };
```
beganovich, line 61: Oof, 3 levels if. What are we trying to do here? There has to be easier/cleaner way.

src/pages/dashboard/components/Chart.tsx
```diff
@@ -58,13 +55,31 @@ export function Chart(props: Props) {
...
+          new Date(date) >= new Date(data[index].date) &&
+          new Date(date) <= new Date(data[index + 1].date)
+        ) {
+          recordIndex = index;
+        }
+      }
+    });
+
+    return recordIndex;
+  };
```
beganovich, line 79: This looks really complex. Can we assign internal IDs and then look them up like that? Perhaps .find on records?

src/pages/dashboard/components/ExpiredQuotes.tsx
```diff
@@ -52,36 +48,54 @@ export function ExpiredQuotes() {
...
-        withoutPadding
-      />
+      <div className="pl-6 pr-4">
+        <DataTable
+          resource="quote"
+          columns={columns}
+          className="pr-4"
+          endpoint="/api/v1/quotes?include=client&client_status=expired&without_deleted_clients=true&per_page=50&page=1&sort=id|desc"
+          withoutActions
+          withoutPagination
```
beganovich, line 82: Can we bundle this in some object perhaps? Perhaps:

interface StyleOptions {
   withoutFoo: boolean
}

Then pass it <DataTable styleOptions={{ withoutFoo: true }} />

src/pages/dashboard/components/Totals.tsx
```diff
@@ -133,103 +137,137 @@ export function Totals() {
...
+                  {currency.label}
+                </option>
+              ))}
+            </SelectField>
+          )}
+
+          <div className="flex space-x-2">
+            <Button
+              key="day-btn"
+              type={`${chartScale === 'day' ? 'primary' : 'secondary'}`}
```
beganovich, line 159:
```tsx
type={chartScale === "day" ? "primary : "secondary"}
```

src/pages/dashboard/components/Totals.tsx
```diff
@@ -133,103 +137,137 @@ export function Totals() {
...
+              onClick={() => setChartScale('day')}
+            >
+              {t('day')}
+            </Button>
+
+            <Button
+              key="week-btn"
+              type={`${chartScale === 'week' ? 'primary' : 'secondary'}`}
```
beganovich, line 167: Same as previous.

src/components/CompanySwitcher.tsx
```diff
@@ -117,14 +117,23 @@ export function CompanySwitcher() {
...
-            <ChevronDown size={18} className="text-gray-900" />
-          </Menu.Button>
-        </div>
+            <div className="flex flex-col items-between">
+              <span className="text-white text-sm text-start w-28 truncate">
+                {companyName}
+              </span>
+              {(user?.first_name || user?.last_name) && (
+                <span className="text-white text-xs text-start w-28 truncate">
+                  {`${user.first_name} ${user.last_name}`}
```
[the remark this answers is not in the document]
Civolilah: Changed!

src/components/DataTable.tsx
Civolilah: Yeah, you are right, we call it just once so we can call it while passing property to Table component.

```
<Table
        onVerticalOverflowChange={(hasOverflow) =>
          setHasVerticalOverflow(hasOverflow)
        }
      >
```

src/pages/dashboard/components/Totals.tsx
Civolilah: changed

src/pages/dashboard/components/Totals.tsx
Civolilah: changed

src/components/tables/Table.tsx
Civolilah: I changed the logic so I avoided this part with optimized solution.

src/pages/dashboard/components/ExpiredQuotes.tsx
Civolilah: Yes, it's a good idea. I've grouped all the new styling properties under the `StyleOptions` object. However, there are some properties, like `withoutPagination`, that are used throughout the entire application. I believe it wouldn't be a good idea to modify them in this pull request because it would require changing the `DataTable` component in other parts of the app as well. Here's how the `StyleOptions` object looks now:

```
interface StyleOptions {
  withoutBottomBorder?: boolean;
  withoutTopBorder?: boolean;
  withoutLeftBorder?: boolean;
  withoutRightBorder?: boolean;
  headerBackgroundColor?: string;
  thChildrenClassName?: string;
  tBodyStyle?: CSSProperties;
  thClassName?: string;
  tdClassName?: string;
  addRowSeparator?: boolean;
}
```

Civolilah:
> It looks like the mobile company name isn't visible:
> 
> image

@beganovich Actually, no, I just missed changing background color. Now, it is fixed and looks as expected. Screenshot:

Screenshot 2023-05-18 at 01 44 08

src/pages/dashboard/components/Chart.tsx
Civolilah: @beganovich I'm not sure if it's possible to handle IDs in this situation. The problem arises when dealing with week/month intervals. In these cases, we need to compare dates. When a user selects a month, there are only a few dates available, making it difficult to accurately match them and display the correct chart. Comparing the dates is the only practical solution. I've simplified the approach, and I hope you agree with it.

beganovich (approved):
Looks good, thanks :+1:

PR #673, opened 2023-05-15
[Bugfix] Missing Client Name | Client View Page

Civolilah:
@beganovich @turbo124 We had a bug of missing client name in some tables on the Client view page. It is fixed and here is the screenshot:

Screenshot 2023-05-15 at 20 36 27

Let me know your thoughts.


beganovich (approved):
Looks good, thanks!

PR #683, opened 2023-05-23
[Bugfix] Designs list | Combobox

Civolilah:
@beganovich @turbo124 This PR is created from Ben's PR for `Combobox` improvement. This PR resolves bug for the list of Designs in the footer of all entities where the `Design selector` exists. For now, using David's advice we will go with disabling `with` query parameter, later we will go deeply through this behaviour and include it with additional logic improvement. Let me know your thoughts.


beganovich (approved):
Looks good, thanks :+1:

PR #685, opened 2023-05-23
[Bugfix] Improved UX for Payment error messages

Civolilah:
@beganovich @turbo124 Currently on the PR we have improving UX for Payment error messages. David, I can not see here any issues regarding not-displaying 422 error messages. Here is the screenshot:

Screenshot 2023-05-24 at 01 07 39

If we still have some issues, please just give me quick advice where it is.

Let me know your thoughts.


beganovich (approved):
Looks good, thanks :+1:

PR #699, opened 2023-05-30
Fixes for displaying enabled modules

turbo124:
Add in logic for purchase orders and transactions


beganovich (approved):
Looks good, thanks :+1:

PR #703, opened 2023-06-04
Add default headers

turbo124:
Sets default headers so we can differentiate in the API which AP type has sent the request


beganovich (approved):
Looks great, thanks :+1:

PR #704, opened 2023-06-05
Fixes for query invalidation

turbo124:


turbo124:
This PR replaces the base filter is_deleted=false and replaces it with status=active which is the preferred filter

also included invalidation of products on bulk actions.

beganovich (approved):
Looks great, thanks :+1:

PR #714, opened 2023-06-06
[Feature] Phone number confirmation banner

Civolilah:
@beganovich @turbo124 PR includes changing the old way of notifying users about an unverified phone number via the modal. It is now implemented through a permanent banner at the top of the page. Here's a screenshot:

Screenshot 2023-06-06 at 18 42 28

This will be only available for HOSTED users.

Let me know your thoughts.


turbo124:
This looks good, however it just redirects to 2FA page.

2FA + Phone verification can be different flows. So we'll need to add in a dedicated Phone verification modal for this.

Also to note, this is for Hosted platform only.

src/components/PhoneVerificationModal.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+  const handleSendSMSCode = (event: FormEvent<HTMLButtonElement>) => {
+    event.preventDefault();
+
+    setErrors(undefined);
+
+    toast.processing();
+
+    request('POST', endpoint('/api/v1/verify'), {
+      phone: `+38761602460`,
```
beganovich, line 62: I assume this is a mistake unless you want to get all the verification messages.

src/components/PhoneVerificationModal.tsx
```diff
@@ -0,0 +1,166 @@
...
+  const [errors, setErrors] = useState<ValidationBag>();
+
+  const handleClose = () => {
+    props.setVisible(false);
+    setPhoneNumber('');
+    setCountryCode('');
+    setErrors(undefined);
+  };
+
+  const handleSendSMSCode = (event: FormEvent<HTMLButtonElement>) => {
```
beganovich, line 54: Two things are wrong here.

1. If you want to prevent the default behavior of the button, you should just use type=button, or as it's in our app `behavior`.
2. You don't wanna do it this way. Please wrap it in the form and then prevent default on the form.

beganovich (changes requested):
Looks good, few changes needed :+1:

turbo124:
country codes used were not correct.

PR #715, opened 2023-06-06
[Feature] Default Password Timeout | Password Confirmation Modal

Civolilah:
@beganovich @turbo124 PR includes implementing functionality for `default_password_timeout` property. So, if timeout is not `expired` comparing with last time entered password we will not show password confirmation modal for endpoints where it is mandatory if timeout has been expired. Let me know your thoughts.


src/components/PasswordConfirmation.tsx
```diff
@@ -23,36 +27,55 @@ export function PasswordConfirmation(props: Props) {
...
           ? navigate('/settings/users')
           : props.onClose(false)
       }
-      visible={isModalOpen}
+      visible={isModalOpen && isPasswordTimeoutExpired}
       title={t('confirmation')}
       text={t('please_enter_your_password')}
     >
-      <form onSubmit={handleConfirm}>
+      <form onSubmit={(event) => handleConfirm(event, true)}>
```
beganovich, line 78: Why do we need  second param if it's always true?

beganovich (commented):
This looks to me like ton of repetition.
Can we abstract time & pass confirmed automatically if it's filled already?

src/components/PasswordConfirmation.tsx
Civolilah: @beganovich Actually, this parameter is not always true, here is the code:

```
useEffect(() => {
    if (company && isModalOpen && !isPasswordTimeoutExpired) {
      handleConfirm();
    }
  }, [isModalOpen]); 
```

When this function is called on a form, submitting this variable will provide us with the information we need to update the atom with the time of the last password entered. Otherwise, we don't need to store it because that's the time the password hasn't expired.

Civolilah:
> This looks to me like ton of repetition. Can we abstract time & pass confirmed automatically if it's filled already?

@beganovich This is how I think: The entire logic regarding password expiration is implemented in the `PasswordConfirmation` component, with the `lastPasswordEntryTime` variable abstracted into an atom. The purpose of setting `setLastPasswordEntryTime(0)` for every endpoint that requires `x-api-password` in the header is to ensure that if we encounter a 412 error on any of these endpoints, we force the user to enter their password again, even if the expiration time has not yet passed.

I'm not certain if we can completely avoid repeating some actions in the catch block for endpoints that require `x-api-password`. We must do something, but my approach is to reuse this atom for the implementation of this specific logic.

beganovich:
I guess it should be ok for now, we can iterate.

PR #718, opened 2023-06-07
White label link

turbo124:
There can be scenario's where if .env is not available, whiteLabelLink will return null. hardcoding this here


beganovich (approved):
Looks good, thanks :+1:

PR #719, opened 2023-06-08
[Cleanup] Notifications | User Details

Civolilah:
@beganovich @turbo124 PR includes aligning User Notifications with Flutter app, that is switching from toggle to select menu with additional option. Here is the screenshot:

Screenshot 2023-06-08 at 20 50 16

Let me know your thoughts.


beganovich (approved):
Looks good, thanks :+1:

PR #726, opened 2023-06-09
Fixes for redundant queries

turbo124:
invalidate statics only on language change


beganovich (approved):
Looks good, thanks :+1:

PR #727, opened 2023-06-09
Add project name as optional column on tasks table

turbo124:


beganovich (approved):
Looks good, thanks :+1:

PR #729, opened 2023-06-09
[Cleanup] Email Confirmation Banner | Implementation

Civolilah:
@beganovich @turbo124 PR includes implementing banner at the top of the page instead of modal we already had. Here is the screenshot:

Screenshot 2023-06-09 at 19 02 54

The CTA `Resend Email` will resend email via API through `/api/v1/user/:id/reconfirm` endpoint. This banner will be displayed only for `HOSTED` users.

Let me know your thoughts.


beganovich (approved):
Looks good, thanks :+1:

PR #730, opened 2023-06-09
[Cleanup] Company Activation Banner

Civolilah:
@beganovich @turbo124 PR includes implementation of company activation banner instead of modal we already had. Here is the screenshot:

Screenshot 2023-06-09 at 20 06 26

This banner will be only available for `HOSTED` users.

Let me know your thoughts.


beganovich (approved):
Looks good, thanks :+1:

PR #731, opened 2023-06-10
[Cleanup] React Settings | React Table Columns

Civolilah:
@beganovich @turbo124 PR includes resolving bug for `show_pdf_preview` and additionally `react_table_columns` are moved to the `react_settings` object.

@turbo124 Everything is prepared on this PR to work as expected, but before testing, please ensure that real object is returned from the API instead of JSON.

Let me know your thoughts.


beganovich:
It looks like my PR is a duplicate of this. #738. 

Can you push the fallback the same way I did? We don't wanna people lose their old settings.

Civolilah:
> It looks like my PR is a duplicate of this. #738.
> 
> Can you push the fallback the same way I did? We don't wanna people lose their old settings.

@beganovich Really no probs, I will do that to the end of the day my time!

Civolilah:
@beganovich Throughout the entire solution for moving `react_table_columns` from the `settings` property to `react_settings`, I used the `useReactSettings` hook. It allowed me to globalize the fallback and implement it in just one place. In my opinion, this approach looks cleaner and more globalized. It looks like this:

```
const settings: ReactSettings = {
    show_pdf_preview: true,
    react_notification_link: true,
    ...reactSettings,
    // This is legacy fallback for old settings location. If you see this in 2 years, feel free to remove it.
    react_table_columns: {
      ...previousReactTableColumns,
      ...reactSettings.react_table_columns,
    },
  };
```

I hope that you agree with my suggestion for the solution and the reason why I implemented it like this, if you do not I can change it as you implemented.

Let me know your thoughts.

src/common/hooks/useInjectUserChanges.ts
```diff
@@ -0,0 +1,23 @@
...
+import { useCurrentUser } from './useCurrentUser';
+
+export function useInjectUserChanges() {
+  const user = useCurrentUser();
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    dispatch(injectInChanges());
+  }, [user]);
+}
```
beganovich, line 23: I am not sure I see the point of this. Can you elaborate on why we need this pattern of injecting?

beganovich (approved):
Looks good, need an explanation on why we push injected user pattern for this.

beganovich (approved):

PR #738, opened 2023-06-12
Change table columns path

beganovich:
This moves column settings into the `company_user.react_settings` object.


beganovich:
Closed in favor of #731

PR #754, opened 2023-06-15
Task invoice description filter

turbo124:


src/pages/tasks/common/hooks/useInvoiceTask.ts
```diff
@@ -117,12 +115,34 @@ export function useInvoiceTask() {
               hoursDescription = `• ${hours} ${t('hours')}`;
             }
 
+            const start_date_log = `${dayjs.unix(start).format(`${dateFormat}`)}`;
+            const start_time_log = `${dayjs.unix(start).format(`${timeFormat}`)}`;
+            const end_time_log = `${dayjs.unix(stop).format(`${timeFormat}`)}`;
```
beganovich, line 120: Any particular reason why is this part of single quote?

beganovich (approved):

src/pages/tasks/common/hooks/useInvoiceTask.ts
turbo124: sorry I don't follow, is the question why assign to a prop?

beganovich: No, I was just wondering why we're wrapping the JS in template literals, when the output is already a string.

```ts
const start_date_log = `${dayjs.unix(start).format(`${dateFormat}`)}`;

// vs

const start_date_log = dayjs.unix(start).format(dateFormat)
```

PR #768, opened 2023-06-15
Improve client selector UX

beganovich:
This will hide the client combobox when viewing the invoices if the user has no permission to view it.
The fallback is just the name of the client.

image


turbo124:
Thanks, this is great. Is it also possible to have this as the default functionality when editing invoices. When the client is assigned to an invoice, it can never be changed, so it may be more efficient to just display the name in those cases also which will prevent an additional API hit.

Should this functionality be linked to a new PR?

beganovich:
Sure, no problem. I think this PR is fine. I'll update it in the morning.

PR #786, opened 2023-06-17
Invoices history

beganovich:
This implements a history preview for the invoices.


beganovich:
Ping @turbo124 for the first pass QA.

turbo124:
image

Not sure this works well. We can remove the client link. it may also help to have alternating backgrounds with top/bottom borders?

Finally, is there a clean way to automatically render the invoice when an edit is made? it looks like the user now needs to click on another menu link?

PR #789, opened 2023-06-19
Fixes for filtering line items

beganovich:
Previously we only allowed `InvoiceItemType.Product` (1) to be part of the products table. This fixes it to support other types except for tasks.

The current filter:

```
InvoiceItemType.Product,
InvoiceItemType.UnpaidFee,
InvoiceItemType.PaidFee,
InvoiceItemType.LateFee,
```

### DIFF
--- a/src/common/interfaces/invoice-item.ts
+++ b/src/common/interfaces/invoice-item.ts
@@ -11,6 +11,9 @@
 export enum InvoiceItemType {
   Product = '1',
   Task = '2',
+  UnpaidFee = '3',
+  PaidFee = '4',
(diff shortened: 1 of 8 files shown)


turbo124:
just need to support purchase orders too... otherwise looks good.

beganovich:
Looks like PO has no filtering, anyways. Is it missing? https://github.com/invoiceninja/ui/blob/main/src/pages/purchase-orders/edit/Edit.tsx#L146

turbo124:
Looks good, thanks!

PR #798, opened 2023-06-19
[Bugfix] Custom designs

Civolilah:
@beganovich @turbo124 PR includes a couple of fixes:

- Fixes for query invalidation on Invoice Design pages
- Improving query invalidation on Data Table
- Adding active status query parameter to invoice design endpoints

Let me know your thoughts.

### DIFF
--- a/src/components/DataTable.tsx
+++ b/src/components/DataTable.tsx
@@ -112,7 +112,7 @@ export function DataTable<T extends object>(props: Props<T>) {
   );
 
   const setInvalidationQueryAtom = useSetAtom(invalidationQueryAtom);
-  setInvalidationQueryAtom(props.endpoint);
+  setInvalidationQueryAtom(apiEndpoint.pathname);
 
   const queryClient = useQueryClient();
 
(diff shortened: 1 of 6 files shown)


src/components/DataTable.tsx
```diff
@@ -160,7 +160,15 @@ export function DataTable<T extends object>(props: Props<T>) {
   }, [perPage, currentPage, filter, sort, status, customFilter]);
 
   const { data, isLoading, isError } = useQuery(
-    [props.endpoint, perPage, currentPage, filter, sort, status, customFilter],
+    [
+      apiEndpoint.pathname,
```
Civolilah, line 164: This has changed because `props.endpoint` includes the pathname and query parameters (/api/v1/designs?status=active...), and this becomes the first query key while invalidation. This logic forces us to include the whole URL when invalidating queries and unnecessarily complicates the logic. With this change, the first item in the query keys array will be just (/api/v1/designs), allowing us to handle invalidation more easily.

src/components/DataTable.tsx
```diff
@@ -210,6 +226,8 @@ export function DataTable<T extends object>(props: Props<T>) {
       })
       .finally(() => {
         queryClient.invalidateQueries([props.endpoint]);
```
Civolilah, line 228: I leave this line of code because of some potential queries that include query parameters in the first query key while invalidating (`/api/v1/designs?status=active...` instead of just `/api/v1/designs`).

Civolilah:
@beganovich @turbo124 I additionally tested whether this change affected anything else in the app. It seems that this method of query invalidation works perfectly throughout the entire app. However, please let me know your thoughts.

turbo124:
This looks good. Thanks!

beganovich (approved):
Looks good, thanks :+1:

PR #799, opened 2023-06-19
[Bugfix] Client Create Modal | Client Combobox

Civolilah:
@beganovich @turbo124 While testing query invalidation for comboboxes, I discovered a few bugs in the Client creation modal that I have prioritized to resolve in this PR:

- The combobox for clients had sorting by `display_name|asc`, which caused an issue when there were more than 20 clients. If a newly created client appeared after the first 20 clients based on this sorting, it wouldn't be selected in the combobox after creation. To resolve this, I have decided to use sorting by `created_at|desc`, which seems to be a logical solution. I hope you agree with this change, as it will fix the bug.

- When opening the modal twice in a row, we encountered 2-3 strange bugs related to setting the Client object. As a result, we ended up with a situation where a Client was created without any entered details. It seems that the API was creating a blank Client object, leading to additional issues. Additionally we had bugs in the console for uncontrolled toggles which is also resolved here. I just slightly changed the logic in the modal to resolve those bugs.

- The Cancel button had `type='minimal'`, which made it look unusual. I have simply changed it to `secondary`, which is consistent with how Cancel buttons look throughout our app. I hope you agree with this UI change.

Screenshot 2023-06-19 at 18 45 30

Let me know your thoughts.

### DIFF
--- a/src/components/clients/ClientSelector.tsx
+++ b/src/components/clients/ClientSelector.tsx
@@ -75,7 +75,7 @@ export function ClientSelector(props: ClientSelectorProps) {
         entryOptions={{ id: 'id', label: 'display_name', value: 'id' }}
         onChange={(value) => value.resource && props.onChange(value.resource)}
         staleTime={props.staleTime || 500}
-        sortBy="display_name|asc"
+        sortBy="created_at|desc"
         exclude={props.exclude}
         action={{
           label: t('new_client'),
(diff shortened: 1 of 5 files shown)


turbo124:
This looks good, thanks!

beganovich (approved):
Looks good, thanks :+1:

PR #800, opened 2023-06-19
Add payment type cash

turbo124:
Add cash payment type


beganovich (approved):
Looks good, thanks :+1:

PR #810, opened 2023-06-21
[Bugfix] Query Invalidation Across the App

Civolilah:
@turbo124 @beganovich I went through the entire app to ensure how we `defined` queries and how we `invalidate` them regarding the bug that Shalom reported for Client view page and David noticed in the Transactions. Both of them have been resolved, and additionally, I fixed the queries' definition where we were not doing it correctly. I have extensively tested this, and it appears to me that we do not have any instances of invalidation query bugs.

Let me know your thoughts.


beganovich (approved):
Looks good, thanks :+1:

PR #812, opened 2023-06-21
[Bugfix] Stripe Displaying | Hosted Platform

Civolilah:
@beganovich @turbo124 The PR includes fixing a bug for the non-displaying `Stripe` card gateway on the hosted platform. Now, instead of using the `provider` property, we use the `key` of the gateway to display cards. Therefore, the keys have been adjusted from the local data, and an additional key for `stripe_connect` that David provided me on Slack has been included to resolve this bug.

Let me know your thoughts.

### DIFF
--- a/src/components/DataTable.tsx
+++ b/src/components/DataTable.tsx
@@ -161,7 +161,8 @@ export function DataTable<T extends object>(props: Props<T>) {
 
   const { data, isLoading, isError } = useQuery(
     [
-      apiEndpoint.pathname,
+      props.endpoint,
+      apiEndpoint.href,
       perPage,
       currentPage,
       filter,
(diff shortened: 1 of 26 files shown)


src/components/DataTable.tsx
```diff
@@ -161,8 +161,8 @@ export function DataTable<T extends object>(props: Props<T>) {
 
   const { data, isLoading, isError } = useQuery(
     [
+      apiEndpoint.pathname,
       props.endpoint,
```
Civolilah, line 165: @beganovich This change is just in purpose of testing this PR, because the same change is on the previous PR. So, when we merge that PR I will resolve conflicts here. But, this is just for David's testing.

turbo124:
Thanks, when selecting a gateway, can you please land on the Credentials tab so the user can commence the configuration. Otherwise looks good.

Civolilah:
> Thanks, when selecting a gateway, can you please land on the Credentials tab so the user can commence the configuration. Otherwise looks good.

@turbo124 Added logic for switching tabs `Credentials` when changing gateway type. Let me know your thoughts.

PR #813, opened 2023-06-22
Fixes for payment notification presets

turbo124:
Add Zelle payment type
Fixes for default settings for manual payment notifications
Fixes for adding default task rates when creating tasks.

### DIFF
--- a/src/pages/tasks/common/components/TaskDetails.tsx
+++ b/src/pages/tasks/common/components/TaskDetails.tsx
@@ -47,7 +47,7 @@ export function TaskDetails(props: Props) {
       handleChange('client_id', client.id);
 
       if(!task?.id)
-        handleChange('rate', client?.settings?.default_task_rate ?? company?.settings?.default_task_rate );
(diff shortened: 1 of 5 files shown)


src/pages/tasks/common/components/TaskDetails.tsx
```diff
@@ -41,26 +42,43 @@ export function TaskDetails(props: Props) {
...
+
+  }
+
+  const handleChangeProject = (project_id: string) => {
+
+    handleChange('project_id', project_id);
+    handleChange('client_id', '');
+
+  }
+
```
beganovich, line 60: I don't think we really need such methods, to be honest. Can we just cast multiple events out?

```tsx
onChange={(client) => {
    handleChange('client_id', client.id);
    handleChange('rate', ...)
}}
```

beganovich (changes requested):

src/pages/tasks/common/components/TaskDetails.tsx
turbo124: as long as we can handle the conditional !task.id ... yep!

PR #815, opened 2023-06-22
[Bugfix] Activate Company Banner & Settings Break App On New Sign Up

Civolilah:
@beganovich @turbo124 This PR includes resolving bugs for the Activation Company banner, the number of banners shown at once, and the bug that occurs when navigating to settings pages during sign up, which breaks the app.

Let me know your thoughts.


turbo124:
is any of this necessary if we include `company` after registration? that way company will be present and we don't have to perform these additional checks?

please see slack discussion.

Civolilah:
> is any of this necessary if we include `company` after registration? that way company will be present and we don't have to perform these additional checks?
> 
> please see slack discussion.

@turbo124 I saw the discussion and here is the thing. 

Your suggestion to include `company` in the response for `signup` makes sense and will avoid certain checks. It will prevent the two checks for the `undefined` state of the `company` object.

But here are a few things:

- Our current logic for retrieving the `company` object operates based on the principle of extracting it from the `company_users` array stored in Redux after `Logging` in. By utilizing the current company index, we are able to obtain the correct `company_user` and its corresponding `company` object. -> https://github.com/invoiceninja/ui/blob/main/src/common/hooks/useCurrentCompany.ts.

- So, if we want to maintain the current logic, we need to obtain the same response as during `Login` (a list of `company_users`, although on the `signup` it will be only one). Afterward, we can utilize the same logic as during `Login` to update Redux, which will provide us with the correct details about the current company at a later stage. By doing this, we can avoid the need for two checks: whether the company exists for displaying the banner and displaying Settings in the left menu.

- The other minor checks in this PR are only aimed at preventing the presence of more than one banner at a time.

I hope that I explained well, let me know your advices.

turbo124:
@Civolilah 

thanks for the explanation, if using the company as an include is not a win, then this PR is the solution :+1:

Civolilah:
> 

@turbo124 No probs!!

beganovich (approved):
Looks good, thanks :+1:

PR #819, opened 2023-06-23
Improve contact editing

turbo124:
Add Icon only clipboard
Fixes for adding contacts to existing client
Only display contacts that are receiving emails in the mailer preview


beganovich (approved):
Looks good, thanks :+1:

PR #821, opened 2023-06-23
Fixes for invoice invalidation after payment applied

turbo124:
ensures paid invoices are invalidated


beganovich (approved):
Looks good, thanks :+1:

PR #822, opened 2023-06-23
Fixes for previews

turbo124:
Ensure we pass the entity ID to the live preview engine


beganovich (approved):
Looks good, thanks :+1:

PR #832, opened 2023-06-24
Refactor for activities

turbo124:
Offloads the majority of work for rendering activities onto the API, whilst maintaining the ability to have client side links.

I have used versioning on the API and client to ensure backward compatiblity


beganovich (approved):
Looks good, thanks :+1:

PR #838, opened 2023-06-26
Fixes for duration and passing rates into tasks

turbo124:
Fixes for displaying task duration in list view when time > 1 day
Fixes for injecting project rates into tasks.


beganovich (approved):
Looks good, thanks :+1:

PR #839, opened 2023-06-26
Fixes for showing active tax rates only.

turbo124:


src/pages/settings/tax-settings/components/Selector.tsx
```diff
@@ -25,8 +25,8 @@ export function Selector() {
   const companyChanges = useCompanyChanges();
   const dispatch = useDispatch();
 
-  const { data } = useQuery('/api/v1/tax_rates', () =>
-    request('GET', endpoint('/api/v1/tax_rates'))
+  const { data } = useQuery('/api/v1/tax_rates?status=active', () =>
+    request('GET', endpoint('/api/v1/tax_rates?status=active'))
```
beganovich, line 29: Is this what we want? This means the cache item will be with status=active? Perhaps we wanna use the key as endpoint only without param, but actual request with status active?

beganovich (changes requested):

src/pages/settings/tax-settings/components/Selector.tsx
turbo124: That makes sense, can you adjust please

beganovich:
Looks good, thanks :+1:

PR #846, opened 2023-06-26
Implement ComboboxAsync on payments apply page

beganovich:
This replaces the old DebouncedCombobox with ComboboxAsync... and most importantly:

image


turbo124:
image

Looks good.

We just need to pop the invoice off the list when it has been selected so that it does not get selected again.

Also, if we can have this combo box on the create page that would be great, we can do that in separate PR if required.

beganovich:
Ready for another QA pass @turbo124

turbo124:
=> main please :)

PR #847, opened 2023-06-26
[Bugfix] Improving UX While Updating Company On POST/PUT For Some Entities

Civolilah:
@beganovich @turbo124 This PR improves the UX on pages where we saving company, even though it is a page for a completely different entity. For example, when creating or editing an invoice, the user can set custom fields, which may trigger a request to save the company. However, the saving of the company was not awaited, resulting in the user receiving a message that the invoice is created, followed by a delayed toaster notification stating that the company settings are updated. Additionally, if NO changes were made to the company, the request to update the company would NOT even be called.

This PR aims to enhance the UX by ensuring that when the user receives a message indicating the creation or update of an entity, they can be certain that all required actions have been completed. By making these improvements, users can seamlessly continue using the app without any confusion. I hope you agree with this proposed solution.

Let me know your thoughts.


Civolilah:
> 

@beganovich Let me explain this:

David recorded a video where he started explaining totally different bug to me. He jumped to the Invoice edit page and updated some custom fields. He received the message "updated_invoice" but when he left the page and returned, the custom fields were not updated. He literally asked himself, "Why isn't anything updating?" However, the actual reason for this is that the company request did not finish. I believe this would confuse any users who experience the same issue. If they receive the "updated_invoice" message, they would expect the custom fields to be included in that toaster message because they are part of the page, regardless of the underlying logic.

Therefore, the purpose of creating this PR is to improve the UX on pages where the company request triggers but is not the company page. When a user receives the "updated_entity" message, it should indicate that everything is finished. When they return to the page, the custom fields should be the same as the ones they updated. I hope this clarifies the point better.

However, please note that the company updating works perfectly when it should be updated. The logic you created ensures that the request is only triggered when necessary. The issue lies in the UX when it is needed. We have personally experienced this problem so I created PR with this solution.

Please let me know your thoughts.

PR #851, opened 2023-06-27
[Feature] Add Task To Invoice

Civolilah:


src/pages/tasks/common/hooks/useFetchInvoicesForTasksAdding.ts
```diff
@@ -0,0 +1,93 @@
...
+}
+
+export function useFetchInvoicesForTasksAdding(params: Params) {
+  const queryClient = useQueryClient();
+
+  const setTasksToAddOnInvoice = useSetAtom(tasksToAddOnInvoiceAtom);
+
+  const { setInvoices, setIsAddTasksOnInvoiceVisible } = params;
+
+  return (tasks: Task[]) => {
```
Civolilah, line 36: @beganovich The idea of this hook is to use for both bulk and single task adding to invoice. So, that is the reason for accepting array instead of one task.

PR #852, opened 2023-06-27
Improve & migrate selectors 

beganovich:
This migrates the user & project selector to the new component.


turbo124:
This is great and ready for merge, but it looks like we are missing debounce for record count <=1

beganovich:
Added to Jira, I'll tackle it in the morning :)

PR #853, opened 2023-06-27
[Feature] Bulk Invoice Project

Civolilah:
@beganovich @turbo124 The PR includes implementation for bulk invoicing project functionality. Screenshot:

Screenshot 2023-06-27 at 20 56 22

@turbo124 Additionally, this PR includes improvements for UX as we discussed when saving an invoice while the tasks tab is selected. Now, when a user saves the invoice with `table=tasks` in the URL (for invoicing projects and similar functionalities), the tasks tab will also be selected on the edit page.

Let me know your thoughts.

### DIFF
--- a/src/components/DataTable.tsx
+++ b/src/components/DataTable.tsx
@@ -56,7 +56,10 @@ export type DataTableColumns<T = any> = {
   format?: (field: string | number, resource: T) => unknown;
 }[];
 
-export type CustomBulkAction = (selectedIds: string[]) => ReactNode;
+export type CustomBulkAction<T> = (
+  selectedIds: string[],
+  selectedResources?: T[]
+) => ReactNode;
 
 interface StyleOptions {
   withoutBottomBorder?: boolean;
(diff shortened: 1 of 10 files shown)


src/components/DataTable.tsx
```diff
@@ -233,6 +238,16 @@ export function DataTable<T extends object>(props: Props<T>) {
       });
   };
 
+  useEffect(() => {
+    if (props.includeObjectSelection && data) {
+      const filteredSelectedResources = data.data.data.filter((resource: any) =>
+        selected.includes(resource.id)
```
Civolilah, line 244: @beganovich I'm not sure if it is possible to pass the type checker when defining the type for the resource. However, I have tried using `as keyof typeof resource` or `as keyof T`, but it does not pass the checker. In this case, I believe it should be written as follows.

turbo124:
:fire: :fire: :fire:

beganovich (approved):
Looks good, thanks :+1:

PR #856, opened 2023-06-28
Fixes for nullables

turbo124:


turbo124:
Fixes edge case where entry prop is null

beganovich (approved):
Looks good, thanks :+1:

PR #857, opened 2023-06-28
Refactor send reminders toggle

turbo124:
This fixes an issue currently where we potentially send incorrect types for client settings.

If no setting is present, we should remove the prop from the settings object completely. 

This PR resolves the issue, unsure if this is the cleanest way to tackle the problem, but it is certainly safer than the current implementation.

### DIFF
--- a/src/pages/clients/edit/components/AdditionalInfo.tsx
+++ b/src/pages/clients/edit/components/AdditionalInfo.tsx
@@ -24,7 +24,7 @@ import { CustomFieldsPlanAlert } from '$app/components/CustomFieldsPlanAlert';
 import { DocumentsTable } from '$app/components/DocumentsTable';
 import Toggle from '$app/components/forms/Toggle';
 import { TabGroup } from '$app/components/TabGroup';
-import { set } from 'lodash';
+import { cloneDeep, set } from 'lodash';
 import { Upload } from '$app/pages/settings/company/documents/components';
 import { Field } from '$app/pages/settings/custom-fields/components';
 import {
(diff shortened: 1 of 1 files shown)


src/pages/clients/edit/components/AdditionalInfo.tsx
```diff
@@ -61,13 +60,43 @@ export function AdditionalInfo(props: Props) {
...
+
+    setErrors(undefined);
+
+    let value: any = event.target.value;
+
+    if (event.target.id === 'settings.send_reminders'){
+      value = event.target.value === 'enabled' ? true : false
+    }
+
+    setClient(set(client as Client, event.target.id, value));
```
beganovich, line 96: We can do better here.

I'd suggest following signature for this method:

```ts
const handleSettingsChange = (property: keyof Company['settings'], value: string | number) =>
```

This way we are not dealing with event directly and we can use `onValueChange` on select field.

src/pages/clients/edit/components/AdditionalInfo.tsx
```diff
@@ -173,21 +204,27 @@ export function AdditionalInfo(props: Props) {
...
+              defaultValue={Object.prototype.hasOwnProperty.call(props.client?.settings, 'send_reminders') ? props.client?.settings?.send_reminders ? 'enabled' : 'disabled' : ''}
+              className={
+                'appearance-none block px-3 py-1.5 text-base font-normal  text-gray-700 bg-white bg-clip-padding bg-no-repeat border border-solid border-gray-300 rounded transition ease-in-out m-0 focus:text-gray-700 focus:bg-white focus:border-blue-600 focus:outline-none'
               }
-            />
+              onChange={(handleChangeSettings)}
+            >
+              <option value=""></option>
+              <option value="enabled">{`${t('enabled')}`}</option>
+              <option value="disabled">{`${t('disabled')}`}</option>
```
beganovich, line 224: No need to escape into JSX, it's already interpolated.

```tsx
<option>{t('enabled')}</option>
```

beganovich (changes requested):
Looks good, few changes are needed :+1:

src/pages/clients/edit/components/AdditionalInfo.tsx
beganovich: We can push this even further with type safety. Here's demo: https://tsplay.dev/mxJDKw

turbo124: This still pushes the type checks into the method now that value is either string/number.

In settings the concept of a blank string unsets the prop from the object. but now that value is either string/number we cannot use .length so it does complicate the logic further down?

PR #862, opened 2023-06-29
Fixes for queries for adding expense to invoice

turbo124:
This fixes the queries we use to populate the expense to invoice modal.

One issue which I was unable to resolve was the Add To Invoice action from the list view does not carry the correct expense across.

image


beganovich (approved):
Looks good, thanks 👍

PR #863, opened 2023-06-29
Fixes for charts

turbo124:
The issue was:

```
setCurrency(currency ?? parseInt(currencies[0].value));
```

because we set
```
  const [currency, setCurrency] = useState(1);
```
currency is never null,

this PR resolves by switching the ternary around

For merge to main pls.


beganovich (approved):
Looks good, thanks 👍

PR #876, opened 2023-06-30
temporary disablement of send reminders config

turbo124:


beganovich (approved):
Looks good, thanks 👍️

PR #877, opened 2023-06-30
Remove tinymce context menu

turbo124:
prevents tinymce from overriding the browser context menu on right click.


beganovich (approved):
Looks good, thanks :+1:

PR #880, opened 2023-07-01
[Bugfix] Client Selector Additional Query Parameter

Civolilah:
@beganovich @turbo124 The PR includes resolving a bug when the ID value of the client is passed, but there is no selected client with that ID. This query parameter will bring an additional client, which should be selected. Let me know your thoughts.


beganovich:
Can you double check this is needed at all? ComboboxAsync has this built in and it *always* passes the ?with param.

https://github.com/invoiceninja/ui/blob/main/src/components/forms/Combobox.tsx#L326

PR #881, opened 2023-07-01
[Bugfix] Auto Bill Recurring Invoices Dropdown Removing Blank Option

Civolilah:
@beganovich @turbo124 The PR includes removing blank option for `Auto Bill Recurring Invoices Dropdown` on Payment Settings page. If we do not have any value returned from the API we will set `off` as default value. Screenshot:

Screenshot 2023-07-01 at 12 38 24

Let me know your thoughts.


beganovich (approved):
Looks good, thanks 👍️

PR #883, opened 2023-07-02
[Bugfix] Toaster On Settings Pages Fixing

Civolilah:
@beganovich @turbo124 The PR includes fixing toaster displaying on company settings pages. Let me know your thoughts.

### DIFF
--- a/src/pages/settings/common/hooks/useHandleCompanySave.tsx
+++ b/src/pages/settings/common/hooks/useHandleCompanySave.tsx
@@ -40,10 +40,10 @@ export function useHandleCompanySave() {
       return;
     }
 
-    const adjustedToaster =
+    const adjustedExcludeToaster =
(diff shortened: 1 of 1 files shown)


src/pages/settings/common/hooks/useHandleCompanySave.tsx
```diff
@@ -40,7 +40,10 @@ export function useHandleCompanySave() {
       return;
     }
 
-    !excludeToasters && toast.processing();
+    const adjustedExcludeToaster =
+      typeof excludeToasters === 'boolean' && !excludeToasters;
```
Civolilah, line 44: @beganovich On the company settings page, we did not pass the `excludeToaster` value because we do not need it. Instead, by default, the `event` object is passed from the form's side. That is the reason for this expression.

beganovich (approved):
Looks good, thanks 👍️

PR #889, opened 2023-07-03
[Cleanup] Stripe_Connect Message On Credentials Tab

Civolilah:
@beganovich @turbo124 The PR includes the implementation of additional UI for notifying the user that he needs to oauth their account with Stripe. Screenshot:

Screenshot 2023-07-05 at 20 55 28

This functionality is only available for `stripe_connect`, and it applies when the `gateway.config.account_id` is empty or does not exist.

Let me know your thoughts.


beganovich (approved):
Looks good, thanks :+1:

PR #904, opened 2023-07-05
[Cleanup] Refactoring Logic For Showing Bulk Actions & Fixing Bugs For Some Of Them

Civolilah:
@beganovich @turbo124 The PR contains bug fixes specifically for `more_actions` dropdowns and their actions. Additionally, it improves the structure of two related elements to make maintaining this part of the app easier in the future. While testing the functionality for which the ticket was created, I discovered additional bugs concerning the `more_actions` dropdowns. For example, there is a missing dropdown on the `Documents` tab of the Expense and Recurring Expense pages. Screenshot:

Screenshot 2023-07-05 at 18 04 07

This PR also resolves that issue. Since we have this dropdown for other edit pages on different tabs, I consider the absence of it on those two pages as a bug. I hope you agree with me as well.

Otherwise, let me know your thoughts.


src/pages/expenses/documents/Documents.tsx
```diff
@@ -51,12 +53,26 @@ export default function Documents() {
...
+      navigationTopRight={
+        expense && (
+          <ResourceActions
+            resource={expense}
+            label={t('more_actions')}
+            actions={actions}
+          />
+        )
+      }
+    >
```
Civolilah, line 75: @beganovich After reviewing this section of the code, I noticed that it could potentially be improved by creating separate components, `Expense.tsx` and `RecurringExpense.tsx`, to handle the tabs on the edit pages for these entities.

I believe we have implemented a similar approach wherever we have tabs on edit pages. This will ensure that we don't miss the `Save` button in the navigation bar on `Documents` tab for example and avoid duplicating the setting of the `navigationTopRight` property.

Although I didn't include this implementation in the current PR since it's unrelated, I wanted to recommend it.

@turbo124 If we consider the absence of the `Save` button on the `Documents` tab for those Edit pages as a bug, we could create a ticket to address it and make some minor improvements to the code implementation.

Please let me know your thoughts on this suggestion.

turbo124: @Civolilah yes, please create a ticket for this, save button should be present.

beganovich (approved):
Looks good, thanks 👍️

PR #907, opened 2023-07-06
[Bugfix] Tax Rate Selector Fixes

Civolilah:
@beganovich @turbo124 The PR resolves all known bugs with `TaxRateSelector`. Regarding the other task of converting all current selectors to Async, I have also implemented it here. Additionally, in order to address the specific requirements for `TaxRateSelector` and its values, I had to make slight adjustments to the `Combobox` component. Please let me know your thoughts.


src/components/tax-rates/TaxRateSelector.tsx
```diff
@@ -45,7 +45,7 @@ export function TaxRateSelector(props: Props) {
           props.onChange && props.onChange(record)
         }
         value="rate"
-        defaultValue={props.defaultValue === 0 ? null : props.defaultValue}
```
Civolilah, line 48: @beganovich Please let me know if there is a specific reason for this condition that we had, or maybe it was something required by an older version of the Combobox component.

src/components/forms/Combobox.tsx
```diff
@@ -284,6 +284,8 @@ interface EntryOptions<T = any> {
   label: string;
   value: string;
   labelFn?: (resource: T) => string | JSX.Element;
+  formatLabel?: (resource: T) => string;
```
beganovich, line 287: Is `labelFn` not enough here or?

src/components/forms/Combobox.tsx
```diff
@@ -339,8 +341,10 @@ export function ComboboxAsync<T = any>({
           response.data.data.map((entry) =>
             data.push({
               id: entry[entryOptions.id],
-              label: entry[entryOptions.label],
-              value: entry[entryOptions.value],
+              label:
+                entryOptions.formatLabel?.(entry) ?? entry[entryOptions.label],
+              value:
+                entryOptions.formatValue?.(entry) ?? entry[entryOptions.value],
```
beganovich, line 347: This is generally a bad idea. Can you let me know the reasons why we want to format the value? It's generally a pitfall. Looking at the current tax rate selector, I don't see any formatting of value.

src/components/forms/Combobox.tsx
Civolilah: @beganovich Hmm, I do think so. The `labelFn` is good as it returns a `JSX.Element`, which is very useful for formatting labels and options in the dropdown. However, the issue is that this function doesn't fit well in the following context:

`displayValue={(entry: Nullable<Entry>) => entry?.label ?? ''}`

`displayValue` property only accepts string but NOT the JSX.Element.

I think you used the `labelFn` for formatting options in the `ProductSelector`, but you still used `product_key` as the label when it is selected. In this case using `rate` or `name` properties will give us bad UX because I think if users select some kind of option value they will expect to see it in selected section.

Let me know your thoughts.

src/components/forms/Combobox.tsx
Civolilah: @beganovich The reason for doing this is actually specific `defaultValue` that should be passed in the `TaxRateSelector` component for example `${expense.tax_name1}${expense.tax_rate1}`. If we do not have formatted value when we pass this specific defaultValue nothing will be selected because the default value is `rate` property.

The actual part of the code where I did value formatting is this one: 

`formatValue: (taxRate) => ${taxRate.name}${taxRate.rate}`

I did label formatting because of the reason that I explained in the previous comment dicussion.

Actually I have some other ideas for formatting this but I thought that this one is the most clear and provide us better adjusting values.

Let me know your thoughts.

src/components/forms/Combobox.tsx
```diff
@@ -284,6 +287,7 @@ interface EntryOptions<T = any> {
   label: string;
   value: string;
   labelFn?: (resource: T) => string | JSX.Element;
+  displayValueFn?: (resource?: T) => string;
```
beganovich, line 290: I got the idea now, but the naming is wrong.

Can we rename `labelFn` to maybe `dropdownLabelFn` and `displayValueFn` to `inputLabelFn`? This way the situation is a bit more clear.

Civolilah: @beganovich Renamed!

PR #910, opened 2023-07-06
Bulk download pdfs

turbo124:
Implements bulk downloads across all entities


rephlex:
Are there any plans to merge this into main?

turbo124:
It already exists as a bulk action

PR #911, opened 2023-07-06
Bulk download pdfs

turbo124:
Implements bulk PDF downloads across all entities


beganovich (approved):
Looks great, thanks 👍️

PR #918, opened 2023-07-07
Invoice slideout

beganovich:
This implements slideout UX similar to the one in the Flutter application. It will allow us to have a clear distinction between the view and edit pages.


beganovich:
Ping @turbo124 for a QA pass.

Note: Activities are still WIP - do we wanna replicate/reuse the same logic as dashboard activities? It depends on all different entities, so I'm not sure if that's a way to go or we have a more efficient way.

beganovich:
Ping @turbo124 for review.

turbo124:
@beganovich this looks good, we can queue this for merge on Monday after the API changes have been released.

PR #919, opened 2023-07-07
[Bugfix] Ensuring All Fields On The App Showing 422 Error

Civolilah:
@beganovich @turbo124 The PR includes checking the validation throughout the entire app and adding it where it is missing or fixing it where it has been applied incorrectly.

The most important pages to mention are:

- Validation has been fixed on the Payment creation page
- Validation on some Settings pages have been missed or incorrectly applied, so both cases have been fixed.

I believe that if there are any other cases where the validation was missed, the reason is 99% likely that I do not know the logic behind that part of the app/validation.

I hope that now everything is correctly covered. If there are any specific logic cases, please let me know and I will customize the solution.

@turbo124 You probably created this ticket because you noticed missing validation somewhere. I hope that these fixes resolve all the cases you noticed.

Let me know your thoughts.


turbo124:
when editting a deleted client, the 400 error is not being displayed, just to note, you should be displaying all messages >400 && <500 http response code.

beganovich (approved):
Looks good, thanks :+1:

PR #921, opened 2023-07-08
Localization fixes

turbo124:
Resolves several localization issues. also attempts to dynamically import locale for antd components (atom type was unresolvable hence the => any type )


beganovich (approved):
Looks good, thanks 👍️

PR #924, opened 2023-07-11
usePreference

beganovich:
This creates a replacement for useState/useAtom with the same API, but it operates on the preferences/user level.

Example:

```ts
const [defaultView, setDefaultView] = usePreference('preferences.dashboard_charts.default_view')

setDefaultView('day' | 'week' | 'month') // has to be same type as default_view defined in ReactSettings
```

image

It is also completely type-safe with autocomplete for each setting.

image


turbo124:
lgtm!

PR #925, opened 2023-07-11
[Bugfix] Removing HTML Tags From Tooltips And Private/Public Notes Columns In Tables

Civolilah:
@beganovich @turbo124 The PR includes a solution for removing HTML tags from the string for the private/public_notes properties in the tooltips and tables. Screenshot:

Screenshot 2023-07-11 at 19 41 42

Let me know your thoughts.


beganovich (approved):
Looks good, thanks 👍️

PR #929, opened 2023-07-12
Export maps preferences

beganovich:
This integrates (and saves) the export maps variables into user preferences.


turbo124:
Final step here is to populate the "report_keys" array with the column values so that the report can use the saved keys.

beganovich:
Ping @turbo124 for another QA pass.

PR #944, opened 2023-07-13
[Cleanup] Applying Enable Modules Across The App Same As We Did For Quotes

Civolilah:
@beganovich @turbo124 The PR includes implementation `enable_modules` across the app. This affects the `Client show page (tabs)`, `Dashboard page (tables)`, `Project show page for the Tasks table`, and `Vendor show page (tabs)`. If a module is disabled, the corresponding table/tab will not be shown on the mentioned pages. 

I think that this PR cover all pages where we need implementation of enabling modules, but let me know if there are some additional. Since the implementation for enabling modules is already present in the main left sidebar, no additional changes are required there. 

Let me know your thoughts.

### DIFF
--- a/src/components/Tabs.tsx
+++ b/src/components/Tabs.tsx
@@ -10,8 +10,14 @@
 
 import { route } from '$app/common/helpers/route';
 import { useAccentColor } from '$app/common/hooks/useAccentColor';
-import { MouseEvent, useRef } from 'react';
-import { Link, Params, useLocation, useParams } from 'react-router-dom';
+import { MouseEvent, useEffect, useRef } from 'react';
+import {
+  Link,
+  Params,
+  useLocation,
+  useNavigate,
+  useParams,
+} from 'react-router-dom';
 
 interface Props {
   className?: string;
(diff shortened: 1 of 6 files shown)


src/components/Tabs.tsx
```diff
@@ -51,6 +59,18 @@ export function Tabs(props: Props) {
...
+    if (props.tabs.length) {
+      const doesDefaultUrlExist = props.tabs.some(
+        ({ href }) => href === location.pathname
+      );
+
+      if (!doesDefaultUrlExist) {
+        navigate(props.tabs[0].href);
+      }
+    }
+  }, []);
```
Civolilah, line 72: @beganovich Just quick explanation for this part.

We have a case where the first/default tab can be removed. For example, on the Client show page, the Invoices tab can be removed based on the logic in enabled_modules.

If we keep the default URL, none of the tabs will be selected, and the UI will be rendered incorrectly. Therefore, if the current URL upon mounting this component is not included in the array of tab hrefs, we will simply navigate to the first available tab.

beganovich (approved):
Looks good, thanks 👍️

PR #945, opened 2023-07-13
[Feature] Auto Billing Invoices Functionality

Civolilah:
@beganovich @turbo124 The PR includes implementation of `auto_billing` functionality for Invoices. The `auto_bill` option will be shown if three requirements are passed:

- The `balance` of invoice must be > 0
- The Invoice must be in one of two statuses `Send` | `Partial` (2,3)
- The Invoice's Client must have at least one "gateway_token" in the array of `gateway_tokens`

To display the bulk action in the dropdown, we can have two situations, if no invoices are selected, the option will be shown or if only one of the selected invoices passes those three checks, the action will also be shown in the dropdown.

`Important note`: Only the `invoice_ids` that are eligible for auto-billing will be included in the payload. Others will be excluded, even if they are selected during the bulk action.

Screenshots:

Screenshot 2023-07-13 at 19 41 22

Screenshot 2023-07-13 at 19 40 48

Let me know your thoughts.

### DIFF
--- a/src/common/queries/invoices.ts
+++ b/src/common/queries/invoices.ts
@@ -8,14 +8,18 @@
  * @license https://www.elastic.co/licensing/elastic-license
  */
 
-import { AxiosResponse } from 'axios';
+import { AxiosError, AxiosResponse } from 'axios';
 import { endpoint } from '$app/common/helpers';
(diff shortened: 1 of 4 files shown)


beganovich (approved):
Looks good, thanks 👍️

PR #951, opened 2023-07-16
send_reminder may not be set

turbo124:


beganovich (approved):
Looks good, thanks 👍️

PR #959, opened 2023-07-18
[Bugfix] Navigation Links For Quotes Tables On Dashboard

Civolilah:
@beganovich @turbo124 The PR includes bug fixes for incorrect navigation when a user clicks on the `number` link in the `number column` on the Quotes tables on the Dashboard. Previously, it would navigate to the `invoices edit page` instead of the quotes edit page. Now, this has been corrected, and it navigates to the quotes edit page as intended. Let me know your thoughts.


beganovich (approved):
Looks good, thanks 👍️

PR #968, opened 2023-07-20
Webhook overview for Payment Gateways

turbo124:
Implements webhook overview for payment gateways.


beganovich (approved):
Looks good, thanks 👍️

PR #970, opened 2023-07-21
[Improvement] Combobox Async Component | Weekly Challenge

Civolilah:
@beganovich @turbo124 As per David's recommendation during Monday's meeting, I created a PR to improve a any part of the app. My choice was to enhance the Combobox component, specifically by improving the UX when the user clicks on an already selected option in the dropdown. 

So, currently in the app, if the user clicks on already selected value in the dropdown, the selection icon on the right side will hide, and the text will look like all the others (unselected), but nothing will happen with the value. As a result, in the dropdown, it looks like the specific value is unselected, but actually, it is not.

I recorded two short videos to show the difference between the old behavior and the new one.

Old:

https://github.com/invoiceninja/ui/assets/51542191/366a34a7-3cd3-46d7-b615-86aeec1848ce

New:

https://github.com/invoiceninja/ui/assets/51542191/c1ce0b09-c186-4e3e-b392-9ab0e73ad4a7

Let me know your thoughts.

### DIFF
--- a/src/components/forms/Combobox.tsx
+++ b/src/components/forms/Combobox.tsx
@@ -110,6 +110,16 @@ export function ComboboxStatic({
     [filteredValues]
   );
 
+  const handleChangeValue = (value: Entry | null) => {
+    if (value) {
+      if (selectedValue && value.value === selectedValue.value) {
+        onDismiss && onDismiss();
+      } else {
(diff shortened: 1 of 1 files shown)


beganovich (approved):
Looks good, thanks v

PR #976, opened 2023-07-24
Hide removed users from user management

turbo124:


beganovich (approved):
Looks good, thanks 👍️

PR #983, opened 2023-07-25
[Bugfix] Money Formatting

Civolilah:
@beganovich @turbo124 The PR includes fixes across the whole app regarding the formatting of money. We had some cases where we did not have a fallback to company settings, some cases where formatting was not correctly applied, and some other cases. So, currently, this PR should always provide the fallback to company settings and fix all the issues we had. 

Additionally, the PR includes improving the structure for falling back to company settings to avoid a lot of repetition.

Let me know your thoughts.

### DIFF
--- a/src/common/queries/expenses.ts
+++ b/src/common/queries/expenses.ts
@@ -61,6 +61,7 @@ export function useExpenseQuery(params: ExpenseParams) {
 interface ExpensesParams extends Params {
   enabled?: boolean;
   matchTransactions?: boolean;
+  include?: string;
 }
 
 export function useExpensesQuery(params: ExpensesParams) {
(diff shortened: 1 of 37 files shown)


src/common/queries/expenses.ts
```diff
@@ -70,14 +71,15 @@ export function useExpensesQuery(params: ExpensesParams) {
       request(
         'GET',
         endpoint(
-          '/api/v1/expenses?filter=:filter&per_page=:per_page&status=:status&page=:page&match_transactions=:match_transactions&includes=:includes',
+          '/api/v1/expenses?filter=:filter&per_page=:per_page&status=:status&page=:page&match_transactions=:match_transactions&includes=:includes&include=:include',
```
turbo124, line 74: why do we have `includes=:includes`

i don't think is a valid parameter at all?

Civolilah: @turbo124 Yeah, I noticed the same thing. I didn't want to touch anything there because I didn't know if it's a valid parameter or not. Let me know if I should remove it.

Civolilah: @turbo124 We never use this param so it is removed.

beganovich (approved):
Looks good, thanks 👍️

PR #991, opened 2023-07-26
Fixes for nan in tax rate selectors

turbo124:


beganovich (approved):
Looks good, thanks 👍️

PR #996, opened 2023-07-26
[Bugfix] Activate Company Link Breaks App

Civolilah:
@beganovich @turbo124 The reason we had a broken app when the user clicked the "activate_company" button on the banner is that we did not have optional chaining in the "set_default_company" logic. So, there was a moment when the "company" object was undefined, and we tried to read "id" and "settings.name" from that undefined value (`company.id` and `company.settings.name`).

The reason you, David, were not able to see the "activate_company" banner locally is that we show the banner only for the `HOSTED` platform. Please let me know if I should remove this constraint and display the banner for both types of platforms.

Let me know your thoughts.

### DIFF
--- a/src/pages/settings/account-management/component/Overview.tsx
+++ b/src/pages/settings/account-management/component/Overview.tsx
@@ -90,7 +90,7 @@ export function Overview() {
         />
       </Element>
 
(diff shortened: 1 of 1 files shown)


beganovich (approved):
Looks good, thanks 👍️

PR #1002, opened 2023-07-27
Include calculated hours in the datatable

beganovich:
This will show relative hours to the current time if the task is running.


turbo124:
Looks great!!

PR #1007, opened 2023-07-28
Passing `update_in_stock_quantity` after updating stock qty

beganovich:
This passes `update_in_stock_quantity` to the backend once the stock qty changes.


turbo124:
Looks good, thanks!!

PR #1008, opened 2023-07-29
[Feature] Client, Product, Recurring Invoice Bulk Actions Implementation

Civolilah:
@beganovich @turbo124 The PR includes the implementation of all missing bulk/regular actions for Client, Product, and Recurring Invoice entities. Additionally, the PR includes the implementation of new actions that we have on the Flutter app. Let me know your thoughts.


src/common/queries/documents.ts
```diff
@@ -28,3 +29,16 @@ export function useDocumentsQuery(params: Params) {
...
+    toast.processing();
+
+    request('POST', endpoint('/api/v1/documents?per_page=100'), {
+      action,
+      ids,
+    })
+      .then(() => toast.success('exported_data'))
+      .finally(() => onActionCall?.());
+  };
+};
```
beganovich, line 44: I think this is an antipattern. You should be returning empty `Promise` at least letting action know it's successful rather than taking your own scope of callback fn.

src/components/DataTable.tsx
```diff
@@ -269,6 +270,14 @@ export function DataTable<T extends object>(props: Props<T>) {
...
   };
 
+  const onBulkActionCall = () => {
+    if (mainCheckbox.current) {
+      mainCheckbox.current.checked = false;
+
+      setSelected([]);
+    }
+  };
+
```
beganovich, line 280: Can you explain why do we need this one?

beganovich:
Ping @Civolilah for comments.

src/common/queries/documents.ts
Civolilah: @beganovich Until now, we've had a situation where even after the successful completion of a specific bulk action, the selected entities remained checked. What I wanted to achieve with this is to improve the UX during the execution of bulk actions. So, when a certain action is performed, we will perform a deselection of all selected entities. In this case, the function `onActionCall` is essentially a reference to this function:

```
 const onBulkActionCall = () => {
    if (mainCheckbox.current) {
      mainCheckbox.current.checked = false;

      setSelected([]);
    }
  };
```

Actually, it's the same thing you did for `archive/restore/delete` bulk actions in the DataTable component under the `finally` block. However, the issue is that the execution of custom bulk actions is not available under this component:

```
.finally(() => {
        queryClient.invalidateQueries([props.endpoint]);
        queryClient.invalidateQueries([apiEndpoint.pathname]);

        setSelected([]);
      });
```

The `onBulkActionCall` is passed through the `bulkAction` function and is available for use in `customBulk` hooks to perform bulk actions.
  
So, once the action is done, deselecting elements will make it much easier to continue using bulk actions. That's sort of my idea.

Now I'm getting the idea that maybe this could be achieved through a `custom event`, or through `atoms`, or via `Redux`.

What are your thoughts on those additional implementation methods?

src/components/DataTable.tsx
Civolilah: @beganovich I believe I've covered the function's purpose quite extensively in the comment above, but I'd like to add a few points here as well. I hope my previous comment hit my main idea. Essentially, this function performs the deselection of previously selected entities once the request is completed, regardless of whether it was successful or ended in an error. This approach significantly simplifies the use of bulk actions, particularly when users need to carry out a series of actions successively involving different entities.

Therefore, this function will be passed alongside each `bulkAction` and will be accessible for use within `customBulkAction` hooks. This involves passing it to every callback function within the custom bulk actions hooks:

```
{props.customBulkActions &&
              props.customBulkActions.map(
                (bulkAction: CustomBulkAction<T>, index: number) => (
                  <div key={index}>
                    {bulkAction(selected, selectedResources, onBulkActionCall)}
                  </div>
                )
              )}
```

I did it this way to avoid making things too global. I think there's flexibility for changes. Another option might be using Redux, custom events, or atoms in a similar way. The aim is to make the deselection of selected entities on the table more widespread.

However, I'm quite certain that invoking some form of action to perform deselection under the "finally" block of every bulk action where we want to do deselection is necessary.

What do you think?

beganovich (approved):
Thanks!

PR #1017, opened 2023-08-02
[Bugfix] Live Preview Payload Error

Civolilah:
@beganovich @turbo124 The PR includes setting the `id` property for all cloning options to an empty string, in accordance with the logic on the API. Let me know your thoughts.


beganovich (approved):
Looks good, thanks 👍️

PR #1021, opened 2023-08-05
[Bugfix] Default Quantity Functionality

Civolilah:
@beganovich @turbo124 The PR includes implementing functionality for the `default_quantity` property in company settings. I believe that we have now replicated the behavior of the Flutter app in the React app.

One note regarding the video in the Jira ticket, David. I believe the product you selected in the video had a quantity of `1`, which is why you ended up choosing a quantity of `1`, even though the `default_quantity` was set to `false`.

Of course, we now have the entire functionality implemented, but for that specific reason, I believe that was the case.

Let me know your thoughts.


turbo124:
Looks good, thanks!

beganovich (approved):
Looks good, thanks 👍️

PR #1026, opened 2023-08-08
Fixes for token creation permissions

turbo124:
Changes permissions from Owner to Admin to create tokens.


beganovich (approved):
Looks good, thanks 👍️

PR #1039, opened 2023-08-12
[Feature] Payments | Quotes | Credits | Bulk Actions

Civolilah:
@beganovich @turbo124 The PR includes implementation of bulk actions for Payment, Quote and Credit entites. Let me know your thoughts.


src/common/queries/credits.ts
```diff
@@ -16,25 +16,46 @@ import { route } from '$app/common/helpers/route';
...
-        queryClient.invalidateQueries(route('/api/v1/credits/:id', { id }));
+        ids.forEach((id) =>
+          queryClient.invalidateQueries(route('/api/v1/credits/:id', { id }))
+        );
 
         invalidateQueryValue &&
           queryClient.invalidateQueries([invalidateQueryValue]);
-      });
+      })
+      .finally(() => onActionCall?.());
```
beganovich, line 59: Isn't this old pattern we decided not to go with?

Civolilah: @beganovich Ah, sure. Actually, it was only used to close the modal. However, the solution has now been updated to avoid using this pattern. Let me know your thoughts.

beganovich (approved):
Looks good, thanks.

Just a tip for next time, please split the same feature impl for diff resources in diff prs. This makes it a lot easier to go thru, thanks.

PR #1042, opened 2023-08-15
Fixes for combo box logic

turbo124:
instead of searching when the result list hits 0, we increment this to 3, so that we preempt the search results, improves user experience when the string partially matching existing records, but the exact record is not shown.


beganovich (approved):
Looks good, thanks 👍

PR #1046, opened 2023-08-15
[Bugfix] Print PDF | Firefox

Civolilah:
@beganovich @turbo124 The PR addresses a bug related to printing PDFs in the Firefox browser. The reason for encountering blank PDFs during printing was the inconsistent behavior of `document.body.appendChild` across various browsers. Particularly in the case of `Firefox`, there was a slight delay when appending the `iframe` element into the `body` of the `document`, leading to the issue of blank PDFs. The problem has been resolved by utilizing an asynchronous approach to ensure the proper appending of this element within the document's body. Here is the screenshot:

Screenshot 2023-08-15 at 18 53 53

Let me know your thoughts.

### DIFF
--- a/src/pages/invoices/common/hooks/usePrintPdf.ts
+++ b/src/pages/invoices/common/hooks/usePrintPdf.ts
@@ -17,8 +17,20 @@ interface Props {
   entity: 'invoice' | 'quote' | 'credit' | 'purchase_order';
 }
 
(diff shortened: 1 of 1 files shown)


turbo124:
Looks good, thanks!

beganovich (approved):
Thanks!

PR #1051, opened 2023-08-16
Fixes for first line selecting

beganovich:
This allows us to perform a subset search using an additional field.


turbo124:
LGTM thanks!

PR #1062, opened 2023-08-20
[Bugfix] Invoice Slider Opening Twice In A Row

Civolilah:
@beganovich @turbo124 The PR includes the same bug that David noticed with the Recurring Invoice Slider. When a user attempts to open the slider twice in succession for the same invoice, the slider will appear blank. This PR resolves this issue. Let me know your thoughts.


turbo124:
@Civolilah 

For some reason, in this PR, recurring invoice slider does not open at all now.

beganovich (approved):
Thanks!

PR #1071, opened 2023-08-22
Report previews

beganovich:


beganovich:
@turbo124 the following API changes are needed:

1. Additional *value* field. Example:

For credits, we have a credit number. We currently ship `display_value` which is 0001 and `value` which is `0001`. For me to create a proper link to the resource and make it clickable in the table, we need an additional `resource_value` field with hashed ID to the credit.

2. Column identifiers should be array of objects, following this interface:

```ts
interface Column {
  identifier: string,
  display_value: string,
}
```

for example:

```
[
  { identifier: 'balance', display_value: 'Credit balance' }
}
```

This lets us sort by identifier, instead of index. At the moment, we can make sorting by index but it's impossible to pass this data back to backend reliably (we would have to say, column index 3 was sorted by ASC, which is bad IMHO).

3. Ship columns and report data as separate JSONs, preferably with the following interface:

```ts
interface PreviewRecord {
    entity: string;
    id: string;
    hashed_id: null;
    value: string;
    display_value: string | JSX.Element | number;
}

// This is what ideal response would look like:

{
    columns: { identifier: string, display_value: string }[],
    data: PreviewRecord[][]
}
```

turbo124:
1. If no identifier is present it is safe to assume it is not needed to be linked.
 2. All sorting should be done on front end, we will never push another query to backend for filtering
 3. I'm not sure I follow this one, perhaps best to catch up on Slack to discuss

beganovich:
Ping @turbo124 for the first QA pass. For now, it's hardcoded to credits, as I am waiting for API to support this globally.

turbo124:
@beganovich should filtering be working? when i start to type, results don't appear and removing input does not restore the preview list

turbo124:
@beganovich 

This looks good from my testing. I am adjusting the API, and then will increase the coverage over the other entities.

My only suggestion would be for a better Toast message as the red x is misleading,

beganovich:
@turbo124 can you please post here once API is done for this part so I can update the PR? Thanks!

turbo124:
API is updated, but not yet in production

beganovich:
Ping @turbo124 for QA passa.

beganovich:
Ping @turbo124 for QA pass.

beganovich:
Ping @turbo124 for QA pass.

beganovich:
Ping @turbo124 for QA pass.

beganovich:
Ping @turbo124 for QA pass.

turbo124:
Please hold merging until we update API. we will need to sync on this.

PR #1074, opened 2023-08-28
Color schemes

beganovich:
This implements color schemes (and dark mode too).


beganovich:
First preview:

image

PR #1078, opened 2023-08-30
[Feature] Custom Bulk Actions Implementation | Recurring Expense

Civolilah:
@beganovich @turbo124 The PR includes the implementation of custom bulk actions for the `Recurring Expense` entity.

I noticed a small bug in the Flutter app while comparing. When stopping and starting a recurring expense, there is actually a message for a recurring invoice.

Note: We are missing two keywords in the translation files: `started_recurring_expense` and `stopped_recurring_expense`. If you agree with me, please just add them.

Let me know your thoughts.

### DIFF
--- a/src/common/queries/documents.ts
+++ b/src/common/queries/documents.ts
@@ -34,7 +34,7 @@ export const useDocumentsBulk = () => {
   return (ids: string[], action: 'download') => {
     toast.processing();
 
-    request('POST', endpoint('/api/v1/documents?per_page=100'), {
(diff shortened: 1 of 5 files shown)


turbo124:
route document download is incorrect

should be documents/bulk

Civolilah:
> route document download is incorrect
> 
> should be documents/bulk

@turbo124 Fixed!

beganovich (approved):
Thanks!

PR #1081, opened 2023-08-31
Align combobox behavior with Flutter app

beganovich:
If we type new, previously unregistered products, this would leave existing line items.


beganovich:
Closed in favor or #1092 .

PR #1089, opened 2023-09-04
[Cleanup] Payment Interfaces Adjustment

Civolilah:
@beganovich @turbo124 The PR includes improved specifications for the Payment object. The reason for this change is to provide a better error specification when Sentry informs us about an error in a very rare edge case that we encountered recently. Furthermore, unnecessary properties have been removed from the `invoices` and `credits` properties. Let me know your thoughts.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -108,6 +108,8 @@ export function App() {
     }
   }, [company]);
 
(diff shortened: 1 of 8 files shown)


beganovich (approved):
Thanks!

PR #1091, opened 2023-09-05
[Cleanup] User Details Preferences Endpoint Adjusting

Civolilah:
@beganovich @turbo124 The PR includes adjustments to the endpoint on the User Details page regarding the user's permissions. So, if the user `IS an admin`, the endpoint will be standard: `'/api/v1/users/:id?include=company_user'`, but if he `IS NOT an admin`, the endpoint will be: `'/api/v1/company_users/:id/preferences'`. Let me know your thoughts.


turbo124:
closed in favour of solution API side

PR #1093, opened 2023-09-07
Import CSV hints

turbo124:
if hints exist in the preimport - these are used to select / match the corresponding fields.


beganovich (approved):
Thanks!

PR #1096, opened 2023-09-07
[Cleanup] Payment Creation Page && Combobox Selector Improvements

Civolilah:
@beganovich @turbo124 The PR includes a couple of improvements:

- The dropdown of the combobox component closes after selecting a value
- Unable to select the same invoice/credit twice from the dropdown menus on the payment creation page
- Setting the `invoices` and `credits` properties in the Payment object to an empty array after clearing the value from the Client combobox
- Resolving a bug with classes in the Combobox Component (as explained in the comment)

Let me know your thoughts.

### DIFF
--- a/src/components/forms/Combobox.tsx
+++ b/src/components/forms/Combobox.tsx
@@ -112,6 +112,7 @@ export function ComboboxStatic({
   );
 
   const comboboxRef = useRef<HTMLDivElement>(null);
+  const comboboxInputRef = useRef<HTMLInputElement>(null);
 
   useClickAway(comboboxRef, () => {
     setIsOpen(false);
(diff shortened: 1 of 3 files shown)


src/components/forms/Combobox.tsx
```diff
@@ -286,12 +294,8 @@ export function ComboboxStatic({
...
                   key={entry.id}
                   value={entry}
-                  className={() =>
-                    classNames(
-                      'min-w-[19rem] relative cursor-default select-none py-2 pl-3 pr-9'
-                      // active ? 'bg-gray-100 text-gray-900' : 'text-gray-900'
-                    )
-                  }
+                  className="min-w-[19rem] relative cursor-default select-none py-2 pl-3 pr-9"
+                  // active ? 'bg-gray-100 text-gray-900' : 'text-gray-900'
```
Civolilah, line 298: @beganovich We had a weird result in the class list due to the callback function. Screenshot:

Screenshot 2023-09-08 at 10 13 06

We actually do not need the `classNames` function in this case, so I just removed it to avoid this behavior. Let me know your thoughts.

Civolilah:
@turbo124 Just to confirm, the edge case that you mentioned in the video on the ticket has now been implemented in the PR. Let me know your thoughts.

beganovich (approved):

PR #1098, opened 2023-09-08
Fixes for types for templates and reminders

### LINKED ISSUE #1097: Templates and Reminders: Should not allow an empty value in the field "Days" for reminders
## Issue description
When you configure a reminder using the ReactUI, the form allows you to delete what's in the "Days" field and save the parameters. On the server side, there is no validation either and an empty string is saved in the database. However, the flutter app expect to receive an integer and so it throws a deserialization error which prevent you from login.

## Expected behavior
The "Days" field should either be set as required or, like in the flutter app, be set to 0 if it's empty.

turbo124:
closes #1097 

This will always set 0, but what is the field is already blank @beganovich ?


beganovich (approved):
Thanks!

PR #1108, opened 2023-09-13
Fixes item calculations with large discounts

turbo124:
Fixes the logic in calculations for inclusive/exclusive taxes when large discounts are applied to invoices.


beganovich (approved):
Thanks!

PR #1109, opened 2023-09-13
Centralised place for user preferences

beganovich:
This will let us see all user preferences and dismiss them accordingly (revert back to defaults).


turbo124:
@beganovich 

I don't see this in the UI yet?

image

PR #1110, opened 2023-09-13
[Cleanup] Increase Prices Action | Percent Field Changing Type To Number

Civolilah:
@beganovich @turbo124 The PR includes changing the type for the `percent` field on the "Increase Prices" action for recurring invoices. This will require the user to enter a numeric value instead of allowing the possibility of a string value. Screenshot:

Screenshot 2023-09-13 at 19 24 06

Let me know your thoughts.


beganovich (approved):
Thanks!

PR #1118, opened 2023-09-16
[Cleanup] Vendor Contact Custom Fields Implementation

Civolilah:
@beganovich @turbo124 The PR includes the implementation of missing Vendor Contact Custom Fields. Screenshots:

Screenshot 2023-09-18 at 04 19 37

Screenshot 2023-09-18 at 04 20 14

Let me know your thoughts.

### DIFF
--- a/src/pages/vendors/create/Create.tsx
+++ b/src/pages/vendors/create/Create.tsx
@@ -28,6 +28,8 @@ import { useDispatch } from 'react-redux';
 import { useNavigate } from 'react-router-dom';
 import { Form } from '../edit/components/Form';
 import { useAdmin } from '$app/common/hooks/permissions/useHasPermission';
+import { VendorContact } from '$app/common/interfaces/vendor-contact';
+import { set } from 'lodash';
 
 export default function Create() {
   const [t] = useTranslation();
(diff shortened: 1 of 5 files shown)


src/pages/vendors/create/Create.tsx
```diff
@@ -54,36 +56,27 @@ export default function Create() {
...
-            is_primary: false,
-            phone: '',
-            custom_value1: '',
-            custom_value2: '',
-            custom_value3: '',
-            custom_value4: '',
-            link: '',
-            last_login: 0,
-          },
-        ],
```
beganovich, line 81: Why are we decoupling the contacts from the main object?

Civolilah: @beganovich Hmm, for the logic regarding Vendor Contacts, I borrowed the logic from the Client Contacts. In fact, we have a state for contacts that handles the entire logic. Then, on save, we simply transfer what we have in the contacts state into the main object. Both approaches look similar, but I adopted this one from the Client Contacts. Perhaps you'll prefer this one more. However, to avoid any potential edge cases that could break functionality, I've removed it. The logic should reside either in the contacts state or in the main object. Please let me know which approach you prefer.

beganovich (approved):
Thanks!

PR #1120, opened 2023-09-17
Fixes for Line Item DnD

turbo124:
Prevent line items less than zero from corrupting DnD indexes


beganovich (approved):
Amazing work, thanks!

PR #1124, opened 2023-09-18
[Cleanup] Changing Types of Input Fields to type=

Civolilah:


turbo124:
@Civolilah 

There is an issue here with any of the "settings" fields - if we force the field to a number, then the default goes from unset to Zero - which we do not want to do.

Can we please remove from the PR any props in settings and then do another PR to handle number inputs for settings please.

src/pages/clients/edit/components/AdditionalInfo.tsx
```diff
@@ -185,7 +185,8 @@ export function AdditionalInfo({ client, errors, setClient }: Props) {
           <Element leftSide={t('task_rate')}>
             <InputField
               id="settings.default_task_rate"
-              value={client?.settings?.default_task_rate || ''}
+              type="number"
+              value={client?.settings?.default_task_rate || 0}
```
turbo124, line 189: This has not been adjusted?

turbo124 (changes requested):
Need to adjust the default task rate

src/pages/clients/edit/components/AdditionalInfo.tsx
Civolilah: @turbo124 Adjusted!

beganovich (approved):
Thanks!

PR #1134, opened 2023-09-21
Combobox: Handling on blur event

beganovich:
This improves combobox a11y and aligns it even more with Flutter experience.


turbo124:
@beganovich  assuming we can close  this in favour of the new incoming component?

beganovich:
Yep :)

PR #1143, opened 2023-09-26
Scaffolds for templates

turbo124:


beganovich:
Ping @turbo124 for a QA pass.

- We should be able to pick the resource on the template create/edit page. I've only put the "Invoice" for now, perhaps better wording is a delivery note - we can iterate.

- It would be good to put here all possible values for "entities" property on the design.

- I've fixed the issue with the page going white screen, _sometimes_.

- I've fixed the issue with the return key making full page submission.

- We need to implement filtering the designs by template (done) and entity, eg: `/api/v1/designs?template=true&entity=invoice`, so we can filter properly.

- Need to decide what happens when delivery note design changes - do we update company settings instantly or we trigger it on btn click.

beganovich:
Ping @turbo124 for QA pass.

beganovich:
Ping @turbo124 for QA.

turbo124:
@beganovich we can close this one now?

PR #1146, opened 2023-09-27
Apple login

turbo124:


beganovich:
@turbo124 please update me on status of this and if we should move forward on this one.

turbo124:
@beganovich yes, we should complete this. My guess is we just need testing?

beganovich:
I've replaced package with one that's maintained. Please QA it when you get a chance :+1:

PR #1149, opened 2023-09-28
[Bugfix] Property Value '0' When All Gateways Are Removed | Group & Client Settings

Civolilah:
@beganovich @turbo124 The PR includes adding logic to cover the edge case of setting the `company_gateway_ids` property to '0' when all gateways are removed from group/client settings. Let me know your thoughts.


beganovich (approved):
Thanks!

PR #1152, opened 2023-09-29
Scope pasting data images

beganovich:
This allows enterprise users to past and inline data images.


turbo124:
It looks like there is a payload size that the tinymce can handle, anything over 20kb kills tiny.

We may want to disable this functionality entirely until we can support passing through via the tinymce api to the doc storage.

beganovich:
I disabled the smart paste, let me know if this is good to merge, @turbo124.

turbo124:
Looks good.

PR #1156, opened 2023-10-01
Add languages to user model

turbo124:
Allows users to define their own language for the admin portal. Please note the changes in app.tsx and advise whether safe to use this style.


src/pages/settings/invoice-design/pages/template-designs/TemplateDesigns.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { Inline } from '$app/components/Inline';
+
+export default function TemplateDesigns() {
+    return (
+        <DataTable
+            endpoint="/api/v1/designs?template=true"
+            columns={[
+                {
+                    id: 'name',
+                    label: 'Name',
```
beganovich, line 22: Does this work? We use labels more like placeholder keys. Like, $model[$label]. Not sure if an uppercased one works?

src/pages/settings/invoice-design/pages/template-designs/pages/edit/common/hooks.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export function useDesignUtilities() {
+    const [payload, setPayload] = useAtom(payloadAtom);
+
+    const handlePropertyChange = (
+        property: keyof Design,
+        value: string | number | Parts
+    ) => {
+
+        console.log("Ddss");
```
beganovich, line 23: Needs to be removed.

PR #1160, opened 2023-10-03
[Cleanup] Correcting Route For Downloading Invoice Documents

Civolilah:
@beganovich @turbo124 The PR includes changing the logic for downloading invoice documents and aligning it with the rest of the project. So, now we use the POST route `/api/v1/documents/bulk` with the action: `download` and document IDs in the payload. Let me know your thoughts.


beganovich (approved):
Thanks!

PR #1166, opened 2023-10-05
[Cleanup] 2FA Workflow When a User Changes Their Phone Number and Saves It

Civolilah:
@beganovich @turbo124 The PR includes the addition of an additional flow when a user changes their `phone number` in "User Details" -> "Phone" and saves it. With minimal effort, I reused the logic we have for enabling Two-Factor Authentication in "User Details" -> "Two-Factor Authentication." If a user changes their phone number after saving the "User Details," we will open a modal for `SMS code confirmation` or a modal for `QR code selection`. If the user doesn't want to verify the phone number at that moment, they can simply close the modal and continue using the app. Please share your thoughts on adding this additional flow.


turbo124:
@beganovich can we merge / push to beta for some UAT before merge to main pls.

beganovich:
@turbo124 I can merge this. Should we put it behind the flag so we it doesn't show in prod?

turbo124:
@beganovich sure, sounds good. assuming we can UAT with the flag in place?

beganovich:
@Civolilah please let me know if it's fairly easy to put this behind dev flag, if not, I'll take care of it.

Civolilah:
> @turbo124 does self-host flag work?
> 
> If not, I'd suggest simple env flags.
> 
> `VITE_FEATURE_2FA` in env.

@beganovich @turbo124 I think we have it on both `SELF-HOSTED` and `HOSTED`. The only difference is that hosted users have an additional SMS code check, but self-hosted users do not.

If there is no need for a specific variable in the environment files for this purpose, can I simply check if the environment is `production` to disallow the actual feature?

Let me know your thoughts.

beganovich:
I think safest way to go about this is build flag.

`VITE_FEATURE_2FA`

I'll update the beta deployment once we have this updated.

Civolilah:
> I think safest way to go about this is build flag.
> 
> `VITE_FEATURE_2FA`
> 
> I'll update the beta deployment once we have this updated.

@beganovich Sure, I just updated the PR with this environment variable, so if `VITE_FEATURE_2FA` is set to `true`, we will have the 2FA feature in our app, if it is not set, we will completely lack this feature. Let me know your thoughts.

beganovich:
This is live on https://beta.invoicing.co 👍️

PR #1170, opened 2023-10-06
Improve search

beganovich:
This improves search by including by increasing searchable scope and returning 100 records (instead of previously 50).


turbo124:
It looks like something has happened to the search results here, when searching the invoice links are being presented as client links?

beganovich:
Hm, not sure. Search now just includes both value/label filtering. Can you post an example?

turbo124:
Looks good, we can push this back to main now please.

PR #1174, opened 2023-10-08
[Cleanup] Missing Project Custom Fields Implementation && Fixing Bug For Switch Custom Field Across the App 

Civolilah:
@beganovich @turbo124 The PR includes an implementation for custom fields for Projects (Creation and Edit pages). Additionally, I found a bug in the Switch (Checkbox) custom fields, where we had cases where strings ('false' or 'true') were being set as the field's value instead of a boolean value. I have fixed that as well. Screenshot:

Screenshot 2023-10-08 at 17 39 56

Let me know your thoughts.

### DIFF
--- a/src/components/forms/InputCustomField.tsx
+++ b/src/components/forms/InputCustomField.tsx
@@ -52,7 +52,14 @@ export function InputCustomField(props: Props) {
       )}
 
       {type === AvailableTypes.Switch && (
-        <Toggle onChange={props.onValueChange} checked={props.defaultValue} />
+        <Toggle
+          onChange={props.onValueChange}
+          checked={
+            typeof props.defaultValue === 'string'
+              ? props.defaultValue === 'true'
+              : props.defaultValue
+          }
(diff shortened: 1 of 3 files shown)


src/components/forms/InputCustomField.tsx
```diff
@@ -52,7 +52,14 @@ export function InputCustomField(props: Props) {
       )}
 
       {type === AvailableTypes.Switch && (
-        <Toggle onChange={props.onValueChange} checked={props.defaultValue} />
+        <Toggle
+          onChange={props.onValueChange}
+          checked={
```
Civolilah, line 57: @beganovich I guess that due to some other cases involving different custom fields, we consistently sent `value.toString()` as the value to custom_fields. However, in this situation, even when we set it to "false," we were consistently receiving a "true" value (checked) instead of "unchecked." So, I have addressed this particular case here.

beganovich (approved):

PR #1178, opened 2023-10-09
20231010 resolve contact in activity

turbo124:


turbo124:
Not sure why we never included this, i see it was stubbed but not implemented... was there an issue that prevented us from including the contact?

beganovich (approved):

beganovich:
Probably just overlooked it.

PR #1183, opened 2023-10-10
Improvements on refunds

beganovich:
This features the following improvements on the refund screen:
- Filterable select
- [ ] Fixes for weird handling of remaining refundable balance


turbo124:
@beganovich 

I think this is now addresses here? 

#1171

beganovich:
Um.. I don't think so?

image

PR #1189, opened 2023-10-14
[Cleanup] Clickability On Rows | Improvement

Civolilah:
@beganovich @turbo124 The PR includes an improvement in clickability for rows across all tables. Now, the entire `<td>` related to the Checkbox will be clickable for checking/unchecking the Checkbox. Additionally, sliders will now only open on the first three columns in a row after the Checkbox column. Let me know your thoughts.


beganovich (approved):
Thanks!

beganovich (approved):
Thanks!

PR #1194, opened 2023-10-15
Searchable selects

beganovich:
This implements searchable selects for currencies, languages, and timezones throughout the app.


turbo124:
If an item is already selected - when we click again (onFocus), can we highlight entire field so that if keystrokes are added, it is assumed the user wishes to replace the current selection.

PR #1199, opened 2023-10-16
Color schemes

beganovich:
This moves the light/dark mode switch into user preferences and makes it available for general usage.

image


beganovich:
Done!

image

PR #1208, opened 2023-10-17
Products table: Move dnd handler to the button

beganovich:
This moves the dnd handler of the products table to the button on the left.

More:
- Fixes for live preview/re-render triggering if we didn't actually move the row
- Fixes for table vertical space acting weird while on dragging


beganovich:
Hey @turbo124 I think feedback is missing in Jira related to the pending revision, please let me know about it.

turbo124:
@beganovich i think feedback is merged with 1207 in jira

beganovich:
Ping @turbo124 for a QA pass.

PR #1209, opened 2023-10-17
[Bugfix] Product Selector Clear Button Fix | Combobox Component Issue

Civolilah:
@beganovich @turbo124 The PR includes fixing the issue with the Combobox component. When the component is nullable, there is no case where the Down arrow icon is displayed, even if no specific value is selected in the Product Combobox. Now, it is fixed, screenshot:

Screenshot 2023-10-17 at 18 42 26

Let me know your thoughts.


turbo124:
@Civolilah should the down chevron activate the product list selector? otherwise it would not be necessary to show?

PR #1216, opened 2023-10-18
[Bugfix] Default Tax Rates Parsing To Number

Civolilah:
@beganovich @turbo124 The PR includes fixing the type for the properties tax_rate1, tax_rate2, and tax_rate3 across all settings levels that we send to the payload. We sent them as strings, but they need to be numbers. Let me know your thoughts.


Civolilah:
@turbo124 The PR has been updated with additional fixes from the group/client settings when checking/unchecking checkboxes. Let me know your thought.

turbo124:
@Civolilah 

Tax name is null when the default tax rate is checked but not selected in group settings.

beganovich (approved):

PR #1218, opened 2023-10-18
Dynamic triggering of live previews

beganovich:
This implements dynamic live previews using an intersection observer.


turbo124:
Looks good, are some selectors excluded from this? it seems the tax combo's trigger live preview regardless.

beganovich:
Hm, it shouldn't be triggering live preview no matter what we change, I'll look at the combobox specifically.

PR #1220, opened 2023-10-19
[Cleanup] Data Table Pereferences

Civolilah:
Almost done!

### DIFF
--- a/src/components/DataTable.tsx
+++ b/src/components/DataTable.tsx
@@ -56,6 +56,7 @@ import collect from 'collect.js';
 import { AxiosError } from 'axios';
 import { ValidationBag } from '$app/common/interfaces/validation-bag';
 import { GenericSingleResourceResponse } from '$app/common/interfaces/generic-api-response';
+import { useLocation } from 'react-router-dom';
 
 export type DataTableColumns<T = any> = {
   id: string;
@@ -80,6 +81,15 @@ interface StyleOptions {
   thClassName?: string;
   tdClassName?: string;
   addRowSeparator?: boolean;
+}
+
+interface TablePreference {
+  filter?: string;
+  customFilter?: string[];
+  currentPage?: number;
+  sort?: string;
+  status?: string[];
+  sortedBy?: string;
 }
 
 interface Props<T> extends CommonProps {
@@ -114,10 +124,15 @@ interface Props<T> extends CommonProps {
 
 type ResourceAction<T> = (resource: T) => ReactElement;
 
-export const datatablePerPageAtom = atomWithStorage('perPage', '100');
+export const datatablePerPageAtom = atomWithStorage<string>('perPage', '100');
+
+const dataTableFilterPreferencesAtom = atomWithStorage<
+  Record<string, TablePreference>
+>('dataTablePreferences', {});
 
 export function DataTable<T extends object>(props: Props<T>) {
   const [t] = useTranslation();
+  const location = useLocation();
 
   const [hasVerticalOverflow, setHasVerticalOverflow] =
     useState<boolean>(false);
@@ -125,6 +140,18 @@ export function DataTable<T extends object>(props: Props<T>) {
   const [apiEndpoint, setApiEndpoint] = useState(
     new URL(endpoint(props.endpoint))
   );
+
+  const tableKey = useMemo(() => `${location.pathname}/${props.resource}`, []);
+
+  const [dataTableFilterPreferences, setDataTableFilterPreferences] = useAtom(
+    dataTableFilterPreferencesAtom
+  );
+
+  const getPreference = (filterKey: keyof TablePreference) => {
+    return dataTableFilterPreferences && dataTableFilterPreferences[tableKey]
+      ? dataTableFilterPreferences[tableKey][filterKey]
+      : '';
+  };
 
   const setInvalidationQueryAtom = useSetAtom(invalidationQueryAtom);
   setInvalidationQueryAtom(apiEndpoint.pathname);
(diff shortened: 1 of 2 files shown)


src/components/DataTable.tsx
```diff
@@ -126,6 +141,18 @@ export function DataTable<T extends object>(props: Props<T>) {
     new URL(endpoint(props.endpoint))
   );
 
+  const tableKey = useMemo(() => `${location.pathname}${props.endpoint}`, []);
```
beganovich, line 144: Please be careful with the overuse of useMemo. If used for smaller chunked values, like this one, it can actually _hurt_ performance.

Civolilah: @beganovich Ah, okay. Thank you for providing me with this previously unknown information. I will be careful when using it and will only apply it in appropriate contexts. It has been updated, and we now use plain values.

beganovich (commented):
Looks okay, but we disconnected from the usual preferences setup with react_settings. 

We _usually_ store the preferences as part of the DB, so they're available on multiple devices, not only in the local storage of currently active devices.

Let me know if local storage preferences are what we're going for @turbo124 .

turbo124:
@beganovich @Civolilah. ideally react_settings if it was possible. I raised the idea of local storage in case it was not possible to achieve this.

Civolilah:
> @beganovich @Civolilah. ideally react_settings if it was possible. I raised the idea of local storage in case it was not possible to achieve this.

Civolilah:
@turbo124, @beganovich and I discussed this approach privately on Slack today. It should be fine to proceed with this approach. However, if we decide to go with any other approach, there will be minor changes to switch to it. I can let you know so that you can make the final review of the functionality. Thanks!

turbo124:
One issue I am seeing here is that we are now having double queries.

The first is the baseline query, then the second one which contains the users preferences?

Civolilah:
> One issue I am seeing here is that we are now having double queries.
> 
> The first is the baseline query, then the second one which contains the users preferences?

@turbo124 The user preferences for the tables are auto-saving. When we change the filters on the tables, the first query is related to the data in the table, in order to show updated data IN the TABLE. The second query is related to saving preferences in the `react_settings`. So, these two queries are necessary if we want to save filter preferences in the `react_settings` object. I just tested it one more time, and apart from these two required queries, there are no additional queries. Let me now your thoughts.

turbo124:
@Civolilah 

Please check again.

From any list view. click on a column to sort. Note that there are two GET calls produced now to the list route, instead of one.

Civolilah:
> @Civolilah
> 
> Please check again.
> 
> From any list view. click on a column to sort. Note that there are two GET calls produced now to the list route, instead of one.

@turbo124 Okay, now I see it as well. I didn't notice it before because I was changing filters quickly. The issue is related to caching queries.

The data caching for almost all tables in the app is set to `5000` milliseconds (5 seconds), except for tables where the `staleTime` is set to `Infinity` (`Dashboard` tables). You attempted to sort the next column after waiting for 5 seconds, which resulted in two calls to the /invoices GET method.

Currently, in the PR, I only modified that time. I guess that we'll want to set the staleTime to Infinity, as we do for most other queries.

So, the current staleTime is set to Infinity, and the API won't be called until we manually invalidate the table query.

Please let me know if we indeed want this behavior, but it should resolve the issue of double querying here.

I've tested various scenarios on tables multiple times, but I believe we'll need to thoroughly test this to ensure that we don't introduce any critical bugs.
 
 Let me know your thoughts.

Civolilah:
@turbo124 After Ben merged the PR today, which was related to query invalidation on tables, I eagerly awaited updates on this one. I extensively tested the entire app in terms of changing the tables' staleTime from 5 seconds to `Infinity`. Everything appears to be working perfectly from my perspective. Please let me know your thoughts.

turbo124:
Looks good.

src/components/DataTable.tsx
```diff
@@ -220,7 +338,7 @@ export function DataTable<T extends object>(props: Props<T>) {
     ],
     () => request('GET', apiEndpoint.href),
     {
-      staleTime: props.staleTime || 5000,
+      staleTime: props.staleTime ?? Infinity,
```
beganovich, line 341: What's the reason behind this change?

turbo124: the 5second staletime was not providing a good experience, and with the change to remembering the user preferences for filtering, it was also causing a double query to be performed. It was decided than with the new invalidation being more aggressive, we could push the staletime out to Infinity.

Civolilah: @beganovich @turbo124 Yep! So, in fact, a `staleTime` of 5 seconds does not make sense unless our invalidation works perfectly fine. We actually needed invalidation even when the `staleTime` was set to 5 seconds in order to enable tables to be updated immediately after calling certain actions directly from the table.

We discussed this briefly on Slack. So, basically, this will significantly reduce the number of API calls.

beganovich (changes requested):
While it looks good, I'd advocate for extracting this whole preferences logic into separate hook. This makes datatable really hard to even _navigate_, yet to understand what's actually happening.

Can we get changes related to filters/preferences in useDataTablePreferences hook, so it's clear and concise what are we actually doing?

Also, I see some useMemos, please be _very careful_ with them. Generally, they're not needed if data is organised in correct way.

Civolilah:
@beganovich @turbo124 Confirmation, David: as we discussed on Slack, the `perPage` property is now included in the properties of `table_filters`, resolving all bugs related to the `per_page` property we had earlier. Additionally, Ben, a hook has been created for the recommended separation of logic from the Data Table preferences, so the refactoring has been completed now. Let me know your thoughts.

turbo124:
Looks good!

beganovich (approved):

PR #1234, opened 2023-10-23
[Cleanup] Client Does Not Have Contact Email | Email Invoice Action

Civolilah:
@beganovich @turbo124 The PR includes the implementation of a modal to inform the user that the client does not have any email contacts to email an invoice. The modal also provides an additional CTA to navigate to the client's edit page and add some contacts. Screenshot:

Screenshot 2023-10-23 at 22 27 14

Let me know your thoughts.

### DIFF
--- a/src/pages/invoices/common/components/EmailInvoiceAction.tsx
+++ b/src/pages/invoices/common/components/EmailInvoiceAction.tsx
@@ -22,7 +22,7 @@ import { Modal } from '$app/components/Modal';
 
 interface Props {
   invoice: Invoice;
-  commonActionSection?: boolean;
+  dropdown?: boolean;
 }
 export function EmailInvoiceAction(props: Props) {
   const [t] = useTranslation();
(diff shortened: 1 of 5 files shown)


Civolilah:
@turbo124 Regarding the merging of the PR related to `common_actions` and the new approach of setting it through the API, I needed to make adjustments to this PR as well. It is now complete, and I believe that with this change, every `email_invoice` action should prevent the user from going to the `/email page` and sending an email if the CLIENT does NOT have any contact emails. Otherwise, it will allow user going to the `/email` page. Let me know your thoughts.

turbo124:
If i use the bulk actions, I am able to bypass.

image

Civolilah:
> @Civolilah
> 
> If i use the bulk actions, I am able to bypass.
> 
> image

@turbo124 Ah, sure, the bulk action is also covered right now. Please let me know your thoughts.

beganovich (approved):

PR #1238, opened 2023-10-24
[Cleanup] Selecting Values After Creation Entities

Civolilah:
@beganovich @turbo124 The PR includes fixes for selecting values after creating Tax Rates through the modal and Clients through the slider. It encompasses Expense pages, Recurring Expense pages, and all entities where we can create a Client through the slider. 

The main reason for not selecting the Client properly after creation through the slider is the improper setting of the `with` parameter in the query at the right time. Therefore, this PR includes quick fixes in the Combobox component to address this bug.

Let me know your thoughts.


src/components/forms/Combobox.tsx
```diff
@@ -780,15 +794,20 @@ export function ComboboxAsync<T = any>({
...
     }
-  }, [enableQuery]);
+  }, [enableQuery, inputOptions.value]);
 
   const { data } = useQuery(
     [
-      new URL(url).pathname,
-      new URL(url).searchParams.toString(),
+      new URL($url).pathname,
+      new URL($url).searchParams.toString(),
```
Civolilah, line 810: @beganovich This is exactly what we discussed in today's meeting. We need to include `$url` instead of the default `url` to call the API, otherwise, React Query will not be able to determine when the API needs to be called. I noticed this during this task and made the change.

src/components/forms/Combobox.tsx
```diff
@@ -580,6 +580,8 @@ export function ComboboxStatic<T = any>({
                   if (onDismiss) {
                     e.preventDefault();
 
+                    setQuery('');
```
Civolilah, line 583: @beganovich Clearing the query by setting it to an empty string allows the user to obtain refreshed results without any filters when they click the X button.

src/components/forms/Combobox.tsx
```diff
@@ -790,23 +798,67 @@ export function ComboboxAsync<T = any>({
...
+    setUrl((c) => {
+      const currentUrl = new URL(c);
+
+      if (sortBy) {
+        currentUrl.searchParams.set('sort', sortBy);
+      }
+
+      currentUrl.searchParams.set('status', 'active');
+
+      currentUrl.searchParams.set('filter', '');
```
Civolilah, line 852: @beganovich Not having the `filter` parameter during the initial render results in double queries. We will have one query with the `filter` parameter and one without it. However, the component itself sets the `filter` to an empty string, so to avoid double queries, I set it during the component building.

Civolilah:
@turbo124 I have just finished testing this PR. I conducted extensive testing of the Combobox component to identify and address any potential bugs. It appears that everything works perfectly fine on my end, including bug resolution for this component. The testing included searching by the term, selecting values, unselecting, and creating through modals... The PR is ready for your review. Let me know your thoughts.

src/components/forms/Combobox.tsx
```diff
@@ -758,22 +760,28 @@ export function ComboboxAsync<T = any>({
...
-  const { data } = useQuery(
-    [new URL(url).pathname, new URL(url).href],
-    () => {
-      const $url = new URL(url);
+  const enableQueryTimeOut = useRef<NodeJS.Timeout | undefined>(undefined);
 
-      if (sortBy) {
-        $url.searchParams.set('sort', sortBy);
-      }
+  const [enableQuery, setEnableQuery] = useState<boolean>(false);
```
beganovich, line 765: Can you walk me through this, why do we need timeouts?

Civolilah: @beganovich Let's provide you with an example. For instance, navigate to the Task creation page.

- Select the project.
- Visit the Projects page and remove the project that has already been selected on the Task creation page.
- Return to the Task creation page. If you debug the Combobox behavior, you'll notice that the initial query is executed using the `project_id` selected in the first step. However, after some re-rendering and updating of the task's state, you'll find that the `project_id` is an empty string. This change occurs after the re-render, but the query has already been executed.

Apart from the issue of double querying, you may experience unusual behavior on the screen. It may appear list of projects once with the deleted project and the second time without it. The reason the project appears the first time in the list is because the old `project_id` value is sent as `with` parameter.

There are two possible solutions: One option is to navigate through all pages, and upon unmounting components, reset the value to its default, thus avoiding any unsaved changes. Alternatively, this can be handled directly within the Combobox component.

In essence, this involves waiting for the final value of `inputOptions.value` during the initial mount of the Combobox component for the specific page.

Let me know your thoughts.

Civolilah: @beganovich However, aside from resolving the double querying bug, this also clarifies the behavior of setting the Combobox value. When testing various scenarios, you may notice instances where we experience set, remove, and then set the value in the Combobox again. This occurs due to the double querying and the `inputOptions.value` changing once or twice before obtaining the correct value.

beganovich (approved):

PR #1241, opened 2023-10-25
Delayed exports

beganovich:
This provides us with delayed functionality for exports.

If API returns 409, we will ping the backend with the hash and then present success with the download.

@turbo124 I've just scaffolded the API route as I don't know if we already have this supported on the backend, please let me know if it's correct.


turbo124:
lets reuse

```
reports/preview/{hash}
```

for the endpoint

turbo124:
I think the logic here is broken, on HTTP 200, you should start polling the new route and await the 409's to continue from that route.

beganovich:
@turbo124 it's configured globally, so every report.

Ping for QA pass!

turbo124:
@beganovich 

This looks good. I'll need to update some report endpoints as it looks like we are using this method for all reports.

one thing to note. I think we need to increase the delay between polling to around 2 seconds. and allow up to 50 retries which gives us ~ 100 seconds for the report to generate.

also, after all retries have been exhausted, we should show a failure modal. perhaps if we can move the feedback from a Toast to a Modal with spinner it may be better ux also?

beganovich:
@turbo124 please let me know when API changes are live so I can update the code.

As for the rest, I've set it to 50 retries with a 2s delay. Keep in mind that in development, it is still 1s.

Also, this differs now from the reports themselves, as they use 10 retries with 5s delays. Let me know if they should be adjusted too. Also, I am not sure about the modal, but I am definitely open to changes. What should be the content of the modal and how do we benefit from just spinning the "Processing" alert?

turbo124:
Should we disable the export button during polling? otherwise this may create duplicate requests?

In regards to modal, we can keep it like this for now, and iterate on feedback

beganovich:
@turbo124 PR is updated with disabled button now.

turbo124:
Looks good, just waiting to sync deployment with API

PR #1248, opened 2023-10-27
Changing expense category in bulk

beganovich:
This allows us to select multiple expenses and change their category in bulk.


beganovich:
Sorry, I missed modal closing, it should work fine now @turbo124.

turbo124:
Thanks,

Where possible I think we should utilize the new combo box where the user can enter searchable and/or select. If it isn't much of a change it would be good to implement this.

PR #1260, opened 2023-10-30
Change logic for auto logout

beganovich:
This changes logout logic to clear the storage only if 429 and 403.


turbo124:
no, this looks fine.

PR #1265, opened 2023-10-30
[Bugfix] Invoice Label Property On Payment Page

Civolilah:
@beganovich @turbo124 Ben had already fixed this issue on this page by improving the labels. Instead of resolving this bug, I made the change that caused us to have the Invoice `number` as the label on the combobox.

The property for the label was set to `name` instead of `number`. Consequently, data was cached using the `name` property for invoices, which does not exist, and this was the initial reason for the bug.

This change will not affect the UI, but I believe it's important to cache the correct data, which is now the case.

Let me know your thoughts.


src/components/forms/Combobox.tsx
```diff
@@ -818,7 +804,16 @@ export function ComboboxAsync<T = any>({
 
   useEffect(() => {
     if (data) {
-      setEntries([...data]);
+      setEntries(
```
Civolilah, line 807: @beganovich The data was adjusted using `entryOptions` only if we had a case triggering the API. The `than` block will never be executed if we retrieve the data from cache memory. This is the reason for the change and the core of the problem why the label was not populated with the initial number property. So, the data needs to be just the response. However, if we need some adjustments, we can do that during the initial mounting of the combobox, even if the data is coming from the cache. Let me know your thoughts.

turbo124:
@beganovich leaving this one for you to QA

beganovich (approved):

PR #1266, opened 2023-10-30
Fixes for users table

beganovich:
This fixes status filtering for the user table. By default, it runs `status=active`. This fixes an issue with not showing deleted records.


beganovich:
@turbo124 this removed `hideRemovedUsers`, please let me know if this is _required_.

Civolilah:
@beganovich @turbo124 I'm noticing two issues with this one. After I removed the user from the user's edit page and returned to the table's page, the list is not updated. I updated the invalidation on the edit page, and it started working fine. However, the `status=active` parameter still doesn't give me the same results as `hideRemovedUsers=true`. The `status=active` parameter always returns all users.

Civolilah:
@turbo124 When I create a user and then attempt to create a second one in a row, I notice that just the details for the first user are changed. I assume this behavior is not what we want. Video:

https://github.com/invoiceninja/ui/assets/51542191/d7cde075-75e8-4557-9221-a988ee8ec810

turbo124:
@Civolilah thanks, i think these issues are in the API, i've checked a fix in.

@beganovich my testing using updated API looks good, we'll need to wait to merge until we have the API changes ready in production.

PR #1267, opened 2023-10-30
Confirmation for recurring invoices send action

beganovich:
This implements confirmation for the send now action when creating the recurring invoices.


turbo124:
I'm not seeing the modal here?

beganovich:
Ping @turbo124 for a QA pass. Previously was implemented only on create page. Now it supports edit, too.

PR #1281, opened 2023-11-01
[Bugfix] Ensuring 400 erros are shown in toaster

Civolilah:
@beganovich @turbo124 The PR includes fixes for displaying a 401 error message in the toaster. To ensure that the entire logic is correct, here it is:

- If a user receives a 429 OR 403 error, they will be logged out of the app.
- If a user receives a 404 error, they will be navigated to the "not_found" page.
- If a user receives a 412 error, it will be displayed as a "wrong password" message in the toaster. It is catched throughout the entire app.
- If a user receives a 422 error, it will be displayed as a validation message under the field.
- If a user receives any error between 400 (including 400) and 500, the message will be displayed in the toaster.
- If a user receives a 500 error, we will display a "Something went wrong" message in the toaster.

Let me know your thoughts.

### DIFF
--- a/src/common/helpers/request.ts
+++ b/src/common/helpers/request.ts
@@ -31,10 +31,9 @@ client.interceptors.response.use(
 
     if (
       error.response?.status &&
-      error.response.status !== 401 &&
       error.response.status !== 412 &&
       error.response.status !== 422 &&
-      error.response.status !== 400 && // Temporary solution. If you see this in few months, please delete (live previews).
+      !error.request.url.includes('live_preview') && // Temporary solution. If you see this in few months, please delete (live previews).
       error.response.status > 399 &&
       error.response.status < 500
     ) {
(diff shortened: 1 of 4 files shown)


src/common/helpers/request.ts
```diff
@@ -31,10 +31,9 @@ client.interceptors.response.use(
 
     if (
       error.response?.status &&
-      error.response.status !== 401 &&
       error.response.status !== 412 &&
       error.response.status !== 422 &&
-      error.response.status !== 400 && // Temporary solution. If you see this in few months, please delete (live previews).
```
Civolilah, line 37: @beganovich @turbo124 I know that we had some issues with a 400 error on live_previews, so this was a temporary solution. Have we resolved it now, so I can remove the unnecessary checks?

turbo124: @beganovich it should be ok to remove line 37 now if we have toast, correct?

beganovich: I think so, yep 👍️

beganovich (approved):

src/common/helpers/request.ts
Civolilah: @beganovich Okay, I just removed the checks for live previews.

PR #1283, opened 2023-11-02
[Bugfix] Disabling Loading Google OAuth Scripts For Self-hosted

Civolilah:
@beganovich @turbo124 The client reported unnecessary loading of the `https://accounts.google.com/gsi/client` script in the HTML body. To prevent the loading of this script, I had to remove the `<GoogleOAuthProvider clientId={googleClientId}>` wrapper for SELF-HOSTED users.

Additionally, we have a `Connected Account` page in the User Details settings section where we also have the `Google Login button` for connecting accounts. This button is available for both types of users. However, when I removed this wrapper and navigated to this page, it broke the entire app because we did not have the necessary wrapper and required scripts. Therefore, I had to remove the Google Login button from the `Connect Accounts` page to prevent the app from breaking.

So, are SELF-HOSTED users able to connect accounts? If that's the case, I don't believe it's possible to have a combination of not having a wrapper and not injecting the required scripts while still having the Connecting Google Account feature at the same time.

But if they are unable to connect accounts on the self-hosted platform, perhaps we can consider removing the OAuth/Mail tab altogether?

Important note: I cannot see any scripts that are injected from Microsoft's side.

Let me know your thoughts.


turbo124:
@beganovich cannot QA, if in doubt, we can test on beta

Civolilah:
> @beganovich cannot QA, if in doubt, we can test on beta

@turbo124 Actually, I believe it is possible to test. Try opening the browser's inspector and search for the `https://accounts.google.com/gsi/client` script in the body. For a self-hosted user, you will not be able to find it. However, if you simply set `VITE_IS_HOSTED` to `true` in the .env file, you will be able to find this script, or let's say the wrapper will automatically inject it. I think that is the whole point. Let me know your thoughts.

turbo124:
@Civolilah good point. looks good!

beganovich (approved):

PR #1286, opened 2023-11-03
[Bugfix] Adjusting Actions And Buttons Per User's Permissions

Civolilah:
In progress...


turbo124:
image
Looks good, one issue. with View and Edit permissions for the client only - I should not be able to see any create options.

PR #1293, opened 2023-11-07
[Bugfix] Caching Fixes For Invoice && Credit Resolver On Payment Creation Page

Civolilah:
@beganovich @turbo124 The PR contains fixes for caching in the Invoice and Credit resolvers on the Payment creation page. Previously, we used an outdated approach for handling these two hooks with classes. Now, they have been updated, and along with these updates, bugs have been fixed. Essentially, if you modify the `balance` on an Invoice and then attempt to enter a Payment with the updated `balance`, the change won't be reflected on the Payment creation page. The same issue occurs with Credits. These two hooks are exclusively utilized on the Payment creation page, so the bug is isolated to this specific location. Quick demo that explains this behavior:

https://github.com/invoiceninja/ui/assets/51542191/3350f68f-5060-4829-84d4-a85f28a7d077

Let me know your thoughts.


src/common/hooks/credits/useCreditResolver.ts
```diff
@@ -8,10 +8,25 @@
...
+import { request } from '$app/common/helpers/request';
+import { Invoice } from '$app/common/interfaces/invoice';
+import { useQueryClient } from 'react-query';
 
 export function useCreditResolver() {
-  return creditResolver;
+  const queryClient = useQueryClient();
+
+  const find = (id: string) => {
+    return queryClient.fetchQuery<Invoice>(
```
beganovich, line 20: I assume this is supposed to be `Credit` type?

beganovich (approved):
Looks good, just to fix the interface reference.

beganovich (approved):

PR #1295, opened 2023-11-07
Add `uninvoiced` to tasks table

beganovich:
This adds `uninvoiced` status to the tasks table.


turbo124:
Works well, to see it in action, you need to remove the All filter...

PR #1329, opened 2023-11-20
[Bugfix] Add Tasks To Invoice Bulk Action | Tasks

Civolilah:
@beganovich @turbo124 The PR addresses the bug with the "Add Tasks To Invoice Bulk Action." The bug was related to setting selected resources as an empty string at the wrong time, which is why we did not have any new task line items on the invoice page. Let me know your thoughts.


turbo124:
This look good, thankyou!

beganovich (approved):

PR #1332, opened 2023-11-21
[Cleanup] Recurring Invoices, Payments And Quotes Permissions Adjustment && Additional Tests

Civolilah:
@beganovich @turbo124 The PR includes adjustments to permissions for Recurring Invoices, Payments, and Quotes, along with minor fixes for document tables related to previously completed entities. Let me know your thoughts.


turbo124:
Looks good!

Civolilah:
@beganovich Quick confirmation, this PR is also ready for your final review.

src/pages/payments/edit/hooks/useTabs.ts
```diff
@@ -20,31 +25,59 @@ interface Params {
...
-      href: route('/payments/:id/payment_fields', { id }),
-    },
+    ...(canOnlyEdit
+      ? [
+          {
+            name: t('apply'),
+            href: route('/payments/:id/apply', { id }),
+          },
+        ]
+      : []),
```
beganovich, line 56: Honestly, this looks super messy. Is it possible to abstract this or integrate logic directly in the tabs component?

src/pages/recurring-invoices/common/components/RecurringInvoiceSlider.tsx
```diff
@@ -273,7 +282,13 @@ export const RecurringInvoiceSlider = () => {
...
-                    <Link to={`/clients/${activity.client_id}`}>
+                    <Link
+                      to={`/clients/${activity.client_id}`}
+                      disableNavigation={disableNavigation(
+                        'client',
+                        recurringInvoice?.client
+                      )}
+                    >
                       {recurringInvoice?.client?.display_name}
                     </Link>
```
beganovich, line 293: Should we maybe conditionally render the link and p tag? This way people will think the app's broken by the link looking clickable and doing nothing.

src/pages/recurring-invoices/routes.tsx
```diff
@@ -82,7 +82,10 @@ export const recurringInvoiceRoutes = (
           <Guard
             guards={[
               enabled(ModuleBitmask.RecurringInvoices),
-              permission('view_recurring_invoice'),
+              or(
+                permission('edit_recurring_invoice'),
+                assigned('/api/v1/recurring_invoices/:id')
+              ),
```
beganovich, line 88: We swapped edit for view here?

beganovich (commented):
Looks good, few thing might need a change.

src/pages/payments/edit/hooks/useTabs.ts
Civolilah: @beganovich Yeah, you are right. I just formatted it to make the code look better. Let me know your thoughts.

src/pages/recurring-invoices/routes.tsx
Civolilah: @beganovich Actually, all actions, including those that navigate to the PDF page, are removed if the user does not have edit permission, did not create the entity, or is not an admin or owner. 

Regarding the approach, I think if the user only has view permission, they can look at the PDF preview at the bottom of the page and see the same content as on the PDF page. However, if, for example, the PDF pages contain some actions, we will need to additionally hide them and implement more logic. This way, we simply remove all actions if the user does not have at least edit permission.

So, yeah, that is the main reason for this change: removing access to the page if the user does not have access to the actual action.

src/pages/recurring-invoices/common/components/RecurringInvoiceSlider.tsx
Civolilah: @beganovich Yeah, I agree with you. So, we only disable navigation for the `Link` component. I have, of course, conditionally rendered the `<span>` tag instead of the `RouterLink` component when the property `disableNavigation` is true.

src/pages/payments/edit/hooks/useTabs.ts
beganovich: Honestly, I thought you were going to implement `enabled` property on tab, seems like reasonable thing to do?

src/pages/recurring-invoices/common/components/RecurringInvoiceSlider.tsx
beganovich: I think you missed a point here. `something ? <Link> : <span>`. No need to change base Link component.

Civolilah: @beganovich As we discussed on Slack, perhaps the best option is to have a component that will render the Link or Span instead of having direct logic in every place. Therefore, the implementation has now been changed.

Civolilah:
@beganovich I cannot find the second comment to reply to. So, a quick confirmation: it is ready for final review, and all needed changes have been made.

beganovich (approved):

PR #1336, opened 2023-11-22
Fixes for typo, causing invalidation issue

turbo124:


turbo124:
close in favor of #1334

PR #1364, opened 2023-11-29
[Cleanup] Adding Missing Events For API Webhooks

Civolilah:
@beganovich @turbo124 The PR includes adding missing events for API webhooks to align with Flutter. Let me know your thoughts.


turbo124:
when i save a sent invoice webhook, it returns with create client.

in the create webhook screen i see all webhooks, but in the edit, i don't see them all?

Civolilah:
> when i save a sent invoice webhook, it returns with create client.
> 
> in the create webhook screen i see all webhooks, but in the edit, i don't see them all?

@turbo124 Ah, okay, the issue is fixed now. It was the same reason for both bugs you mentioned. Let me know if you observe the correct behavior as I do.

PR #1366, opened 2023-11-29
added button to download E-Invoice

stevro:
I've added the option to download the E-Invoice from the admin portal.
In Romania the e-invoice must be submitted to the local authorities by the issuer/supplier, and until now the only option to get the e-invoice(xml) was either from the customer email or from the customer portal, both options not being handy for the admin.


stevro:
Wrong upstream branch for PR. Redoing it to develop branch.

PR #1367, opened 2023-11-29
[Cleanup] Storing The show_mini_sidebar In React Settings

Civolilah:
@beganovich @turbo124 The PR includes some UI fixes for the mobile navbar and improvements for storing the `show_mini_sidebar` property in the `react_settings`. Let me know your thoughts.


turbo124:
:+1:

src/components/HelpSidebarIcons.tsx
```diff
@@ -88,7 +99,21 @@ export function HelpSidebarIcons(props: Props) {
...
+    const showMiniSidebar =
+      user?.company_user?.react_settings?.show_mini_sidebar;
+
+    if (
+      user &&
+      currentUser &&
+      Object.keys(user).length &&
+      typeof showMiniSidebar !== 'undefined' &&
+      currentUser.company_user?.react_settings?.show_mini_sidebar !==
+        showMiniSidebar
```
beganovich, line 112: Please riddle me this part of the code and why can't we just use last part which is important?

beganovich (changes requested):
Generally looks good, just need a clarification.

src/components/HelpSidebarIcons.tsx
Civolilah: @beganovich Okay, yeah, I found two unnecessary parts of the if statement: the check for `currentUser` and the check for the `length of keys for the user object`. I discovered that the second check for the length of user object keys is included in this part of the statement typeof `showMiniSidebar !== 'undefined'`. However, we still need to check for the user object, and the reason for that stems from the `useInjectUserChanges` hook. The actual specification of the return value here is `User | undefined` type. So, while the user object can never be undefined, it can only be the `empty object`. Changing this would be too expensive for this PR to modify the return type under the `useInjectUserChanges` hook. Therefore, I just made the check for the type of the user object to ensure it is not undefined, allowing it to pass through TypeScript rules.

beganovich (approved):

PR #1378, opened 2023-12-02
[Bugfix] Resolving Client/Vendor

Civolilah:
@beganovich @turbo124 The PR addresses the issue related to resolving the wrong client while creating it from the Client show page. The main reason for this issue is not incorrect caching data but rather a flaw in the logic. The bug is not limited to Invoices, as David observed during testing, it occurs with all entities where we use the same method of selecting a client (Quotes, Recurring Invoices, Credits, and so on). Let me know your thoughts.


src/pages/invoices/common/components/ClientSelector.tsx
```diff
@@ -51,10 +53,19 @@ export function ClientSelector(props: Props) {
...
+      const currentTimeout = setTimeout(
+        () =>
+          clientResolver
+            .find(resource.client_id)
+            .then((client) => setClient(client)),
+        25
+      );
+
+      resolveClientTimeOut.current = currentTimeout;
+    }
```
Civolilah, line 68: @beganovich The reason for the bug is that we do not set entities to an `undefined` value upon unmounting `Creation` components. Consequently, `resource.client_id` remains truthy even on the initial render when we have the `"incorrect"` client_id compared to the `passed` client_id that should be utilized. However, the logic cannot discern which one is correct, leading it to call the API on the first render and set the client regardless of subsequent renders.

There are a few ways to resolve this. The first and most straightforward approach is to set the entities to an undefined value when we `unmount` creation components. For instance, when the creation component for Invoices unmounts, we can simply execute `setInvoice(undefined)`.

I opted against this approach due to the current test coverage. Uncertain of potential issues that this change might introduce, especially given the significance of the entities involved, I chose a more secure alternative.

In this alternative approach, we await the last change of `resource.client_id` and then call the API to resolve the client.

Once we achieve extensive test coverage, we can reconsider and potentially adopt a different, perhaps more preferable, approach to address similar bugs.

I hope you agree with this chosen approach.

turbo124:
@Civolilah looks good!.

@beganovich for patch to main please.

beganovich (commented):
I'd suggest using QueryClient with enabled property on itself. This way you can actually control better. I've refactored the client to query client recently and it worked out great.

This and a combination of useEffect clean up fn should do the job better than timeout, if I get this correctly.

Civolilah:
> I'd suggest using QueryClient with enabled property on itself. This way you can actually control better. I've refactored the client to query client recently and it worked out great.
> 
> This and a combination of useEffect clean up fn should do the job better than timeout, if I get this correctly.

@beganovich Sure, as we discussed on Slack, and with your suggestion, I have changed the approach to the solution, incorporating cleanup functions during the creation of components.

src/pages/purchase-orders/create/Create.tsx
```diff
@@ -112,6 +112,10 @@ export default function Create() {
 
       return value;
     });
+
+    return () => {
+      setPurchaseOrder(undefined);
+    };
```
Civolilah, line 118: @beganovich The same issue occurs with vendor selection when creating a purchase order, and that is the reason for implementing this cleanup.

beganovich (approved):

PR #1399, opened 2023-12-07
[Bugfix] Displaying NaN Values

Civolilah:
@beganovich @turbo124 The PR includes fixes for displaying `NaN` values when formatting money. Once we receive a `NaN` value, we will show `0` instead. Screenshot:

Screenshot 2023-12-07 at 19 03 18

Let me know your thoughts.


Civolilah:
> @Civolilah
> 
> Whilst this works, is the base reason for this the input types are set as text? if we can confirm there are no additional edge cases, i think it would be best to apply the correct input type?

@turbo124 Hmm, the root cause of the `NaN` values being displayed is that we have a 'wrong' calculation under the `InvoiceSum` and `InvoiceSumInclusive` classes. So, essentially, the logic under those classes results in this. See screenshot:

Screenshot 2023-12-08 at 20 08 08

However, I believe that there are just cases under those classes that are not covered. Nevertheless, I am quite confident that those classes perform calculations very well. I haven't delved too much into the logic of those classes, but I can attempt to fix it with extensive testing afterward because it is crucial for it to function properly. To ensure a more secure solution, I have implemented `isNaN` checking and display `0` instead. However, the reason for displaying `NaN` as text is that we are rendering it in the `<span>` HTML tag.

So, let me know, do you think a better solution would be this, or should we try to identify the core problem under those classes that calculate these values and avoid this edge case?

turbo124:
@Civolilah you are correct, i'll recheck the sum classes and ensure that we don't have nan's

src/common/hooks/money/useFormatMoney.ts
```diff
@@ -30,7 +31,7 @@ export function useFormatMoney() {
     const currency = resolveCurrency(currentCurrencyId);
 
     if (country && currency) {
-      return Number.formatMoney(value, currency, country, {
+      return Number.formatMoney(isNaN(value) ? 0 : value, currency, country, {
```
beganovich, line 34: Is there a specific reason for not using JS built-in isNaN?

Civolilah: @beganovich Actually, there is no big reason to do that, except that we do not need to cast to a number when passing it to the `isNaN` function from the `Lodash` package, we can pass string also. But yeah, I switched it to the `built-in` isNaN function to check this state. Let me know your thoughts.

beganovich (approved):

PR #1417, opened 2023-12-13
[Cleanup] Remove Logo CTA Implementation On All Edit Pages With PDF Preview

Civolilah:
@beganovich @turbo124 The PR includes the implementation of the Remove Logo CTA for all edit pages with PDF preview, which encompasses `Invoices`, `Recurring Invoices`, `Quotes`, `Credits`, and `Purchase Orders`. Screenshot:

Screenshot 2023-12-13 at 17 57 00

@turbo124 Currently, this is shown only for `HOSTED` users with a `FREE PLAN`. Is it supposed to be like that?

Let me know your thoughts.

### DIFF
--- a/src/components/RemoveLogoCTA.tsx
+++ b/src/components/RemoveLogoCTA.tsx
@@ -15,14 +15,17 @@ import { proPlan } from '$app/common/guards/guards/pro-plan';
 import { enterprisePlan } from '$app/common/guards/guards/enterprise-plan';
 import { route } from '$app/common/helpers/route';
 import { useCurrentUser } from '$app/common/hooks/useCurrentUser';
+import { useAdmin } from '$app/common/hooks/permissions/useHasPermission';
 
 export function RemoveLogoCTA() {
   const [t] = useTranslation();
   const user = useCurrentUser();
 
+  const { isOwner } = useAdmin();
+
   return (
     <>
-      {!proPlan() && !enterprisePlan() ? (
+      {!proPlan() && !enterprisePlan() && isOwner ? (
         <div className="flex text-base space-x-1">
           <Link
             className="capitalize"
--- a/src/components/RemoveLogoCTA.tsx
+++ b/src/components/RemoveLogoCTA.tsx
@@ -1,0 +1,51 @@
+/**
+ * Invoice Ninja (https://invoiceninja.com).
+ *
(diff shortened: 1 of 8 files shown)


turbo124:
@Civolilah in the case where this is an enterprise account, only the is_owner user should see this CTA.

Civolilah:
> @Civolilah in the case where this is an enterprise account, only the is_owner user should see this CTA.

@turbo124 Ah, okay. I've made the change. If the user is `HOSTED` and has the `FREE PLAN`, and if they are the `OWNER` of the account, in that case, the CTA will be shown, otherwise, it will not. So, if someone adds a new user from the `enterprise` account to the `sub-account`, the user will NOT see the CTA. I did not have any enterprise account on the HOSTED platform to test this.

The condition is very straightforward, but if you have an enterprise account, please double-check that everything works perfectly fine.

src/components/RemoveLogoCTA.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { useCurrentUser } from '$app/common/hooks/useCurrentUser';
+import { useAdmin } from '$app/common/hooks/permissions/useHasPermission';
+
+export function RemoveLogoCTA() {
+  const [t] = useTranslation();
+  const user = useCurrentUser();
+
+  const { isOwner } = useAdmin();
+
+  return (
```
beganovich, line 26: You can get way cleaner implementation by early return and escape ? and {} hell.

```tsx
if (!proPlan() && !enterprisePlan() && isOwner) {
...
}

return null;

File: src/pages/credits/edit/Edit.tsx [hunk 2]
```diff
@@ -181,6 +182,8 @@ export default function Edit() {
               endpoint="/api/v1/live_preview?entity=:entity"
             />
           )}
+
+          <RemoveLogoCTA />
```
beganovich, line 186: Is there a reason we didn't bake this directly into the InvoicePreview component, or the component one-step lower?

beganovich (changes requested):

src/components/RemoveLogoCTA.tsx
Civolilah: @beganovich Sure thing! I refactored it, it looks cleaner now!

beganovich (approved):

PR #1419, opened 2023-12-14
[Feature] Preventing Closing Tab/Browser And Leaving Page Without Confirmation Modal For Invoices

Civolilah:
@beganovich @turbo124 The PR includes the implementation of preventing users from leaving the Creation and Edition pages for Invoices if changes have been made to the corresponding Invoice for which the page is open. It `INVOLVES` preventing users from leaving the page through the app buttons/links (navigating to another app page), preventing the closure of the tab or the entire browser, preventing navigation back in the history using the back browser button, preventing direct entry of URLs in the browser, and preventing page reloading. Screenshot:

Screenshot 2023-12-18 at 02 08 38

Let me know your thoughts.


src/App.tsx
```diff
@@ -33,7 +33,28 @@ import {
...
+
+export const preventLeavingPageAtom = atom<PreventLeavingPage>({
+  prevent: false,
+  actionKey: undefined,
+});
+
+export const lastHistoryLocationAtom = atom<HistoryLocation>({
+  lastLocation: '',
+  nonPreventedLocations: [],
+});
```
beganovich, line 57: app.tsx is getting quite chunky, please move this into a separate module or usePreventNavigation, that works, too.

src/App.tsx
```diff
@@ -33,7 +33,28 @@ import {
 import { colorSchemeAtom } from './common/colors';
 import { useCurrentUser } from './common/hooks/useCurrentUser';
 import { useRefetch } from './common/hooks/useRefetch';
+import { PreventNavigationModal } from './components/PreventNavigationModal';
+import { isNavigationModalVisibleAtom } from './common/hooks/usePreventNavigation';
 
+interface PreventLeavingPage {
+  prevent: boolean;
+  actionKey?: 'switchCompany' | 'browserBack';
```
beganovich, line 41: What exactly is actionKey and what it is used for?

src/components/PreventNavigationModal.tsx
```diff
@@ -0,0 +1,127 @@
...
+            window.open(url, '_blank');
+          } else {
+            navigate(url);
+          }
+        }
+      }
+
+      fn?.();
+    }
+  };
```
beganovich, line 87: This looks fairly complex. Any chance we can provide some tests for this one? We definitely don't want navigation just randomly failing, that'd be bad.

beganovich (commented):
While it works, it kinda feels really complex. Can you walk us through the implementation?

Also is it possible to provide test suite for such critical part of the app.

src/App.tsx
Civolilah: @beganovich We have two cases when the `actionKey` is needed. The `actionKey` is a property that defines what is currently triggered in our case—either the switch company functionality or triggering the back button on the browser.

So, why is it important?!

The functionality of switching companies is based on the browser `reloading`, which immediately means that if needed at that moment, the `beforeunload` event will be triggered, and the browser will show the default modal for preventing leaving the page. This modal provides 'Cancel' and 'Leave' buttons, but in the case of switching companies, 'Cancel' will work just as we need. However, the 'Leave' button will not know how to switch the company (it will not remember the 'switchCompany' function and execute it as needed). Therefore, we need to display our own modal and also prevent the default one from being displayed. Essentially, the 'switchCompany' actionKey is only used when it is necessary to prevent the display of the default browser's modal.

The second 'browserBack' action key is used when it is necessary to push a new state into the history. I conducted thorough research, and it is impossible to disable the back button of the browser unless we push the current page into the state. This way, when the user clicks the back button, they will stay on that page. So, this actionKey is used when we need to do that, avoiding unnecessary pushing of a new state into the history. Additionally, there is another case when it is also very important. If the user triggers the back button and then triggers the 'continue editing' button on the prevent navigation modal multiple times in a row (potentially), they will have, let's say, 10+ instances of the current page in the history object. Therefore, when the prevention of navigation is no longer needed, the user would have to click the back button 10+ times in the browser to go back to the previously non-prevented page. To avoid this poor UX, I store the last page once the user is not prevented from navigation, and then I navigate the user to that page instead of navigating to the exact same page where they are at that moment.

Please let me know if you need any additional explanation for this part of the logic.

src/App.tsx
Civolilah: @beganovich Yeah, I separated it into the hook called `useAddPreventNavigationEvents`.

src/components/PreventNavigationModal.tsx
Civolilah: @beganovich Sure, explaining the entire logic in this comment: https://github.com/invoiceninja/ui/pull/1419#pullrequestreview-1790567160

Civolilah:
> While it works, it kinda feels really complex. Can you walk us through the implementation?
> 
> Also is it possible to provide test suite for such critical part of the app.

@beganovich I needed to await the merging of some important PRs to resolve conflicts on this one and create tests. The implementation might appear complex, but I believe it can be explained, and it is carried out in few steps.

When a user enters the Invoice pages, they have a few options to leave the page: close the browser, close the tab, use the back browser button, navigate using the left main navigation, use the Quick Creation Popover in the navigation, use the back button in the navigation, or use actions on the Invoice entity.

To prevent the browser/tab from closing, we need to add a `beforeunload` event in the app. This event will track these actions, and under the function called `handleBeforeUnload`, there is a small logic defining when the default browser's modal should be displayed to prevent the user from leaving the page.

To prevent the user from leaving the page using the back browser's button, we need to use the `popstate` event. This event is triggered when the user clicks that button, but we are unable to disable the button or prevent navigation. Instead, we need to push a new state into the `history` object to prevent navigation. I explained much more about this in the comment you asked for regarding the `actionKey` property.

All other preventions are handled using the main `usePreventNavigation` hook. If prevention is necessary, the hook will know it using the `preventLeavingPageAtom` atom. The state of this atom changes while the user changes the `invoice` object, allowing us to compare the first version when the user joined the page and the version while they have been making changes.

Once the `preventNavigation` function is triggered, we store the necessary details such as the URL, external URL (bool), actionKey, or probably the function because some actions are not only based on navigation, we need to execute some part of the logic.

If the user clicks "continue editing," nothing will change except the modal, which will be closed. If the user clicks "discard changes," then the logic will be executed under the `PreventNavigationModal` based on the stored details.

Of course, there is also a case when we do not need prevention. In this case, the `preventNavigation` function will execute the same action as needed by default.

Literally, everything is implemented in the main components—`DropdownElement`, `Link`. There are only, I believe, 2-3 very rare cases when those components are not sufficient.

Of course, I encountered some cases where the components are used directly from the package, not our custom components, such as the `Link` component in the `Breadcrumbs`. There, I needed to make a change to use the default component so that the entire logic can be globalized. Cases like this one are probably the main reason why the PR is a bit larger and seems complex.

The entire logic is globalized, so the implementation of this functionality for any other entity is a minor part of the job. In the logic of Creation and Edition components, only a part of the code that tracks whether navigation should be prevented needs to be added.

Also, for the main parts of the functionality, such as the navigation bar, main left navigation, or invoice actions, tests have been implemented.

Let me know your thoughts.

turbo124:
@beganovich we'll need to look into this and deploy a solution.

turbo124:
on review, it appears only certain fields are intercepted.

the line items and any data within the text area's are not triggering the modal.

Civolilah:
> on review, it appears only certain fields are intercepted.
> 
> the line items and any data within the text area's are not triggering the modal.

@turbo124 Yeah, I can recreate it, it only happens with text area input fields. I will fix it.

beganovich:
@Civolilah please mark this as QA pending and ready for review once it's ready.

Civolilah:
@beganovich @turbo124 The PR now includes the fixes we discussed, along with fixes to the tests so we can run them locally at any time to check for any issues. The PR is ready for review. Let me know your thoughts.

turbo124:
tinymce input is not being triggered for this action. also need to ensure any inputs in text area in the line items are also captured.

Civolilah:
> tinymce input is not being triggered for this action. also need to ensure any inputs in text area in the line items are also captured.

@turbo124 Ah, great catch! I was able to recreate it. The issue was with the reference. Now, the issue should be fixed, and I cannot recreate it anymore. It should work for any input field, including the TinyMCE markdowns and any text area input fields as you mentioned. Let me know if you also see it fixed.

src/pages/invoices/create/hooks/useInvoiceUtilities.ts
```diff
@@ -69,8 +69,8 @@ export function useInvoiceUtilities(props: Props) {
 
     if (currency && invoice) {
       const invoiceSum = invoice.uses_inclusive_taxes
-        ? new InvoiceSumInclusive(invoice, currency).build()
-        : new InvoiceSum(invoice, currency).build();
+        ? new InvoiceSumInclusive({ ...invoice }, currency).build()
+        : new InvoiceSum({ ...invoice }, currency).build();
```
beganovich, line 73: I wonder if this is still required? Isn't the spread operator part of previous implementation?

Civolilah: @beganovich Yes, you are right, this is unnecessary until we improve the implementation. So, I've removed the spreading. Along with this removal, I've fixed conflicts related to merging other PRs. Let me know your thoughts.

beganovich (approved):

PR #1421, opened 2023-12-16
[Cleanup] Tooltip On Invoice Number Payment Column

Civolilah:
@beganovich @turbo124 The PR includes the implementation of Ben's recommendation to truncate the `invoice_number` in the Payment column when there are many invoices applied to a specific payment. Screenshot:

Screenshot 2023-12-16 at 20 42 01

Let me know your thoughts.

### DIFF
--- a/src/components/DynamicLink.tsx
+++ b/src/components/DynamicLink.tsx
@@ -25,7 +25,7 @@ export function DynamicLink(props: Props) {
   const { renderSpan } = props;
 
   if (renderSpan) {
-    return <span>{props.children}</span>;
+    return <span className={props.className}>{props.children}</span>;
   }
 
   return (
--- a/src/components/Tooltip.tsx
+++ b/src/components/Tooltip.tsx
@@ -9,12 +9,13 @@
  */
 
 import classNames from 'classnames';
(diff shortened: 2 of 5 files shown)


turbo124:
@Civolilah do we have control over the length before truncate takes over? i think we should show a max of 3-4, and perhaps send the rest into tooltip?

Civolilah:
> @Civolilah do we have control over the length before truncate takes over? i think we should show a max of 3-4, and perhaps send the rest into tooltip?

@turbo124 Great! I didn't remember the solution involving a `Tooltip`, but now I've implemented it. Screenshot:

Screenshot 2023-12-17 at 03 00 35

I did a few more things related to the tooltip here:

- I noticed a bug with the tooltip on the Clients table and Invoice slider, so I fixed it. Screenshot:

Screenshot 2023-12-17 at 03 17 04

- The tooltip component was not adjusted for `dark mode`, so I just did that now (you probably already noticed the changed background color from the first screenshot).

All changes are very small, so I don't think it will be too much for this small PR.

Let me know your thoughts.

beganovich (approved):

PR #1426, opened 2023-12-19
[Cleanup] Connect Account Flow Improvement | Modal Implementation

Civolilah:
@beganovich @turbo124 The PR includes an improvement in the flow when the user connects to the email provider. Once Gmail or Office 365 is connected, the user will receive a modal for selecting the provider in the email_settings without having to navigate to the actual page and perform the action manually. In addition to the provider, the `gmail_sending_user_id` property will also be updated to the value of the current user's ID. Screenshot:

Screenshot 2024-01-23 at 18 48 30

Instead of using plain text in the screenshot, I employed the translation keyword `select_email_provider`. I envisioned the translation to include a variable at the beginning, such as "`:provider` now connected, would you like to select this account to send from?" This way, we can implement logic similar to the following:

```
{trans('select_email_provider', {
          provider: user?.oauth_provider_id === 'google' ? 'Gmail' : 'Outlook',
 })}
```
 
When the provider is Google, the `:provider` variable will be set to Gmail, otherwise it will be set to Outlook.

 Let me know your thoughts.


beganovich:
The code looks good, but I'd suggest dropping "No" unless that was a task requirement.

Yes & X on the top right should be enough.

Civolilah:
> The code looks good, but I'd suggest dropping "No" unless that was a task requirement.
> 
> Yes & X on the top right should be enough.

@beganovich @turbo124 Sure, I made the change, and this is what it looks like after the modification. Screenshot:

Screenshot 2024-01-25 at 07 43 59

Let me know your thoughts.

Civolilah:
> The code looks good, but I'd suggest dropping "No" unless that was a task requirement.
> 
> Yes & X on the top right should be enough.

@beganovich @turbo124 Sure, I made the change, and this is what it looks like after the modification. Screenshot:

Screenshot 2024-01-25 at 07 52 23

Let me know your thoughts.

beganovich (approved):

PR #1430, opened 2023-12-20
[Bugfix] Exclusion Of Owner User From Users List

Civolilah:
@beganovich @turbo124 The PR includes the implementation of excluding the account owner from the list of users when the sub-user is logged in. Additionally, a test for this case has been implemented. Let me know your thoughts.


beganovich:
@turbo124 any chance of getting this as an API feature? Like a filter.

turbo124:
sure, i've added in

?hideOwnerUsers=true which can be added to the query parameters on user route

beganovich:
Abedin can you revert these changes from the PR and make sure we use filter, this should simplify code a lot. Thanks!

Civolilah:
> Abedin can you revert these changes from the PR and make sure we use filter, this should simplify code a lot. Thanks!

@beganovich Sure thing, the approach has been changed, and I'm leaving the test as it is since the output should remain the same. @turbo124, I guess we need to await the updating of the API for merging this?

turbo124:
@beganovich ready for merge please.

beganovich (approved):

PR #1433, opened 2023-12-22
E-mail verification protection 

beganovich:
This wraps the email confirmation dialog using Turnstile.


beganovich:
Supported by https://github.com/invoiceninja/invoiceninja/pull/9062

turbo124:
As discussed, we'll look into this on api/web side.

PR #1435, opened 2023-12-25
[Bugfix] Bank Account Balance Formatting

Civolilah:
@beganovich @turbo124 The PR includes fixes for the `balance` value formatting for bank accounts in the table as well as on the Bank Account show page. So, from now on, we will take the `bankAccount.currency` property, using that property we will resolve the currency to get the `currency_id`, and then we will pass it to the `formatMoney` function to get the correct balance value format. Let me know your thoughts.

### DIFF
--- a/src/common/hooks/useResolveCurrency.ts
+++ b/src/common/hooks/useResolveCurrency.ts
@@ -10,8 +10,19 @@
 
 import { useCurrencies } from './useCurrencies';
 
-export function useResolveCurrency() {
+interface Params {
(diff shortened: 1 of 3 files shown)


turbo124:
does this handle null values?

Civolilah:
> does this handle null values?

@turbo124 Yes, there is always a fallback under the `useFormatMoney` hook. If we pass anything falsy, the formatting function will use `company.settings.currency_id`.

beganovich (approved):

PR #1437, opened 2023-12-27
[Cleanup] Select Field UI Change | Matching Searchable Select

Civolilah:
In progress...

### DIFF
--- a/src/components/forms/CustomSelectFieldOption.tsx
+++ b/src/components/forms/CustomSelectFieldOption.tsx
@@ -39,17 +39,7 @@ export function CustomSelectFieldOption({
         backgroundColor: isSelected || isFocused ? colors.$7 : colors.$1,
       }}
     >
-      {React.Children.toArray(children).map((element, index) =>
-        element ? (
-          React.isValidElement(element) ? (
-            React.cloneElement(element)
-          ) : (
-            element
-          )
-        ) : (
-          <div key={index}>{element}</div>
-        )
-      )}
+      {children}
     </Option>
   );
 }
--- a/src/components/forms/CustomSelectFieldOption.tsx
+++ b/src/components/forms/CustomSelectFieldOption.tsx
@@ -1,0 +1,55 @@
+/**
+ * Invoice Ninja (https://invoiceninja.com).
+ *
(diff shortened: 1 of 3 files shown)


Civolilah:
@turbo124 Quick confirmation from your end: I have created the new component, which will replace the current Select field. This is just an example of how it should look. Currently, the PR requires some additional adjustments, but the preview is prepared. So, right now, you can basically do anything you want under the `<option>` element, and it will be applied under the option in the dropdown. In this screenshot, I provided an example of how I transformed a single option to demonstrate that essentially anything is allowed. The selected option will be displayed with the same design. Screenshot:

Screenshot 2024-01-01 at 16 20 28

To significatly minimize the number of changes in the Select fields, we still need to nest everything within the `<option>` element and have the `value` property, which will later be used in the entire logic.

Please let me know if you agree with this default design and the potential for changes, so I can delve into the details and edge cases. I will then review all select fields in the app and prepare the PR for the final review.

turbo124:
why can i not type ahead with this input?

turbo124:
@Civolilah do we have an update on this one?

Civolilah:
> @Civolilah do we have an update on this one?

@turbo124 Yeah, so I stopped working on this one because of some other priorities that we had. However, most of the work or changes for the `SelectField` component itself are done. Also, the same work has been completed here: #1634.

I've been thinking a little bit about this task, and probably the best approach for us at the moment would be applying new design for the SelectField component entity by entity or page by page, similar to what we did for permissions.

If we convert all SelectFields to the new logic and design using the package, I believe there will definitely be a couple of cases that we won't cover, leading to potential issues.

So, probably the best approach that I see right now is once we merge #1634, just to go entity by entity or page by page and apply the prop to the SelectField component for transitioning the field to the new logic/design. Let's say I can do the conversion to new SelectFields for Clients/Invoices/Dashboard for the first PR, and then move further to get all of them done.

Let me know your thoughts.

turbo124:
@Civolilah sure, lets do it in stages. please close this one out if it is not longer needed.

Civolilah:
> @Civolilah sure, lets do it in stages. please close this one out if it is not longer needed.

@turbo124 No problem, I'll go in stages and close this PR since it's unnecessary!

PR #1438, opened 2023-12-28
[Bugfix] dashboard_charts props null values

Civolilah:
@beganovich @turbo124 The PR includes adding optional chaining and adding fallback values for properties of the `dashboard_charts` object. However, I was not able to recreate the bug in the production environment, and also locally, it seems that David found a very edge case here. Nevertheless, I believe that these changes will ensure 100% that we do not have any `null` values passed to the chart. Let me know your thoughts.

### DIFF
--- a/src/pages/dashboard/components/Totals.tsx
+++ b/src/pages/dashboard/components/Totals.tsx
@@ -92,9 +92,10 @@ export function Totals() {
   const [chartData, setChartData] = useState<ChartData[]>([]);
 
   const chartScale =
-    settings.preferences.dashboard_charts.default_view || 'month';
-  const currency = settings.preferences.dashboard_charts.currency || 1;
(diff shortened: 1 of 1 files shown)


turbo124:
@Civolilah can we cover missing dashboard_charts?. too please

Civolilah:
> @Civolilah can we cover missing dashboard_charts?. too please

@turbo124 The PR is updated with the addition of your suggestion for additional optional chaining on other props as well. However, the other suggestion of making `dashboard_charts` prop a default value in the context seems a bit unnecessary in this part of the logic. Here is the hook where it is already implemented (https://github.com/invoiceninja/ui/blob/develop/src/common/hooks/useReactSettings.ts). The way I currently implemented fallback values will cover if any of the earlier props is null/undefined. Let me know your thoughts.

beganovich (approved):

PR #1443, opened 2024-01-01
[Cleanup] Credits, Projects, Tasks Permissions && Tests Adjustment

Civolilah:
@beganovich @turbo124 The PR includes permission adjustments for Credits, Projects, and Tasks, along with test adjustments regarding UI changes when permissions are applied. Let me know your thoughts.


src/common/queries/credits.ts
```diff
@@ -37,7 +37,7 @@ export const useBulk = () => {
     }).then(() => {
       const message =
         successMessages[action as keyof typeof successMessages] ||
-        `${action}d_invoice`;
+        `${action}d_credit`;
```
Civolilah, line 40: @turbo124 Our tests discovered a minor bug. Instead of receiving messages for archived/deleted/restored credits, we received the same message that is used for invoices. Therefore, it's just a small change in the translation keyword.

turbo124:
ensure user has the correct permissions for Add To Invoice.

image

Needs to have admin/create_invoice permissions

Civolilah:
> ensure user has the correct permissions for Add To Invoice.
> 
> image
> 
> Needs to have admin/create_invoice compatible permissions
> 
> Convert to expense (from purchase order)
> 
> ensure admin/create_expense compatible permissions
> 
> image
> 
> There may be additional ones, please go over all menu actions and ensure any cross entity actions have the correct permissions.

@turbo124 I just double-checked the logic of the "add_to_invoice" action. Isn't the logic of this action to edit an existing invoice rather than creating a new one?

If I'm correct, then we will not need checks for the action, but we will need checks for the list of invoices that are provided to add a task to:

Screenshot 2024-01-08 at 21 08 08

After getting the list from the API, I've already implemented logic: if a user is an `admin` or has the `edit_invoice` permission, they will be able to add a task to a specific invoice. `OR`, if a user is `ASSIGNED` to the invoice either during `creation` or `manually` by the entity owner, they should also be able to do so. I hope my understanding is correct.

Additionally, the current PR covers `Credits`, `Projects`, and `Tasks`, but NOT `Purchase Orders`. I am addressing entities three by three for PRs related to permissions. The next PR will include Vendors, Purchase Orders, and Expenses. We anticipate completing permission adjustments very soon, with only two more PRs remaining, along with ensuring all tests pass successfully.

turbo124:
@Civolilah as long as the correct permissions are shown appropriately.

Civolilah:
> @Civolilah as long as the correct permissions are shown appropriately.

@turbo124 Sure thing! It has been completed for actions on almost all entities in the very first PR!

turbo124:
@Civolilah 

problem persists.

to replicate. give a sub user all permissions, but no permissions for invoices.

View a task, and add to invoice option still appears.

Civolilah:
> @Civolilah
> 
> problem persists.
> 
> to replicate. give a sub user all permissions, but no permissions for invoices.
> 
> View a task, and add to invoice option still appears.

@turbo124 As we discussed on Slack, currently, the "add_to_invoice" action will only be shown if the user is an `admin` OR has the "create_invoice" OR "edit_invoice" permission. Let me know your thoughts.

beganovich (approved):

PR #1444, opened 2024-01-03
[Cleanup] All Checkbox UX Improvement | Permissions

Civolilah:
@beganovich @turbo124 The PR includes an improvement of UX for the check button for `**_all` permissions.

Old behavior:

https://github.com/invoiceninja/ui/assets/51542191/1a51bca1-fb86-4c8f-bbec-c177507fc2f7

New behavior:

https://github.com/invoiceninja/ui/assets/51542191/9f9ccb34-9e5c-49d7-be4a-3bd8fb0706ce

Let me know your thoughts.


turbo124:
I think the main problem here is the UX is weird if a user clicks ALL, and then attempts to remove a single permissions.

Ideally, the ALL check box should just toggle on/off all the sub checkboxes.

If a user then attempts to click on another checkbox - ideally - the system should pop the _all permission, and add back all of the sub options explicity edit_invoice,edit_expense, edit_client etc etc

I understand this can be tricky because of the parent view_all and the logic around this. Alternatively, if the View / Edit / Create all options are clicked, then we should show the sub options as checked BUT disabled... this would make it much clearer.

Civolilah:
> I think the main problem here is the UX is weird if a user clicks ALL, and then attempts to remove a single permissions.
> 
> Ideally, the ALL check box should just toggle on/off all the sub checkboxes.
> 
> If a user then attempts to click on another checkbox - ideally - the system should pop the _all permission, and add back all of the sub options explicity edit_invoice,edit_expense, edit_client etc etc
> 
> I understand this can be tricky because of the parent view_all and the logic around this. Alternatively, if the View / Edit / Create all options are clicked, then we should show the sub options as checked BUT disabled... this would make it much clearer.

@turbo124 Ah, I totally agree. I did not notice that part of the functionality was an issue, so it has now been additionally fixed. Let me know your thoughts.

beganovich (approved):

PR #1453, opened 2024-01-09
[Bugfix] Select Field as Custom Field Does Not Have Selected Value By Default | Products Table

Civolilah:
@beganovich @turbo124 The PR includes fixes for a bug related to not selecting the value once the product is chosen in the line item. So, if the product has its own custom fields, and some of those fields are Select fields, and if the value is already selected in the product's object, it was not being applied to the field in the line item. Now, it is fixed. Let me know your thoughts.


src/pages/clients/edit/components/Contacts.tsx
```diff
@@ -47,7 +47,7 @@ export function Contacts(props: Props) {
 
     set(props.contacts[contactIndex], propertyId, value);
 
-    props.setContacts(props.contacts);
+    props.setContacts([...props.contacts]);
```
beganovich, line 50: Can you explain the reason for this? I thought this PR is only about fixing the default value.

beganovich (commented):
It looks good, tiny info is needed.

src/pages/clients/edit/components/Contacts.tsx
Civolilah: @beganovich After the change in the main component, I tested all custom fields and found that the value does not update correctly in some places. I rechecked the behavior, and it turns out that all other fields can update the value without creating a new reference, but in the case of the SelectField, it does not want to update the value. Specifically, there are three places where this change is needed: Client, Client Contact, and Vendor Contact fields.

So, that is the reason for this change. We were unable to find this bug earlier because it was defaultValue instead of value set on the SelectField. In all other places, we are creating a new reference while changing the value, which is why we did not encounter the issue. Therefore, this change is directly related to the main change aimed at resolving additional small bugs.

beganovich (approved):

PR #1464, opened 2024-01-12
[Bugfix] Updating Products Selector List After Query Invalidation | Reports/Schedules

Civolilah:
@beganovich @turbo124 The PR addresses the bug related to updating the list of products in the Product selector for Reports/Schedules. The bug was not due to invalidation, we had the correct invalidation in place. However, the issue originated from a logic problem within the selector. Let me know your thoughts.

### DIFF
--- a/src/pages/reports/common/components/ProductItemsSelector.tsx
+++ b/src/pages/reports/common/components/ProductItemsSelector.tsx
@@ -99,7 +99,7 @@ export function ProductItemsSelector(props: Props) {
 
   return (
     <>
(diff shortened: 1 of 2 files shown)


turbo124:
@Civolilah 

thanks looks good, please add a matching test also.

Civolilah:
> @Civolilah
> 
> thanks looks good, please add a matching test also.

@turbo124 Sure, test added!

beganovich (approved):

PR #1472, opened 2024-01-15
[Feature] Date Range DataTable Filter

Civolilah:
@beganovich @turbo124 The PR includes the implementation of a date range selector for the `date` and `due_date` properties in the main table of Invoices. Screenshot:

Screenshot 2024-01-18 at 04 46 33

Let me know your thoughts.


turbo124:
Thanks, looks good.

But.

There is no way to clear the date range after it has been selected.

Civolilah:
> Thanks, looks good.
> 
> But.
> 
> There is no way to clear the date range after it has been selected.

@turbo124 Yeh, you are right, that was a missed part of the functionality. I have just pushed changes that should allow you to clear the values. Let me know your thoughts.

turbo124:
After selecting a date range for invoice date, the invoice duedate has the same range?

Civolilah:
> After selecting a date range for invoice date, the invoice duedate has the same range?

@turbo124 Yeah, the `date_range` value was the same for the entire table. 

However, I just changed the behavior and set it according to the picker. Now, each column has its own range selected. Once the user changes the column that needs to be used for the `date_range` query parameter, the start and end date values that are selected for that column will be applied. Let me know your thoughts.

turbo124:
@Civolilah I think we'll need to adjust this.

We will want to be able to support filtering on both columns.

Can we change from date_range=column,start,end

to

```
date_range=start,end
```

and

```
due_date_range=start,end
```

So that we can support filtering on both columns please.

turbo124:
@Civolilah 

Looks really good, thanks! One thing would be nice to have for this is if we can highlight the header text IF a date range is active, just to assist users to understand that a filter is present? 

image

Civolilah:
> @Civolilah
> 
> Looks really good, thanks! One thing would be nice to have for this is if we can highlight the header text IF a date range is active, just to assist users to understand that a filter is present?
> 
> image

@turbo124 Here is the result, screenshot:

Screenshot 2024-01-22 at 01 25 37

You can check it out on the PR that has been pushed. Let me know your thoughts.

beganovich (approved):

PR #1478, opened 2024-01-19
[Bugfix] List Of Custom Templates When Emailing

Civolilah:
@beganovich @turbo124 The PR includes fixes for the list of custom templates on the email pages.

@beganovich Please just let me know once I need to PR this to the main branch to push it to production.

Let me know your thoughts.

### DIFF
--- a/src/pages/credits/email/Email.tsx
+++ b/src/pages/credits/email/Email.tsx
@@ -31,6 +31,9 @@ export default function Email() {
 
   const list = {
     email_template_credit: 'initial_email',
+    email_template_custom1: 'first_custom',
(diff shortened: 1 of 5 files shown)


turbo124:
@beganovich for hot patch please.

Civolilah:
@turbo124 I've made a slight adjustment to the solution after discovering we had a logic in displaying custom templates, which was not aligning with Flutter.

So, Ben can confidently proceed with pushing this to production. Just a quick confirmation from your end that the result is still correct. Thanks!

beganovich (approved):

PR #1484, opened 2024-01-23
[Bugfix] Task Creation Page Is Not Displayed Correctly In Edge Case

Civolilah:
@beganovich @turbo124 I noticed that task tests randomly fail because they do not contain anything on the Task creation page. This inconsistency occurs—sometimes all of them pass, but sometimes one or two of them does not. Upon investigation, I found that there was an unnecessary check for `taskStatuses` that caused this behavior. Once it occurs in one out of 1000 attempts, the page looks like this (without content):

Screenshot 2024-01-24 at 00 11 15

Let me know your thoughts.

### DIFF
--- a/src/pages/tasks/create/Create.tsx
+++ b/src/pages/tasks/create/Create.tsx
@@ -66,8 +66,7 @@ export default function Create() {
       if (
         typeof data !== 'undefined' &&
         typeof value === 'undefined' &&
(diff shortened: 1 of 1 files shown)


turbo124:
@beganovich not able to "QA" this as it is an edge case.

beganovich (approved):

PR #1485, opened 2024-01-24
[Bugfix] Sort/With Parameter Issue Combobox

Civolilah:
@beganovich @turbo124 The PR includes fixes for adding/removing 'sort' and 'with' parameters. Let me know your thoughts.

### DIFF
--- a/src/components/forms/Combobox.tsx
+++ b/src/components/forms/Combobox.tsx
@@ -916,10 +916,6 @@ export function ComboboxAsync<T = any>({
     }
   }, [data]);
 
-  const currentUrl = new URL(url);
-
-  console.log(currentUrl.searchParams.get('with'));
-
   useEffect(() => {
     setUrl((c) => {
       const currentUrl = new URL(c);
--- a/src/components/forms/Combobox.tsx
+++ b/src/components/forms/Combobox.tsx
@@ -825,7 +825,7 @@ export function ComboboxAsync<T = any>({
       return Boolean(entry);
     }
 
-    return false;
+    return true;
   };
 
   const { data, isLoading } = useQuery(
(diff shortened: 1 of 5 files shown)


src/components/forms/Combobox.tsx
```diff
@@ -883,6 +883,26 @@ export function ComboboxAsync<T = any>({
 
         if (inputOptions.value && inputOptions.value.toString().length > 0) {
           currentUrl.searchParams.set('with', inputOptions.value.toString());
+
+          if (currentUrl.searchParams.get('sort')) {
+            currentUrl.searchParams.delete('sort');
+          }
```
Civolilah, line 889: In the moment when the 'with' parameter is set, we need to ensure the removal of the 'sort' parameter if it exists.

src/components/forms/Combobox.tsx
```diff
@@ -883,6 +883,26 @@ export function ComboboxAsync<T = any>({
...
+      setUrl((c) => {
+        const currentUrl = new URL(c);
+
+        if (currentUrl.searchParams.get('with')) {
+          currentUrl.searchParams.delete('with');
+        }
+
+        if (sortBy) {
+          currentUrl.searchParams.set('sort', sortBy);
         }
```
Civolilah, line 906: When removing/clearing a value from the combobox, it is essential to ensure that the 'sort' parameter is included to obtain accurate results without the 'with' parameter (default list).

src/components/forms/Combobox.tsx
```diff
@@ -883,12 +883,32 @@ export function ComboboxAsync<T = any>({
...
```
Civolilah, line 906: When removing/clearing a value from the combobox, it is essential to ensure that the 'sort' parameter is included to obtain accurate results without the 'with' parameter (default list).

beganovich:
I think it looks. Is it possible to make test coverage for this?

Civolilah:
> Looks good. Is it possible to make test coverage for this?

@beganovich Sure, it's a bit tricky to test queries, but I managed to find a way to test those queries. It's implemented to handle cases where we should have the sort parameter but not the sort and reversed. Let me know your thoughts.

beganovich (approved):

PR #1489, opened 2024-01-25
[Bugfix] Fixes For Failing Tests

Civolilah:
@beganovich @turbo124 After finishing the permissions adjustment, 98% of tests are fixed and working great. This PR addresses only a few tests that are not covered by the permissions adjustment. Currently, locally, when I run all tests after seeding the initial database, all of them pass, more than 190 in total. However, on GitHub Actions, all of them fail. I believe this is due to a slight environmental adjustment needed there.

From the current status of tests, it appears that the login functionality in the app is unavailable. None of the tests are even executing their intended actions, they all fail at the login stage.

Therefore, as I see the situation now, I believe that fixing the connection with the API and DB will resolve this issue. Additionally, we need to ensure that the `VITE_ROUTER` environment variable is set to "browser" but NOT "hash," and `VITE_IS_TEST` variable is set to "true".

If I can assist in making any of these fixes, please let me know.


beganovich (approved):

beganovich:
I guess it should be a good idea to fix GitHub actions in this PR? Is that something you can have a  look at at some point @turbo124?

turbo124:
> I guess it should be a good idea to fix GitHub actions in this PR? Is that something you can have a look at at some point @turbo124?

I'm seeing green currently?

Civolilah:
@beganovich @turbo124 David made a PR today to fix actions, added some .env variables, and set up some other small things, but for some reason it still does not work. Writing a comment just to be aligned. Once I'm done with some other tasks, I can take a look at the whole workflow and try to replicate why it does not work the same as locally if you do not resolve it before me.

beganovich:
@Civolilah is this still draft?

Civolilah:
> @Civolilah is this still draft?

@beganovich Actually, David and I worked on making GitHub tests function the same as locally last night. We succeeded in fixing some bugs, but we still have an issue with triggering the API on GitHub actions. We believe the problem lies with Apache and the listening port, as we consistently receive a 404 error instead of a 200 for the login route.

Since this isn't a huge priority, I will work on it after finishing some priorities. Once I've completed some priorities, I will research a final solution to resolve this last issue of receiving a 404 error on API endpoints.

So, this will be ready for review once I've made the tests function the same as locally.

Civolilah:
> @Civolilah is this still draft?

@beganovich Yes, David and I worked on fixing GitHub actions to ensure the tests work the same as locally. We encountered a small problem with the API port, but David managed to find a solution. Currently, it is up to me to identify the reasons for some other minor bugs, and once I resolve them and pass the GitHub actions tests as we do locally, it will be ready for review.

We had some more important tasks than this one, so I addressed it once those were completed. However, I will prioritize fixing this asap and will keep you informed.

PR #1493, opened 2024-01-29
[Feature] Expense Report | Additional Selectors && Additional Test

Civolilah:
@beganovich @turbo124 The PR includes the implementation of new fields for the expense report and a simple tests to check their visibility once the expense report is selected. Screenshot:

Screenshot 2024-01-29 at 17 34 42

Let me know your thoughts.


turbo124:
@Civolilah 

when attempting to create a scheduled report, it does not bring across these new fields.

beganovich (approved):

Civolilah:
> @Civolilah
> 
> when attempting to create a scheduled report, it does not bring across these new fields.

@turbo124 Yeah, I understand you, but I think we have a small conflict between the types of properties in the `parameters` object. We already have the `clients` property under the `parameters` object for the `email_statement` template type.

However, the issue is that this property is an array of strings, compared to the `clients` property on reports for the newly added field, where the clients property is a comma-separated list of IDs.

So, if you can, please advise on the best option for the `clients` property on the `email_report` template type for this part of the scheduling functionality.

Civolilah:
@turbo124 After aligning on Slack, I implemented new fields on the Scheduling pages that are added for Expense reports. Additionally, I created tests for scheduling expense reports with those fields. Let me know your thoughts.

turbo124:
@Civolilah jira updated with this.

Civolilah:
> @Civolilah jira updated with this.

@turbo124 As we discussed on Slack regarding the issue with task schedulers for the Expense report, I have fixed all of them with one additional implementation:

- The bug related to the Clients selector has been fixed, and now removal from the selector is available, as well as all other actions.
- The `report_name` property has been fixed for almost all reports (as you mentioned in the ticket `aged_receivable_detailed_report` -> `ar_detailed`). Similarly, as you mentioned this one, I have fixed almost all others.

There are three reports left to fix (Purchase Order, Purchase Order Item, and Vendor), but I guess the issue is not related to the report_names that are sent from React (`purchase_order`, `purchase_order_item`, and `vendor`). Instead, it is related to the issue on the API that we discussed on Slack, right?

- The issue with selecting rows after attempting to change the Expense Category using the dropdown in the Expenses category has been fixed.
- Additionally, missing bulk actions along with a split button have been implemented on the edit task_scheduler page.

Let me know your thoughts.

turbo124:
@Civolilah I'm still not seeing the report_keys in the schedules?

ie. the report keys are not being passed into the schedule. We require these (if they have been set)

Civolilah:
> @Civolilah I'm still not seeing the report_keys in the schedules?
> 
> ie. the report keys are not being passed into the schedule. We require these (if they have been set)

@turbo124 Ah, okay, the part of logic for `report_keys` has also been added. If the user customizes columns on the report page and then triggers a scheduled action, the report_keys property will be populated with data from the Reports page. Let me know your thoughts.

turbo124:
image

I'm seeing an empty client array being passed in.

Civolilah:
> image
> 
> I'm seeing an empty client array being passed in.

@turbo124 Ah, okay, I found the case. It occurred only when scheduling a report with no selected clients. It appears to be working as expected on my end now. Please let me know your thoughts.

beganovich (approved):

PR #1506, opened 2024-02-01
[Feature] Customer Stripe URL

Civolilah:
@beganovich @turbo124 The PR includes implementation of a link to the customer Stripe page. So, the gateway token contains a `company_gateway_id`. I used that ID to find the `company_gateway` object from the pulled list from the API. After that, we have a `gateway_key` prop in `company_gateway`, and instead of searching for "cus_" in the `gateway_customer_reference` property, I will compare the `gateway_key` with the Stripe keys that David provided in the ticket. If it matches, then the icon/link to the customer Stripe page will be shown. I think this is the most secure and cleaner way. Screenshot:

Screenshot 2024-02-01 at 18 26 08

Let me know your thoughts.

### DIFF
--- a/src/pages/clients/show/components/Gateways.tsx
+++ b/src/pages/clients/show/components/Gateways.tsx
@@ -78,12 +78,11 @@ export function Gateways(props: Props) {
                 <div>
                   <MdPayment fontSize={22} />
                 </div>
-                <div className="flex items-center">
+                <div className="inline-flex items-center">
                   <span>{t('gateway')}</span>
                   <MdChevronRight size={20} />
                   <span>
-                    {getCompanyGatewayLabel(token.company_gateway_id) ||
-                      'Stripe'}
+                    {getCompanyGatewayLabel(token.company_gateway_id)}
                   </span>
                   {showCustomerStripeLink(token.company_gateway_id) && (
                     <Link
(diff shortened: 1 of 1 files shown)


turbo124:
image

We need to change:

1. Move the link to edit the gateway into the STRIPE breadcrumb if possible.

2. instead of the person icon, use the current (square/arrow) for the link to stripe please.

Civolilah:
> image
> 
> We need to change:
> 
> 1. Move the link to edit the gateway into the STRIPE breadcrumb if possible.
> 2. instead of the person icon, use the current (square/arrow) for the link to stripe please.

@turbo124 Sure, the UI is updated and it looks like this:

Screenshot 2024-02-04 at 17 40 53

Let me know your thoughts.

turbo124:
for stripe, can you just make the word Stripe in the breadcrumbs the link?

Civolilah:
> for stripe, can you just make the word Stripe in the breadcrumbs the link?

@turbo124 Sure thing, here is how it looks when the word "Stripe" is the only link there:

Screenshot 2024-02-05 at 10 53 20

Let me know your thoughts.

turbo124:
the links are around the wrong way.

Stripe => invocie ninja record

icon => stripe external link

Civolilah:
> the links are around the wrong way.
> 
> Stripe => invocie ninja record
> 
> icon => stripe external link



> the links are around the wrong way.
> 
> Stripe => invocie ninja record
> 
> icon => stripe external link



> the links are around the wrong way.
> 
> Stripe => invocie ninja record
> 
> icon => stripe external link

@turbo124 Ahh, okay. Just changed! So, the "Stripe" word link will navigate the user to the gateway edit page, but the icon will navigate the user to the external Stripe customer page.

beganovich (approved):

PR #1534, opened 2024-02-06
[Bugfix] Expenses/Recurring Expenses Amount Calculation

Civolilah:
@beganovich @turbo124 The client reported the issue on Slack (https://invoiceninja.slack.com/archives/C9KSLL0TE/p1707207816191849) that the expense amount does not include entered taxes by rate or by amount. On the table, we had only been showing the pure `expense.amount` value, and we didn't do anything with taxes there. So, I included calculations for taxes by rate or by amount to show the gross amount rather than the net amount. After implementation, I compared the results with the Flutter app, which always gave me exactly the same result. 

The exactly same issue occurred with Recurring Expenses and their amounts, so that has also been fixed.

Let me know your thoughts.

### DIFF
--- a/src/common/interfaces/company.interface.ts
+++ b/src/common/interfaces/company.interface.ts
@@ -65,6 +65,8 @@ export interface Company {
   fill_products: boolean;
   convert_products: boolean;
   bank_integrations: BankAccount[];
+  calculate_expense_tax_by_amount: boolean;
+  expense_inclusive_taxes: boolean;
 }
 
 export interface Settings {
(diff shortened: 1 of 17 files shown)


turbo124:
pls see jira

turbo124:
Calculations are wrong when using inclusive taxes.

When the taxes are inclusive you should not mutate the amount ever.

turbo124:
There is an unhandled edge case here.

1. Clone a recurring expense with taxes set by Tax Rate.
2. Change to Taxes by amount.
3. enter a tax amount.
4. front end sends both the tax rate and amount, creating an illegal state.

If user selects by rate (it should clear all tax_amounts) 
If user select by Amount (it should clear all tax rates and names

Civolilah:
> There is an unhandled edge case here.
> 
> 1. Clone a recurring expense with taxes set by Tax Rate.
> 2. Change to Taxes by amount.
> 3. enter a tax amount.
> 4. front end sends both the tax rate and amount, creating an illegal state.
> 
> If user selects by rate (it should clear all tax_amounts) If user select by Amount (it should clear all tax rates and names

@turbo124 Ah, great catch! I succeeded in recreating and fixing it. Let me know if it works correctly on your end also.

beganovich (approved):

PR #1538, opened 2024-02-08
[Bugfix] Trail Duration And Refund Period Changing Fields && Bulk Actions Implementation

Civolilah:
@beganovich @turbo124 The PR includes the implementation of bulk actions on the edit subscription page and changes the trail_duration and refund_period fields to dropdowns with seconds sent in the payload as values. Screenshot:

Screenshot 2024-02-08 at 11 09 26

Let me know your thoughts.

### DIFF
--- a/src/common/queries/subscriptions.ts
+++ b/src/common/queries/subscriptions.ts
@@ -10,10 +10,14 @@
 
 import { endpoint } from '$app/common/helpers';
 import { request } from '$app/common/helpers/request';
-import { useQuery } from 'react-query';
+import { useQuery, useQueryClient } from 'react-query';
 import { GenericSingleResourceResponse } from '$app/common/interfaces/generic-api-response';
 import { Subscription } from '$app/common/interfaces/subscription';
 import { useAdmin } from '$app/common/hooks/permissions/useHasPermission';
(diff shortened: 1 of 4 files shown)


turbo124:
@Civolilah 

Looks good, one small changed required.

If for example, Trial Duration has been set to One Day, but then later the user disables trials, we'll also want to ensure that we unset "" the trial duration to keep the request/response sane.

Civolilah:
> @Civolilah
> 
> Looks good, one small changed required.
> 
> If for example, Trial Duration has been set to One Day, but then later the user disables trials, we'll also want to ensure that we unset "" the trial duration to keep the request/response sane.

@turbo124 Ah, okay. It is improved right now. So, if you disable the `allow_cancellation` property, the `refund_period` property will be set to 0, but if you disable the `trial_enabled` property, the `trial_duration` property will be set to 0. I checked the Flutter app, and the empty option has a value of 0, so I'm considering it for unsetting the value. Is that okay?

Also, after checking the Flutter app, I noticed that we also missed this unsetting behavior there as well, so maybe it would be great to improve that there also. Let me know your thoughts.

beganovich (approved):

PR #1539, opened 2024-02-09
[Bugfix] Adding Permission On new_expense_category Action In Dropdown On Category Status Badge && Additional Tests

Civolilah:
@beganovich @turbo124 The PR includes adding permission for the `new_expense_category` action in the dropdown on the category status badge. So, this action should only be displayed if the user is an `admin` or `owner` because it is the only permission that can create the expense category. Additionally, I've added two tests to check the visibility of the action. Screenshot:

Screenshot 2024-02-09 at 16 45 26

Let me know your thoughts.

### DIFF
--- a/src/common/queries/expense-categories.ts
+++ b/src/common/queries/expense-categories.ts
@@ -14,7 +14,7 @@ import { useQuery } from 'react-query';
 import { Params } from './common/params.interface';
 import { ExpenseCategory } from '$app/common/interfaces/expense-category';
 import { GenericSingleResourceResponse } from '$app/common/interfaces/generic-api-response';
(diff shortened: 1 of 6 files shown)


turbo124:
Create expense should have permission to create category too

turbo124:
@Civolilah just to note on this one, when creating the expense category via modal ( from list, i am see the checkbox beneath the modal response to onclick.

Civolilah:
> @Civolilah just to note on this one, when creating the expense category via modal ( from list, i am see the checkbox beneath the modal response to onclick.

@turbo124 Ah, okay, I tested it out of the first three columns, so I did not notice it. It looks good on my end now. Let me know your thoughts.

beganovich (approved):

PR #1544, opened 2024-02-12
[Cleanup] Adjusting Logic For Manipulation With Line Items Without Selecting Client/Vendor

Civolilah:
@beganovich @turbo124 The PR includes adjusting the logic for manipulation with line_items on creation pages when the client/vendor is not selected. The only thing that is important to mention is that the formatting of money is done with company settings details (currency_id and country_id) instead of using client/vendor details for doing it. Once the user selects the client/vendor, formatting will be done using those details. Let me know your thoughts.

### DIFF
--- a/src/common/helpers/country/country-resolver.ts
+++ b/src/common/helpers/country/country-resolver.ts
@@ -1,0 +1,27 @@
+/**
+ * Invoice Ninja (https://invoiceninja.com).
+ *
+ * @copyright Copyright (c) 2022. Invoice Ninja LLC (https://invoiceninja.com)
+ *
+ * @copyright Copyright (c) 2022. Invoice Ninja LLC (https://invoiceninja.com)
+ *
+ * @license https://www.elastic.co/licensing/elastic-license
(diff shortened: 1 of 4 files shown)


src/common/helpers/country/country-resolver.ts
```diff
@@ -0,0 +1,27 @@
...
+ * @copyright Copyright (c) 2022. Invoice Ninja LLC (https://invoiceninja.com)
+ *
+ * @copyright Copyright (c) 2022. Invoice Ninja LLC (https://invoiceninja.com)
+ *
+ * @license https://www.elastic.co/licensing/elastic-license
+ */
+
+import { useStaticsQuery } from '$app/common/queries/statics';
+
+export function useCountryResolver() {
```
beganovich, line 13: We do have useResolveCountry. Can we use it?

src/common/hooks/useResolveCountry.ts

beganovich (commented):
Looks good, but needs a tiny clarification.

PR #1548, opened 2024-02-14
[Feature] SMTP Configuration

Civolilah:
@beganovich @turbo124 The PR includes adding a new SMTP field under the `email_settings` page. The value will be stored in the `company.settings.smtp`. I did not find this property by default, so I tried saving it in the `smtp` property, and it succeeded on my end. So, I guess it is the correct property?

Additionally, we have the `VITE_HOSTED_SHOW_SMTP_SETTINGS` env variable for enabling or disabling this feature for the `HOSTED` platform, but for `SELF-HOSTED`, it will always be available. On the `HOSTED` platform, the field will only be available for `PRO` and `ENTERPRISE` plans. Screenshot:

Screenshot 2024-02-14 at 19 22 08

Let me know your thoughts.


turbo124:
@Civolilah 

SMTP configuration would end up looking like this:

image

Each of those props are a string input except for the Encryption option, for that, the select would look like this:

<select name="encryption" class="input w-full form-select">
    <option value="tls">STARTTLS</option>
    <option value="ssl">SSL/TLS</option>
</select>


SMTP configuration is another Mail Driver option like Default / PostMark  etc... so you would add a new driver option SMTP followed by the appropriate form.

In regards to the column. This will be stored on the companies table - encrypted (i'll follow up with the column name). So once the configuration is POSTed, i would only ever return **** in the response for these fields. So you can expect a map/array of these values under a column ie

company.smtp.username

turbo124:
@Civolilah 

smtp_verify_peer could you please make this a boolean select field - default false please

Civolilah:
> @Civolilah
> 
> smtp_verify_peer could you please make this a boolean select field - default false please

@turbo124 If I understood your advice correctly, you want to have a select field instead of a toggle. So, I did it with an additional blank option. Also, we are missing "false" and "true" as translation keywords. If the solution is okay with you, please just add them. Screenshot:

Screenshot 2024-02-16 at 19 47 50

Let me know your thoughts.

Civolilah:
> > @Civolilah
> > smtp_verify_peer could you please make this a boolean select field - default false please
> 
> @turbo124 If I understood your advice correctly, you want to have a select field instead of a toggle. So, I did it with an additional blank option. Also, we are missing "false" and "true" as translation keywords. If the solution is okay with you, please just add them. Screenshot:
> 
> Screenshot 2024-02-16 at 19 47 50
> 
> Let me know your thoughts.

.

turbo124:
This particular field is a boolean true/false, so feel free to use a toggle/switch, the default should be true.

Civolilah:
> This particular field is a boolean true/false, so feel free to use a toggle/switch, the default should be true.

@turbo124 Ah okay, sure thing, here is the updated implementation. Screenshot:

Screenshot 2024-02-16 at 20 31 32

PR #1549, opened 2024-02-15
[Feature] Payment Slider

Civolilah:
@beganovich @turbo124 The PR includes implementing the Payment slider. The main difference between this slider and others that we have implemented is that we do not have a `History` and `Email History` tabs here. I noticed that the response of the payment actually does not include the `activities` array that we used in other entities, even though I clearly included `activities.history` in the query, the same as for other entities. Is there any issue, or do we not support history for this entity?

Also, when I tried to pull `email_history`, I received a 422 error indicating that the passed `entity` is invalid. So, I assume that for payments, we also do not support the Email History tab here?

Screenshot:

Screenshot 2024-02-16 at 18 52 31

Let me know your thoughts.

### DIFF
--- a/src/common/interfaces/payment-activity.ts
+++ b/src/common/interfaces/payment-activity.ts
@@ -1,0 +1,28 @@
+/**
+ * Invoice Ninja (https://invoiceninja.com).
+ *
+ * @link https://github.com/invoiceninja/invoiceninja source repository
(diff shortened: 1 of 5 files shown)


Civolilah:
.

beganovich (approved):

PR #1565, opened 2024-02-20
[Bugfix] Product Items Multi Selector Search Implementation

Civolilah:


Civolilah:
> @Civolilah
> 
> After configuring a Product Sales report with products and then scheduling it. When viewing the scheduled report, if you attempt to remove items from the products, there are issues with the incorrect products being removed/additional ones being added in.
> 
> This edge case may be caused by products with the same product_key (ie duplicate product_key)

@turbo124 Yeah, I succeeded in finding the edge case of duplicating product keys. I tested multiple ways of tracking the payload while removing/adding products on my end, and it works great. Please let me know your thoughts.

beganovich:
Can you explain:

> Please make sure you are testing this outside of strict mode.

Civolilah:
> Can you explain:
> 
> > Please make sure you are testing this outside of strict mode.

@beganovich We are handling the pulling of products that are not in the first list in the `useEffect` of the `ProductItemsMultiSelector` component (`lines 76-126`). Before, I fixed the issue that David reported in the StrictMode, where useEffect was called twice, resulting in duplicated values. Once I resolved the issue David reported regarding duplicated products in edge cases, it also addressed the problem with StrictMode.

beganovich:
Can you confirm that functionality behaves the same with/without strict mode?

Civolilah:
> Can you confirm that functionality behaves the same with/without strict mode?

@beganovich Yes, I can confirm that functionality behaves the same with/without strict mode.

PR #1573, opened 2024-02-23
[Feature] Import Templates

Civolilah:
@beganovich @turbo124 The PR includes the implementation for the templates in imports. Screenshots:

Screenshot 2024-02-27 at 15 42 06

Screenshot 2024-02-25 at 06 58 56

Note: If you agree with the translation keyword "save_sa_template" as the title on the modal, please just add it to the files.

Let me know your thoughts.

### DIFF
--- a/src/common/hooks/useReactSettings.ts
+++ b/src/common/hooks/useReactSettings.ts
@@ -44,6 +44,8 @@ export interface Preferences {
   };
 }
 
+type ImportTemplates = Record<string, Record<string, Record<number, string>>>;
+
 export interface ReactSettings {
   show_pdf_preview: boolean;
   react_table_columns?: Record<ReactTableColumns, string[]>;
@@ -54,6 +56,7 @@ export interface ReactSettings {
   table_filters?: Record<string, TableFiltersPreference>;
   common_actions?: Record<Entity, string[]>;
   show_mini_sidebar?: boolean;
+  import_templates?: ImportTemplates;
 }
 
 export type ReactTableColumns =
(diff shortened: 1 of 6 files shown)


Civolilah:
@turbo124 As we discussed, the flow of this feature has been slightly changed. Now, after successfully importing the file, we will offer to save the selected column_map as a template, if desired. Regarding the change in flow, I also made slight adjustments to the modal for saving the template. Let me know your thoughts.

Civolilah:
@turbo124 As we discussed today, there is one more modal added here for improving UX, before opening the final modal for saving the actual template. Once we get the translated keyword, it will be replaced with the text from the screenshot. Screenshot:

Screenshot 2024-02-28 at 16 48 19

Let me know your thoughts.

turbo124:
Please provide a Yes and No button options please.

Civolilah:
> Please provide a Yes and No button options please.

@turbo124 Sure, the "No" button is now also implemented. Screenshot:

Screenshot 2024-02-29 at 13 50 23

Let me know your thoughts.

turbo124:
@Civolilah 

this looks good, after the user  has acknowledged the modal, can you please redirect automatically to the listview of the entity please.

Civolilah:
> @Civolilah
> 
> this looks good, after the user has acknowledged the modal, can you please redirect automatically to the listview of the entity please.
> 
> also, we should standardize the delete / trash icons we are using, can you ensure this is consistent across the app
> 
> image

@turbo124 Sure, so redirection is adjusted just as we discussed. Also, the trash icon is always the same from the same package across the whole app. I just chose the red color instead of our convention to use the accent color. So, I changed just that detail. Screenshot:

Screenshot 2024-03-01 at 00 31 24

Let me know your thoughts.

beganovich (approved):

PR #1578, opened 2024-02-27
[Feature] Clients Document Table

Civolilah:
@beganovich @turbo124 The PR includes the implementation of a documents table on the show client page. Screenshot:

Screenshot 2024-02-28 at 01 43 42

Let me know your thoughts.


Civolilah:
@turbo124 I'm just not sure about one tiny thing here. We have statuses like public/private/image, etc. We filter statuses on other tables using the "status" or "client_status" query parameters, but on this one, I did not succeed in filtering data with those. I additionally tried with the "type" query parameter.

Please let me know which parameter name I can use for filtering documents.

Note: Also, currently, no matter what textual filter I enter, I get the same document list. It seems like the filtering functionality is not working currently?

turbo124:
@Civolilah because this tab is on the client tab itself, i don't think we should show to client as the linked to reference.

I've added a synthetic prop to documents.link

this will provide the constructed link to the resource ie invoices/xxxxx/edit

use the translation resource and the link reference  to create a link to the parent document instead please.

Civolilah:
> @Civolilah because this tab is on the client tab itself, i don't think we should show to client as the linked to reference.
> 
> I've added a synthetic prop to documents.link
> 
> this will provide the constructed link to the resource ie invoices/xxxxx/edit
> 
> use the translation resource and the link reference to create a link to the parent document instead please.

@turbo124 Sure, once we receive the `link` property in the response, the UI with the link will look like this. Screenshot:

Screenshot 2024-02-29 at 14 35 02

Let me know your thoughts.

turbo124:
@Civolilah looks really good :+1:

beganovich (approved):

PR #1589, opened 2024-03-01
[Cleanup] Uploading Docs Tasks Edit Page && Additional Tests

Civolilah:
@beganovich @turbo124 The PR includes the implementation of the `Documents` tabs for uploading documents for Tasks, along with additional tests for it. Screenshot:

Screenshot 2024-03-01 at 23 53 59

Let me know your thoughts.


turbo124:
@Civolilah looks good, thanks!

Civolilah:
> Is this the standard width we use for half-screen layouts or it was a bit wider? 2/3?
> 
> Also please default to w-full on mobile.

@beganovich Actually, we set the same width (`w-2/3`) on the Expense and Recurring Expense Documents page. So, to maintain consistency, I have applied it here as well.

Also, I changed the width for all screens except `large` to full width, but for large screens, it will be w-2/3.

Let me know your thoughts.

beganovich (approved):

PR #1610, opened 2024-03-11
[Bugfix] Rows Length On Import

Civolilah:
@beganovich @turbo124 The PR includes additional checks for row lengths on files uploaded under the Import | Export functionality. Therefore, if we encounter a single row/header, we will not consider those files as prepared for the final import, and we will display a toaster error message with the translation keyword `csv_rows_length`. This translation keyword should display a message similar to "Please check your files, as it appears the data is too short...". Please let me know if you agree with this translation keyword. Let me know your thoughts.

### DIFF
--- a/src/components/import/UploadImport.tsx
+++ b/src/components/import/UploadImport.tsx
@@ -243,29 +243,72 @@ export function UploadImport(props: Props) {
     setFormData(updatedFormData);
   };
 
+  const checkRowsLengthInCSV = (file: File) => {
+    return new Promise((resolve) => {
+      try {
+        const reader = new FileReader();
+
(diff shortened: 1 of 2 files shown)


turbo124:
[invoices-none.csv](https://github.com/invoiceninja/ui/files/14565359/invoices-none.csv)

Doesn't work when testing with this sample.

Civolilah:
> [invoices-none.csv](https://github.com/invoiceninja/ui/files/14565359/invoices-none.csv)
> 
> Doesn't work when testing with this sample.

@turbo124 Yeah, so the logic of checking row length is also added to the imports from the main table pages. Let me know your thoughts.

beganovich (approved):

PR #1620, opened 2024-03-15
[Feature] Implementing Table Footer

Civolilah:
@beganovich @turbo124 The PR includes the implementation for the table footers across all tables, which includes columns for summing values in them. Screenshot:

Screenshot 2024-03-16 at 19 34 50

`Note`: We are missing the "footer_columns" translation keyword in the files, so please add it there.

Let me know your thoughts.


turbo124:
@Civolilah only invoices / Recurring Invoices and quotes for now.

Civolilah:
> Member

@turbo124 Sure, the implementation is done for those three entities for now.

turbo124:
@Civolilah looks good, can we change the translation to just Footer, and also - we will need some form of ui to seprate the "footer" and columns, perhaps a Pipe or similar

image

Civolilah:
> @Civolilah looks good, can we change the translation to just Footer, and also - we will need some form of ui to seprate the "footer" and columns, perhaps a Pipe or similar
> 
> image

@turbo124 Sure, here is the changed translation and added separators between the words "Footer" and "Columns." Screenshot:

Screenshot 2024-03-18 at 10 15 00

Let me know your thoughts.

turbo124:
@beganovich thoughts here, should we convert these to buttons, or nest the footer option in user preferences?

beganovich:
It's getting quite busy there with so many labels. If we leave it there I'd suggest replacing both labels with ⚙️ and perhaps unifying the popup interface.

Otherwise, I vote for settings.

Civolilah:
> It's getting quite busy there with so many labels. If we leave it there I'd suggest replacing both labels with ⚙️ and perhaps unifying the popup interface.
> 
> Otherwise, I vote for settings.

@beganovich @turbo124 In my opinion, it appears to be a great option to implement something very similar to the popup that we have in the navigation bar. Screenshot:

Screenshot 2024-03-20 at 14 25 06

So, something similar but for table labels should work great, I think. Let me know your thoughts.

turbo124:
@Civolilah I think we should pop this in settings then > Settings > User Details > Preferences

Civolilah:
> @Civolilah I think we should pop this in settings then > Settings > User Details > Preferences

@turbo124 Sure thing, it is implemented through the user_preferences with the `show_table_footer` prop. Screenshot:

Screenshot 2024-03-21 at 16 53 45

Note: We are missing the "show_table_footer" translation keyword from files, so if you agree, please just add it there.

Let me know your thoughts.

turbo124:
@Civolilah 

Looks good, please also add in the helper text label:

show_table_footer_help

Civolilah:
> @Civolilah
> 
> Looks good, please also add in the helper text label:
> 
> show_table_footer_help

@turbo124 Sure, the help label has been added. Screenshot:

Screenshot 2024-03-22 at 13 54 34

Let me know your thoughts.

beganovich (approved):

PR #1622, opened 2024-03-16
[Cleanup] Reports Available For Pro/Enterprise On Hosted Platform

Civolilah:
@beganovich @turbo124 The PR includes making the Reports page available only to PRO/ENTERPRISE users on the `HOSTED` platform. Therefore, if a user has a free account, they will not have the `Reports` option in the main left app menu. If they try to access the page through the URL, they will see the message from the following screenshot. Screenshot:

Screenshot 2024-03-16 at 03 22 51

Let me know your thoughts.


Civolilah:
@turbo124 As we discussed last night, the implementation for the plan alert here has been changed. So, now the user will be able to access the Reports page but won't be able to use the actual actions there, they will be disabled. Screenshot:

Screenshot 2024-03-22 at 15 24 39

Let me know your thoughts.

src/pages/reports/common/components/ReportsPlanAlert.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export function ReportsPlanAlert(props?: CommonProps) {
+  const [t] = useTranslation();
+
+  const user = useCurrentUser();
+
+  return (
+    <>
+      {!proPlan() && !enterprisePlan() && (
+        <div className={props?.className}>
```
beganovich, line 29: Please, just return early null if a user is not pro/enterprise.

beganovich (approved):

beganovich (approved):

PR #1635, opened 2024-03-21
Conditionally show template entities

beganovich:
This shows template resources only if design.is_template.


turbo124:
Looks good!

PR #1638, opened 2024-03-22
[Feature] Creating useSetting hook

Civolilah:
@beganovich @turbo124 The PR includes the implementation of the `useSetting` hook, which will be used to retrieve the value of a specific prop from the settings. After initializing the hook with `const getSetting = useGetSetting();`, we only need to pass the `client` object as the first parameter and the `propertyKey` as the second one, and it will provide the value of the specific property. Let me know your thoughts.


Civolilah:
@turbo124 I've just made an additional tiny fix for the `industry_id` and `size_id` props. As we discussed this night regarding the issue with updating client settings, we noticed that null values were being sent for those two props in the payload. Therefore, I've simply added default values for those two props as for all other to avoid sending null values to the API.

src/common/hooks/useGetSetting.ts
```diff
@@ -0,0 +1,52 @@
...
+      if (client.settings[propertyKey] !== undefined) {
+        return client.settings[propertyKey];
+      }
+
+      if (
+        client.group_settings &&
+        client.group_settings.settings[propertyKey] !== undefined
+      ) {
+        return client.group_settings.settings[propertyKey];
+      }
```
turbo124, line 32: i think this block should be below the next client settings check? otherwise client.group_settings will override client.settings?

turbo124 (commented):
Please review logic here

src/common/hooks/useGetSetting.ts
Civolilah: @turbo124 Hmm, no. So, the logic for `client.settings` is above this logic on lines `23-24`. Then, this part of the logic is the second one when we are checking the `group_settings` object (Group Settings level) to see if the object exists in the Client object.

The logic below this one is when the `group_settings` object does not exist but the `group_settings_id` prop is populated, so we need to resolve the group_settings object from the pulled list, and finally, we have fallback on the company settings level.

Therefore, I'm sure that this does not override the Client settings level, the `client.settings` logic is just above. Let me know your thoughts.

turbo124: I'm not sure why you have the wrapping groupSettings && client... because

if client.settings.prop !== undefined..... return  it immediately.

Civolilah: > I'm not sure why you have the wrapping groupSettings && client... because
> 
> if client.settings.prop !== undefined..... return it immediately.

@turbo124 Yeah, so the reason for wrapping the whole logic first in two simple conditions for client && groupSettings is pretty straightforward.

I wrapped it in the `client` part of the condition so we do not have to do it in the logic where we call the function. We will probably always need to check `if client !== undefined` in the logic where the function is called because we are aware that the result of the function will not be what we EXPECT if the client is `undefined`. We know that we will get the first value from the company or group level settings for sure.

The second condition is related to the groupSettings array, which is fetched from the API or pulled from the cache. So, let's imagine this scenario: we are expecting to get the value from this function, and it needs to be from the group settings level, but the object `group_settings` does NOT exist in the passed client object. So, basically, if we do not have this check, the first value which will be returned from the function will be incorrect because there is a case for sure when the API call for groupSettings is not finished, and the function is returning the company level value.

So basically, we are returning the value only when we are sure that it should be the correct one or let's say when we have the client object !== undefined and the groupSettings array pulled from the API. Otherwise, we have those checks for edge cases that might happen.

The second thing you mentioned is checking the value if it is undefined or not `(if client.settings.prop !== undefined)`. The reason we cannot return it immediately is simply because if I return it like this `return client.settings.prop`, the logic will always return what is on the client settings level, it will never go to the next level to check if there is a populated value.

There is only immediate return for the company level, which is the last one (`return company.settings.prop`), because we are aware that there are no other checks after that and we can return the value. Without if statements in the first three cases, we will always get only the first one returned.

beganovich (approved):
This looks fine, but what would be even more confidence-inspiring is a unit test. Is it possible to make it?

Civolilah:
> This looks fine, but what would be even more confidence-inspiring is a unit test. Is it possible to make it?

@beganovich Sure, I will find a way to test this one and implement tests for the main cases.

Civolilah:
> This looks fine, but what would be even more confidence-inspiring is a unit test. Is it possible to make it?

@beganovich I can confirm that the test for the useGetSetting hook for one property (`military_time`) is implemented for all settings levels.

beganovich (approved):

PR #1642, opened 2024-03-23
[Bugfix] Adding bcc_email and send_time fields for SMTP config

Civolilah:
@beganovich @turbo124 The PR includes adding the "bcc_email" and "send_time" fields to the SMTP configuration. Screenshot:

Screenshot 2024-03-23 at 17 00 40

Let me know your thoughts.

### DIFF
--- a/src/pages/settings/email-settings/common/components/SMTPMailDriver.tsx
+++ b/src/pages/settings/email-settings/common/components/SMTPMailDriver.tsx
@@ -16,6 +16,7 @@ import { Element } from '$app/components/cards';
 import { Button, InputField, SelectField } from '$app/components/forms';
 import Toggle from '$app/components/forms/Toggle';
 import { useHandleCurrentCompanyChangeProperty } from '$app/pages/settings/common/hooks/useHandleCurrentCompanyChange';
+import dayjs from 'dayjs';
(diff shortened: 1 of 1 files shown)


Civolilah:
@turbo124 Quick confirmation on this one: Should those two new fields be added to the payload of the `/api/v1/smtp/check endpoint`?

turbo124:
i think the main point here is why are these two fields included in the settings for the email provider? these should be independent from the email provider as they are do not interact at all with the email provider at all?

Civolilah:
> i think the main point here is why are these two fields included in the settings for the email provider? these should be independent from the email provider as they are do not interact at all with the email provider at all?

@turbo124 Actually, that was my bad. When I created the logic for the SMTP email provider, I included only the fields that you mentioned in the ticket, thinking that those two fields were related to the email provider because they were positioned below the email_provider selector.

Therefore, they were only excluded for the SMTP email_provider, but for any other provider, we have included them.

Since those two fields are not related to the email_provider, what do you think about moving them above the email_provider selector?

Civolilah:
> Does this need any further changes?

@beganovich I think not, but let's leave the final confirmation for @turbo124.

beganovich (approved):

PR #1645, opened 2024-03-26
[Feature] E-Quote Type

Civolilah:
@beganovich @turbo124 The PR includes the implementation of a couple of features:

- Implementing the "e_quote_type" field under the Email Settings page.
- Implementing the "routing_id" field on the Vendor edit page.
- Implementing the "download_e_quote," "download_e_credit," and "download_e_purchase_orders" actions

`Note`: We missed the translation keywords "e_quote_type" and "download_e_purchase_order" from the files. If you agree, please just add them there.

Let me know your thoughts.


Civolilah:
@turbo124 I pulled the latest changes from the API today, and it looks like the "download_e_quote" action works for me without any errors. However, the "download_e_credit" and "download_e_purchase_order" actions still show the 500 error. The logic of implementation is exactly the same for all of them. Let me know your thoughts.

turbo124:
@Civolilah looks good, the issue with credits is api side.

beganovich (approved):

PR #1648, opened 2024-03-26
[Bugfix] Fixing Date Formatting On Range Picker Value Change

Civolilah:
@beganovich @turbo124 The PR includes fixes with date formatting on the range picker on the invoices table and dashboard. So, the issue was related to the `dayjs` library not supporting our two formats (`'DD. MMM. YYYY'` and.`'ddd MMM D, YYYY'`). Once we select those two formats, we should not pass them to the `dayjs` function. Only if we do not pass the format, it will be formatted correctly. Let me know your thoughts.


turbo124:
@beganovich this one for patch to prod please.

beganovich (approved):

PR #1653, opened 2024-03-28
[Feature] Implementing Import Customers Payment Gateway Action

Civolilah:
@beganovich @turbo124 The PR includes the implementation of the "import_customers" payment gateway action. Screenshot:

Screenshot 2024-03-28 at 19 36 56

Let me know your thoughts.


turbo124:
@Civolilah looks like the URL is not being injected with the :id prop

Civolilah:
> @Civolilah looks like the URL is not being injected with the :id prop

@turbo124 Ah, my bad. It should be fixed right now. On my end, it works just as expected. Let me know your thoughts.

beganovich (approved):

PR #1656, opened 2024-03-29
[Cleanup] Next Send Date Property Formatting In Date Time Format | Recurring Invoices

Civolilah:
@beganovich @turbo124 The PR includes formatting the "next_send_date" column for Recurring Invoices. It involves formatting it to the date-time format using the formats entered in the company settings along with the "military_time" property. This has been done in three places:

- All Recurring Invoices tables including the dashboard table
- Recurring Invoice Slider

Screenshot:

Screenshot 2024-03-30 at 02 36 52

Let me know your thoughts.

### DIFF
--- a/src/common/helpers.ts
+++ b/src/common/helpers.ts
@@ -15,6 +15,7 @@ import { route } from '$app/common/helpers/route';
 import entityState from './constants/entity-state';
 import { request } from './helpers/request';
 import { useCurrentCompanyDateFormats } from './hooks/useCurrentCompanyDateFormats';
+import { useCompanyTimeFormat } from './hooks/useCompanyTimeFormat';
 
 export function isHosted(): boolean {
   return import.meta.env.VITE_IS_HOSTED === 'true';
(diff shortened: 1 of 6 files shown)


turbo124:
I don't think this is correct. Are you using the prop

```
next_send_datetime
```

This is the UTC send time, which we'll want to convert to the users local time based on the timezone selected.

Civolilah:
> I don't think this is correct. Are you using the prop
> 
> ```
> next_send_datetime
> ```
> 
> This is the UTC send time, which we'll want to convert to the users local time based on the timezone selected.

@turbo124 Ah, okay, there was a little misunderstanding. Now, everything should be correctly formatted. I utilized the `next_send_datetime` property, and additionally, the UTC time provided within this property has been converted into the time zone selected in the company settings, along with the adjustment for military time. Let me know your thoughts.

turbo124:
Still not appearing as expected. I am now seeing the UTC datetime:

image

for example here, this company timezone is australia/sydney - and the send time 9am. so it should show 3rd april 9am, instead of its utc equivalent.

beganovich (approved):

PR #1669, opened 2024-04-04
[Feature] Implementing client_brevo Mailer

Civolilah:
@beganovich @turbo124 The PR includes the implementation of a new `client_brevo` mailer, which is exactly the same as `client_postmark` with field setup, except for the secret field, which will be saved in the `brevo_secret` property. Screenshot:

Screenshot 2024-04-04 at 21 01 22

Let me know your thoughts.


Civolilah:
@turbo124 The mailer has been added for both platforms, `HOSTED/SELF-HOSTED`. I hope that is the correct setup.

beganovich (approved):

PR #1670, opened 2024-04-04
[Feature] Implementing Expense Preview | Scaffold

Civolilah:
@beganovich @turbo124 The PR includes the implementation of a "preview" mode for documents on the Expense edit page. Therefore, I believe that we are only able to natively preview jpg, jpeg, svg, webp, tiff, png, and pdf file types. All other types require the extension for displaying them. Here is a screenshot of how the preview looks when we can display the file:

Screenshot 2024-04-05 at 20 37 54

Here is the screenshot when the file type cannot be shown natively in React:

Screenshot 2024-04-05 at 20 44 24

Let me know your thoughts.

### DIFF
--- a/src/components/FileIcon.tsx
+++ b/src/components/FileIcon.tsx
@@ -10,17 +10,23 @@
 
 import { File, Jpg, Pdf, Png, Svg } from './icons';
 
-export function FileIcon(props: { type: string }) {
+interface Props {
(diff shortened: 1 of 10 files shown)


turbo124:
Looks great!

beganovich (approved):

PR #1671, opened 2024-04-05
[Bugfix] JSON parsing For FormData Payload

Civolilah:
@beganovich @turbo124 The PR includes fixes for parsing the payload in the response. The logic is created for invalidating `activities` queries when deleting/updating entities. However, there is a missed case when the payload is `formData` or when uploading documents, which causes an issue with not returning a response once the upload method succeeds. Furthermore, the logic within the then block is not invoked, it only catches the error when parsing the payload is not possible. Let me know your thoughts.


beganovich:
This feels kinda odd to globally if and then refetch activities. When do we run in this case? When uploading documents? 

Can we just wrap that part with this try-catch?

Civolilah:
> This feels kinda odd to globally if and then refetch activities. When do we run in this case? When uploading documents?
> 
> Can we just wrap that part with this try-catch?

@beganovich So, the `activities` endpoint needs to be refetched on any `DELETE` entity action or if any entity is `UPDATED/CHANGED` ('PUT' method). Thus, I believe that there will be a significant number of cases where we need it, so basically I implemented it in the interceptor.

I researched additionally and improved the solution we have here. So, basically, the issue arises when the payload sent in the request is not valid JSON, we encounter an issue, and the response in the interceptor is never returned. This occurs only when uploading documents, which is what I noticed. However, the core of the issue is that we need to determine what can be parsed or not, so we can further check if the refetching for activities needs to be triggered. Let me know your thoughts.

beganovich:
Since the fn has changed significantly, @turbo124 can you QA this?

turbo124:
@Civolilah do playwright tests pass in full? I would imagine as this is a global modification, we would need to be careful that all cases are covered and we don't create a regression.

turbo124:
@Civolilah ping for comment

Civolilah:
> @Civolilah ping for comment

@turbo124 , @beganovich and I just tested the functionality and the PR, everything works fine for us. Also, I ran tests and all tests for uploading docs passed along with other tests. Additionally, Ben and I discussed the solution, and this one will be fine even at the global level. It is carefully checked to include only what we need: the method ('put', 'delete', 'post'), after which we just check the `action` property in the payload. If all passes, we simply refetch the activities query. So, the logic is very simple, and it will help us avoid a lot of repetition. Let me know your thoughts.

beganovich (approved):

PR #1672, opened 2024-04-06
[Cleanup] Removed Switch To Flutter Option

Civolilah:
@beganovich @turbo124 The PR includes removing the option to switch to Flutter. Screenshot:

Screenshot 2024-04-06 at 17 27 25

Let me know your thoughts.


beganovich:
What about component? Can we remove it if there's no other reference to it.

Civolilah:
> What about component? Can we remove it if there's no other reference to it.

@beganovich Ah, sure, the component has also been removed, we do not have any additional references.

beganovich (approved):

PR #1673, opened 2024-04-06
[Bugfix] Fixes For Opening ninja_portal_url In New Tab | Templates And Reminders

Civolilah:
@beganovich @turbo124 The PR includes fixes for opening new tabs for `ninja_portal_url` under the Templates and Reminders section for the Body field. Screenshot:

Screenshot 2024-04-06 at 18 19 29

Let me know your thoughts.


beganovich:
Don't we have a couple of other references to the Button with `external` prop on itself? 
Is there a bug with `external` or?

Civolilah:
> Don't we have a couple of other references to the Button with `external` prop on itself? Is there a bug with `external` or?

@beganovich Hmm, we do not have an `external` property on the `Button` component. We do have a `to` property, but that will navigate us through our app only, not open new tabs. So, I decided to go with this approach, which I believe is very simple and clear. It's just configuring the `onClick` function with one line of code. Let me know your thoughts.

beganovich (approved):

PR #1675, opened 2024-04-07
[Feature] Chrome Extension Banner | Tasks

Civolilah:
@beganovich @turbo124 The PR includes the implementation of a banner for the Chrome extension on the Tasks main page. Screenshot:

Screenshot 2024-04-08 at 19 17 02

Let me know your thoughts.


turbo124:
@Civolilah looks good. we should only show this banner once. once it has been dismissed, it should not reappear.

Civolilah:
> @Civolilah looks good. we should only show this banner once. once it has been dismissed, it should not reappear.

@turbo124 Okay, so the logic was implemented through `sessionStorage`, which is the reason for having the banner appear once you log out of the app or start a new session. However, it is now implemented through `localStorage`, and once the user dismisses the banner, it will never be shown again. Let me know your thoughts.

beganovich (approved):

PR #1679, opened 2024-04-08
[Cleanup] Validation Message Improving | Clients Limit

Civolilah:
@beganovich @turbo124 The PR includes improving the validation message for clients when we encounter a validation error for the 'id' property. Screenshot:

Screenshot 2024-04-09 at 17 22 20

`Note`: I utilized the translation keywords `clients_limit` and `unlock_unlimited_clients` to retrieve the sentences "Amazing! You've hit the 20 client limit on Free accounts. Congrats on your success!" and "Please upgrade to unlock unlimited clients!" respectively. If you agree with those keywords, please just add them to the files.

Let me know your thoughts.


beganovich (commented):
Seems like we're using this only on create.. for the simple if the case for `id` looks like we are abstracting two steps away.. is this  _really_ needed?

Civolilah:
> Seems like we're using this only on create.. for the simple if the case for `id` looks like we are abstracting two steps away.. is this _really_ needed?

@beganovich As we discussed on Slack, it has been refactored using a less globalized solution. However, if we need it in the future, it will be easy to improve. Let me know your thoughts.

PR #1681, opened 2024-04-09
[Cleanup] Implementing idempotency_key In POST method | Payments

Civolilah:
@beganovich @turbo124 The PR includes the implementation of an `idempotency_key` (64-character hex) for the POST method on Payments. The property will be sent in the body of the request along with the rest of the Payment details. Let me know your thoughts.


Civolilah:
@turbo124 The `/api/v1/payments` route has more than one POST metho, it is used for refunding payments and for bulk actions (such as archiving/deleting, etc.). Currently, I have implemented this for the POST method when CREATING the Payment, but please let me know if it needs to be implemented for the rest of them as well.

turbo124:
@Civolilah only the initial record creation for this particular item currently.

src/common/hooks/useGenerateHash.ts
```diff
@@ -0,0 +1,20 @@
...
+ * @link https://github.com/invoiceninja/invoiceninja source repository
+ *
+ * @copyright Copyright (c) 2022. Invoice Ninja LLC (https://invoiceninja.com)
+ *
+ * @license https://www.elastic.co/licensing/elastic-license
+ */
+
+import { sha256 } from 'js-sha256';
+
+export function useGenerate64CharHash() {
```
beganovich, line 13: Is there some specific reason this is a hook?

It doesn't access React's state nor lifecycle?

Civolilah: @beganovich As we discussed on Slack, the approach has changed slightly. Instead of a hook, we simply have the function for generating the hash. Let me know your thoughts.

beganovich (approved):

PR #1682, opened 2024-04-09
[Bugfix] Auto-Fill Quantity/Hours | Products/Tasks Tables

Civolilah:
@beganovich @turbo124 The PR includes fixing a bug for auto-filling the quantity/hours for the products/tasks table when the `fill_products` property is turned off. Now, the quantity/hours will be the same as the user entered earlier. Let me know your thoughts.

### DIFF
--- a/src/pages/invoices/common/hooks/useHandleProductChange.ts
+++ b/src/pages/invoices/common/hooks/useHandleProductChange.ts
@@ -41,9 +41,11 @@ export function useHandleProductChange(props: Props) {
       return props.onChange(index, lineItem);
     }
 
-    lineItem.quantity = company?.default_quantity ? 1 : product?.quantity ?? 0;
-
     if (company.fill_products) {
+      lineItem.quantity = company?.default_quantity
+        ? 1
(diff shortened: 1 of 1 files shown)


Civolilah:
@turbo124 `Quick confirmation`: You mentioned in the ticket, or I just understood you as requiring us to ensure that the Tasks table is not auto-filled. However, the bug also occurred on the Products table, where the `quantity` property was auto-filled even when the `fill_products` option is turned off. So, just to confirm, should it be for both products (quantity) and tasks (hours)?

turbo124:
From my testing, the auto fill did not impact product table currently. But yes. if auto fill is disabled, we should never inject / mutate the row. ever. on products or tasks table.

turbo124:
@Civolilah this is fine for merge as is. But I am noticing that there is a slight difference with the way the tables are reacting when a product is selected.

On product table, selecting product does not update any columns.

On tasks table, if description is empty, the description of the product is inserted.

beganovich (approved):

beganovich (approved):

PR #1685, opened 2024-04-10
[Cleanup] Moving "E-Quote Type" Field Above the "Divider Line"

Civolilah:
@beganovich @turbo124 The PR includes moving the `e_quote_type` field above the "divider_line" to just below the `e_invoice_type`. Screenshot:

Screenshot 2024-04-10 at 20 08 52

Let me know your thoughts.


turbo124:
@Civolilah  Need to hide e-quote type also if enable e-invoice is not toggled on.

Civolilah:
> @Civolilah Need to hide e-quote type also if enable e-invoice is not toggled on.

@turbo124 Okay, so the 'e_quote_type' selector will be shown only if the 'enable_e_invoice' prop is set to `true`. Let me know your thoughts.

beganovich (approved):

PR #1692, opened 2024-04-12
[Bugfix] Updating Value On Custom Design (Body, Footer etc.)

Civolilah:
@beganovich @turbo124 The PR includes fixes for updating values such as body, footer, etc., for custom designs. Let me know your thoughts.


turbo124:
There are problems when you add content now. try adding content and then click save.

beganovich (approved):

PR #1694, opened 2024-04-12
[Cleanup] Avoiding Third Party Links On Self-Hosted | RSMS-INTER css

Civolilah:
@beganovich @turbo124 The PR includes adjusting the logic in the website header to avoid the usage of third-party URLs on `SELF-HOSTED`. Previously, we had the third-party URL (`https://rsms.me/inter/inter.css`) set in the header, importing the CSS. However, now for the SELF-HOSTED platform, we will import that CSS from a locally added file. For `HOSTED`, the mentioned third-party URL will still be used instead of a local file. Let me know your thoughts.


turbo124:
@beganovich does this work with our cf deployment?

beganovich:
It should yep as it's just a static asset. 

However, I wonder why don't we just bundle it always? Why ever load from the 3rd party.. would be much simpler.

turbo124:
@beganovich i think bundling makes sense.

Civolilah:
> @beganovich i think bundling makes sense.

@beganovich @turbo124 Sure, then we will simply have local loading of the needed files, avoiding third-party URLs on the `HOSTED` platform as well. It's done. Let me know your thoughts.

beganovich (approved):

PR #1695, opened 2024-04-13
[Feature] Implementing Dnd For Task Statuses (Columns) | Kanban

Civolilah:
@beganovich @turbo124 The PR includes the implementation of sorting columns/dnd (task_statuses) on the Kanban page and fixing a UX bug with the sorting of tasks under those columns. Let me know your thoughts.


turbo124:
@Civolilah 

There are a few issues here.

Columns can be reordered, however tasks inside the columns do not DnD anymore.

also, i note that the API is hit on ever DnD event, even if the order does not change.

Ensure DnD can function for both the Task Status columns and tasks inside each column
And ensure that we only hit the API when a change in order/status is actually performed.

Civolilah:
@turbo124 The implementation for combining DND for columns/task statuses and cards/tasks has been completed. Let me know your thoughts.

beganovich:
Looks like it's behaving weird... same behavior on FF and Chrome (latest versions).

https://github.com/invoiceninja/ui/assets/13711415/3ac45fc5-8425-45fa-a360-149ff31e79bf

Civolilah:
> Looks like it's behaving weird... same behavior on FF and Chrome (latest versions).
> 
>  Screencast.from.2024-05-07.19-18-26.mp4

@beganovich Ah, great catch. The reason for this behavior is that I used task_id and task_status_id as the draggableId prop. However, in some cases, the IDs are the same, which causes the issue. When you drag the card/task, the whole column is also dragged because the ID for task_status is the same. I've ensured that the draggableId is unique, and on my end, everything works correctly. Let me know your thoughts.

beganovich (approved):
Looks good, thanks 👍️

PR #1700, opened 2024-04-14
[Bugfix] Adjusting Landing Page URL When Switching Companies

Civolilah:
In progress...

### DIFF
--- a/src/pages/settings/company/create/CompanyCreate.tsx
+++ b/src/pages/settings/company/create/CompanyCreate.tsx
@@ -23,6 +23,7 @@ import { useQueryClient } from 'react-query';
 import { useDispatch } from 'react-redux';
 import { ValidationBag } from '$app/common/interfaces/validation-bag';
 import { useColorScheme } from '$app/common/colors';
+import { useNavigate, useParams } from 'react-router-dom';
 
 interface Props {
   isModalOpen: boolean;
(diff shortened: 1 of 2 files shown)


src/pages/settings/company/create/CompanyCreate.tsx
```diff
@@ -56,6 +60,16 @@ export function CompanyCreate(props: Props) {
...
 
+    if (id) {
+      const basePage =
+        '/' +
+        (location.pathname.includes('/settings/gateways')
+          ? 'settings/online_payments'
+          : location.pathname.split('/')[1]);
+
+      navigate(basePage);
+    }
```
beganovich, line 71: Is it possible to refactor this using URL API?

beganovich, line 71: Is it possible to refactor this using URL API?

Civolilah: > Is it possible to refactor this using URL API?

@beganovich Hmm, I don't think so. We still need to split the pathname by '/', because even when using the URL API, we can't get exactly '/clients'. Screenshot:

Screenshot 2024-04-15 at 17 56 39

Let me know your thoughts.

beganovich (approved):

PR #1705, opened 2024-04-16
[Feature] Implementing Copy To Clipboard Icon For Number Columns

Civolilah:
@beganovich @turbo124 The PR includes the implementation of a copy-to-clipboard icon for number columns on `Invoices`, `Quotes`, `Credits`, and `Purchase Orders`. Screenshot:

Screenshot 2024-04-16 at 17 29 36

Let me know your thoughts.


beganovich (commented):
This really feels like overengineering, to be honest. Can't we just add a copy to the clipboard button to a few table columns? 

Is the whole component that you can't even expand upon, really needed?

Civolilah:
> This really feels like overengineering, to be honest. Can't we just add a copy to the clipboard button to a few table columns?
> 
> Is the whole component that you can't even expand upon, really needed?

@turbo124 Yeah, with this implementation, I aimed to reduce the clutter in the column hooks a bit. However, I made the change without creating any specific hooks for this feature. Let me know your thoughts.

src/components/CopyToClipBoardIconOnly.tsx
```diff
@@ -9,24 +9,28 @@
...
   className,
   secure,
   limit = 0,
+  stopPropagation,
 }: Props) {
   const value = text || '';
 
-  const handleClick = () => {
+  const handleClick = (event: MouseEvent<HTMLButtonElement>) => {
+    stopPropagation && event.stopPropagation();
```
beganovich, line 33: Can you guide me what this has to do with this PR? Thanks.

beganovich (commented):
Looks good, just a tiny clarification needed 👍

src/components/CopyToClipBoardIconOnly.tsx
Civolilah: @beganovich Actually, we open the slider when the user clicks on one of the first three columns, and this can be one of them, which is actually quite common. So, to prevent the slider from opening when the user clicks on the icon just to copy the number, I implemented stopPropagation here. This event's function will simply disable the parent's onClick event, which, in our case, is opening the slider. I believe that opening the slider at the moment of copying the number breaks the UX. Let me know your thoughts.

beganovich (approved):

PR #1711, opened 2024-04-18
[Cleanup] Minimizing Actions List Dropdown on the Invoices Table

Civolilah:
In progress...


Civolilah:
@beganovich I tried a few solutions here to minimize the height of more_actions dropdown, especially for entities where we have a bunch of actions, but actually we did not get the right solution in those few attempts. As we have more prioritized tasks, we just stopped working on this one. Maybe this one will be great to close as new design is coming, so we can think about this improvement after refactoring the design of tables and actions there.

PR #1722, opened 2024-04-22
Set return as BR not new paragraph

turbo124:


turbo124:
@beganovich for hotpatch pls.

beganovich (approved):

PR #1729, opened 2024-04-24
[Feature] Implementing private_notes On The Project Show Page

Civolilah:
@beganovich @turbo124 The PR includes the implementation of the "private_notes" on the project show page, but the notes will only be displayed if the user is an `ADMIN` or `OWNER`. Screenshot:

Screenshot 2024-04-24 at 20 19 08

Let me know your thoughts.

### DIFF
--- a/src/components/InfoCard.tsx
+++ b/src/components/InfoCard.tsx
@@ -9,6 +9,7 @@
  */
 
 import { useColorScheme } from '$app/common/colors';
+import classNames from 'classnames';
 import { CSSProperties, ReactNode } from 'react';
 
 interface Props {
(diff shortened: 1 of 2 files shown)


turbo124:
@Civolilah permission level for this is only view _ project.

Project private notes are only hidden from the client.

Civolilah:
> @Civolilah permission level for this is only view _ project.
> 
> Project private notes are only hidden from the client.

@turbo124 Okay, so I've removed the admin/owner permission. However, the show page of the Project is only visible if the user has at least a `view_project` permission, so it will work as expected. Let me know your thoughts.

beganovich (approved):

PR #1736, opened 2024-04-28
Use foreign amount if available for expenses

turbo124:


turbo124:
@beganovich for hotpatch please.

beganovich (approved):

PR #1737, opened 2024-04-30
[Bugfix] Refetch Tasks Query After Invoicing Task && Kanban List Filter For Invoiced Tasks

Civolilah:
@beganovich @turbo124 The PR includes two fixes: one for refetching the tasks query after the task is invoiced, and another for filtering the list in Kanban so that invoiced tasks are not displayed. Let me know your thoughts.


turbo124:
@beganovich for hotpatch please.

beganovich (approved):

Issue #1747, opened 2024-05-06
Product Cost not Accurately Displaying in Purchase Order


HiloKC:
Current Situation:
- Create New Product
- Add "Price" (Price Charged to Customer) eg: $1.00
- Add "Cost" (Amount Vendor Charges You for Product) eg: $0.7523
-Create New Purchase Order
- Input Vendor, Dates, etc.
- Click "+ Add Item"
- Select Existing Product (created earlier) from drop down
- Product Information is populated in Purchase Order 
- Unit Cost Displayed is not  product "Cost" (eg $0.75) but instead is the product "Price" (eg $1.00)

Can't see anywhere in issues or documentation where this specific issue has been addressed. 

The simple request here is for the purchase order to accurately pull up the product "Cost" so that PO with accurate cost can be sent to Vendor. The only workaround is to manually input the Cost in the the "Unit Cost" line (eg manually change cost from $1.001 to $0.7523). Is there a fix to ensure proper Unit Cost is displayed in the PO as well as ensuring cost can extend out 4 decimals rather than automatically round to 2 decimals (which causes large calculation errors in higher volume products)?

turbo124:
@Civolilah for purchase orders, if product cost > 0, then we should show item.product_cost instead of item.price

HiloKC:
Please note that now in the Invoice Ninja Mobile app, the product cost is correctly showing as the cost in new Purchase Orders created. However in the website (app.invoicing.co) the cost in new Purchase orders is still showing up incorrectly as the product price, not the cost.

Civolilah:
@HiloKC Thanks for your advices! The issue will be fixed. However, you also mentioned rounding the unit_cost to more than 2 decimals, which is the default. In the Settings > User Details > Preferences > Number precision, there is an input field where you can specify the number of decimal places you want to round the unit_cost column to.

HiloKC:
Thank you for the response!

Kevin S. Castle, PhD

Cofounder, SVP

Hilo Industries LLC

10432 Balls Ford Rd., Suite 300

Manassas, VA 20109

________________________________

PR #1751, opened 2024-05-07
[Bugfix] Displaying Product Cost Instead Of Product Price | Purchase Orders Line Items | #1747

Civolilah:
@beganovich @turbo124 The PR includes logic changes for displaying the value in the `unit_cost` column. So, if `product.cost` is greater than zero (0), then the `product.cost` will be displayed in the `unit_cost` column instead of `product.price`. This will only happen for Purchase Orders. Let me know your thoughts.


Civolilah:
@turbo124 I hope I understood the issue correctly and resolved it, but it seems that the same issue occurs on the Flutter web. Let me know your thoughts.

turbo124:
@Civolilah flutter handles this functionality correctly. for Purchase Orders only, we show the product cost rather than price.

Civolilah:
> @Civolilah flutter handles this functionality correctly. for Purchase Orders only, we show the product cost rather than price.

@turbo124 Ahamm, I adjusted that part of the logic. So currently, for the PO, if the `cost` is greater than zero, the `unit_cost` column will display the `cost` prop value instead of the `price` once you select the product. Please let me know if this is the correct behavior.

beganovich (approved):

PR #1753, opened 2024-05-08
[Bugfix] Endpoints Permissions | Create Expense Category & Create Task Status

Civolilah:
@beganovich @turbo124 The PR includes two fixes for the `/api/v1/task_statuses/create` endpoint and `/api/v1/expense_categories/create` endpoint. These changes ensure that only users with at least `ADMIN` permissions, can create these entities. The client experienced a toaster message indicating that the action was not authorized because they assigned view/edit/create permissions without setting the user to be an admin. This was the reason for the toaster issue. Let me know your thoughts.


turbo124:
@Civolilah just to note, this issue impacts _all_ routes where the user only have view permissions, the UI appears to always call the /create endpoint regardless of permission levels so this PR should be upgraded to cover all /create endpoints please

Civolilah:
> @Civolilah just to note, this issue impacts _all_ routes where the user only have view permissions, the UI appears to always call the /create endpoint regardless of permission levels so this PR should be upgraded to cover all /create endpoints please

@turbo124 I just checked all of them. Most of them were resolved when I was working on the permissions adjustment. However, three of them still missed the permission check, so I've just added it. So, basically, the queries that require create permission have a check, and also queries that require at least admin permission also have a check. Let me know your thoughts.

beganovich:
@turbo124 is this ready for merge?

beganovich (approved):

PR #1754, opened 2024-05-08
[Cleanup] Adding Max Date For Date Type Input Fields

Civolilah:
@beganovich @turbo124 The PR includes adding the `9999-12-31` date as the maximum date for date input fields to prevent invalid dates larger than this one, which are unrealistic. Let me know your thoughts.

### DIFF
--- a/src/components/forms/InputField.tsx
+++ b/src/components/forms/InputField.tsx
@@ -84,6 +84,7 @@ export function InputField(props: Props) {
             ...props.style,
           }}
           min={props.min}
+          max={props.type === 'date' ? '9999-12-31' : undefined}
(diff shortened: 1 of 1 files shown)


Civolilah:
@turbo124 Currently, we do not have any localization on input fields of type "date". I implemented it a long time ago, but I was on vacation, and you found some issues there. I'm not sure if we should plan on implementing localization on the date picker to align the format with what is entered in the Localization settings page. Also, I believe that implementing military_time will be necessary to format time in military or non-military time. Currently, all input fields of type date have the same format, regardless of what the user enters in the Localization settings. Let me know your thoughts.

beganovich (approved):

PR #1764, opened 2024-05-16
[Cleanup] Updating Country Selector Code To "iso_3166_3"

Civolilah:
@beganovich @turbo124 The PR includes updating the country code to "iso_3166_3". Screenshot:

Screenshot 2024-05-16 at 20 41 24

Let me know your thoughts.


Civolilah:
@turbo124 You also mentioned in the ticket that the country needs to be searchable by both name and code at the same time, but I see it is already implemented. Please let me know if you see anything different.

beganovich (approved):

PR #1767, opened 2024-05-19
[Cleanup]  Adjusting Column Width For Notes

Civolilah:
@beganovich @turbo124 The PR includes the notes column's width adjustment. Screenshot:

Screenshot 2024-05-20 at 19 22 08

Let me know your thoughts.


turbo124:
@beganovich ping

src/common/hooks/useExtractTextFromHTML.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Invoice Ninja (https://invoiceninja.com).
+ *
+ * @link https://github.com/invoiceninja/invoiceninja source repository
+ *
+ * @copyright Copyright (c) 2022. Invoice Ninja LLC (https://invoiceninja.com)
+ *
+ * @license https://www.elastic.co/licensing/elastic-license
+ */
+
+export function useExtractTextFromHTML() {
```
beganovich, line 11: Is there a reason why this is marked as a hook?

Seems like more work and more lines of code if this is marked as hook, but it has no access to React's lifecycle nor state.

src/common/hooks/useSanitizeHTML.ts
```diff
@@ -0,0 +1,19 @@
...
+ * @link https://github.com/invoiceninja/invoiceninja source repository
+ *
+ * @copyright Copyright (c) 2022. Invoice Ninja LLC (https://invoiceninja.com)
+ *
+ * @license https://www.elastic.co/licensing/elastic-license
+ */
+
+import DOMPurify from 'dompurify';
+
+export function useSanitizeHTML() {
```
beganovich, line 13: Same as previous one.

beganovich (commented):
Looks good, please just clarify the situation about hooks.

Civolilah:
> Looks good, please just clarify the situation about hooks.

@beganovich Yes, good catch. It can be converted to a clear function instead of hooks since we do not use any states in the logic. It has been converted, please let me know your thoughts.

PR #1775, opened 2024-05-23
[Feature] Implementing Links To Entities | Recurring Expenses

Civolilah:
In progress...

### DIFF
--- a/src/pages/recurring-expenses/components/Details.tsx
+++ b/src/pages/recurring-expenses/components/Details.tsx
@@ -23,11 +23,14 @@ import { VendorSelector } from '$app/components/vendors/VendorSelector';
 import { useTranslation } from 'react-i18next';
 import frequencies from '$app/common/constants/recurring-expense-frequency';
 import dayjs from 'dayjs';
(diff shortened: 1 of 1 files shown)


turbo124:
@Civolilah looks good, can we make these links open new tab instead of current tab please.

Civolilah:
> @Civolilah looks good, can we make these links open new tab instead of current tab please.

@turbo124 Sure thing, it is implemented that way, so the entity's page will open in a new tab. Let me know your thoughts.

turbo124:
@Civolilah i''m seeing a issue locally when clicking on this, it appears to lose the /#/ in the URL ? and does not work like other internal links in the application.

Civolilah:
> @Civolilah i''m seeing a issue locally when clicking on this, it appears to lose the /#/ in the URL ? and does not work like other internal links in the application.

@turbo124 Ah, okay. The issue was only with the `hash` router, but I tested with the `browser` only. I've fixed it. Let me know your thoughts.

beganovich (approved):

Issue #1744, opened 2024-05-04
Running task live timer does not use the real system clock, gets out of sync on time warps after suspend/resume


nekohayo:
In the v5 pro SaaS version, I will typically leave the task editor UI opened in a browser window (specifically, Epiphany's web app mode) while I'm meeting with a client, by clicking that "play" button:

![image](https://github.com/invoiceninja/ui/assets/479401/2e8fe55c-dfaf-413a-8d83-842cc5dc1ed1)

During that time, it is extremely likely that my computer (whether the desktop workstation in the office, or my laptop in the field, etc.) will either auto-suspend to RAM (after a certain amount of idle time, or on lid close), lose the internet connection when leaving wifi or losing a hotspot, etc.

The problem is, when I come back to it and wake+reconnect the computer, the webpage's live timer is apparently still running, but the time count is wrong, as it only incremented time while the page was actively displayed by the computer.

----

**Workaround:** Reloading the task's page shows the correct running time and then I can stop the live timer, but it would be nicer if the live timer was always correct even when there has been a system "time jump".

**Potential caveat I could imagine:** watch out for system timezone changes though, if for example someone timelogs their travel time across different countries on a sufficiently fast means of transportation ;)

turbo124:
@nekohayo thanks for reporting.

PR #1781, opened 2024-05-28
[Bugfix] Task Timer Offline Mode | #1744

Civolilah:
@beganovich @turbo124 The client reported an issue where the timer is not updated when they turn on the laptop from sleep mode, and that the time is only appropriately updated when the window is visible or focused. This issue is fixed now. I have tested this functionality as follows:

- Start timer
- Focus on a different tab
- Put your machine in sleep mode
- Turn on your machine and focus on the app's tab

After that, the time should be updated as if the tab was always visible. Please make sure you are testing this using the `https://staging.invoicing.co` API_URL. When you turn off your machine, the local server is also turned off, and once you stop the time, it is incorrectly calculated from the server's side.

Let me know your thoughts.


turbo124:
@beganovich are you able to test the functionality, i am not able to test this accurately.

turbo124:
ping @beganovich

PR #1789, opened 2024-05-31
[Bugfix] Changes Modal Initial Value Fix

Civolilah:
@beganovich @turbo124 The PR includes fixes for the changes modal, particularly for setting the initial value that needs to be compared with the current changes. Let me know your thoughts.

### DIFF
--- a/src/common/hooks/useAtomWithPrevent.ts
+++ b/src/common/hooks/useAtomWithPrevent.ts
@@ -47,7 +47,7 @@ export function useAtomWithPrevent(
 
   useEffect(() => {
     if (entity && currentInitialValue) {
-      setCurrentInitialValue(cloneDeep(entity));
+      setCurrentInitialValue(undefined);
     }
   }, [entity?.updated_at]);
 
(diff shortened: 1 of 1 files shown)


src/common/hooks/useAtomWithPrevent.ts
```diff
@@ -47,7 +47,7 @@ export function useAtomWithPrevent(
 
   useEffect(() => {
     if (entity && currentInitialValue) {
-      setCurrentInitialValue(cloneDeep(entity));
+      setCurrentInitialValue(undefined);
```
Civolilah, line 50: @beganovich Quick explanation for the change: setting the currentInitialValue to undefined will trigger the main useEffect responsible for setting the real value for this state.

```
useEffect(() => {
    if (entity && !currentInitialValue) {
      clearTimeout(initialValueTimeOut.current);

      const currentTimeout = setTimeout(
        () => setCurrentInitialValue(cloneDeep(entity)),
        350
      );

      initialValueTimeOut.current = currentTimeout;
    }
  }, [entity, currentInitialValue]);
```
  
In this case, it is important because in some entities, like invoices, there are calculations over the invoice object. Setting this to `undefined` will ensure that the second useEffect with a timeout guarantees that the initial value is always 100% correct.

The timeout will always await the latest change, even if some calculations change some props.

PR #1829, opened 2024-06-15
[Bugfix] Encoding CSV Files

Civolilah:
@beganovich @turbo124 The PR includes bug fixes for uploading CSV files with any different style of UTF-8. The logic created in this PR will take the text from the CSV file and then change all `";"` into `","` to ensure the API always receives consistent data. After that, we will convert all text into the UTF-8 standard and create a file from it. The modified file will be sent in the payload. Let me know your thoughts.


Civolilah:
@turbo124 I'm just wondering about one thing for the solution I made. So I replaced literally all the `";"` with `","` which means if the user potentially enters a value in a cell where they used `";"` character, which will also be replaced. I'm not sure there is a way to convert just the `";"` which separates the columns. Let me know your thoughts.

turbo124:
@Civolilah i think the most important part here is just the encoding, the API does manage the delimiters, so I don't think we need that implemented in front end.

Civolilah:
> @Civolilah i think the most important part here is just the encoding, the API does manage the delimiters, so I don't think we need that implemented in front end.

@turbo124 Okay, then the only applied logic here is for encoding style to always be UTF-8, so the API will never receive anything but the UTF-8 standard. Let me know if uploading works on your end the same as before.

beganovich (approved):

PR #1835, opened 2024-06-16
Disable TinyMCE auto link convert

beganovich:
This fixes an issue with relative URLs for the current window href.


beganovich:
cc @turbo124

PR #1841, opened 2024-06-17
[Feature] Implementing Additional Filters | Dashboard

Civolilah:
In progress...

### DIFF
--- a/src/components/forms/Button.tsx
+++ b/src/components/forms/Button.tsx
@@ -27,6 +27,7 @@ interface Props extends CommonProps {
   behavior?: 'button' | 'submit';
   disableWithoutIcon?: boolean;
   noBackgroundColor?: boolean;
+  minimalTypePadding?: boolean;
 }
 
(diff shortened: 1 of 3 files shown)


turbo124:
@Civolilah have we lost this one on jira, i cannot find tracking

Civolilah:
> @Civolilah have we lost this one on jira, i cannot find tracking

@turbo124 Actually, this PR was created as the start of the solution for additional filters on every table card on the Dashboard. We mostly discussed this privately on Slack. As I remember, you imagined it as additional buttons in the top right corner of every card, and by clicking on those buttons, the results on the card would be additionally filtered. The reason we actually stopped working on this one was specifically the e-invoices functionality at that point.

Civolilah:
@beganovich This PR is actually related to adding new buttons on the dashboard cards as additional filters. I started working here to create a kind of UI so once Dave is aligned with it he can start working on API implementation. However, this is stopped because we have more prioritized tasks.

PR #1842, opened 2024-06-18
[Cleanup] Removing Unnecessary Checks Under "Default" Component | Double Requesting

Civolilah:
@beganovich @turbo124 The PR includes removing checks under the "Default" component related to the `onSaveClick` prop to see if this causes an issue with double-requesting. So, only the necessary checks are left. Let me know your thoughts.


turbo124:
@beganovich i'll default to you on QA for this.

beganovich (approved):

PR #1849, opened 2024-06-19
[Bugfix] Improving UI/UX (Field Width) | Tasks

### LINKED ISSUE #1847: React Web App: Visual Fields Too Small
In the new React Web App the some visual fields are too small, here especually custom fields:

Old non-react web ui:
image

You can see that the custom field “Vorgang Kunde” is of an ok length and shows the field value correctly.

New React Web UI:
image

The field “Vorgang Kunde” is much smaller than in the old web interface and displays far less characters. With this important information is hidden which I need to see immediately when looking at the record.

Please change the layout so that more information is displayed and less space is used for nice UI elements. :-)

(For reference https://forum.invoiceninja.com/t/react-web-app-visual-fields-too-small/16313)

Civolilah:
@beganovich @turbo124 The client struggled with reading the details he entered, which should be visible without scrolling. To fix this, I adjusted the UI and made the fields significantly wider. Screenshot:

Screenshot 2024-06-19 at 21 46 32

Resolves #1847 

Let me know your thoughts.

### DIFF
--- a/src/pages/tasks/common/components/TaskDetails.tsx
+++ b/src/pages/tasks/common/components/TaskDetails.tsx
@@ -9,7 +9,7 @@
  */
 
 import { Card, Element } from '$app/components/cards';
-import { InputField, Link } from '$app/components/forms';
+import { InputField, InputLabel, Link } from '$app/components/forms';
 import { useCurrentCompany } from '$app/common/hooks/useCurrentCompany';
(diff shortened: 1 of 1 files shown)


turbo124:
Looks good, thanks!

beganovich:
Superseded with #1864.

PR #1858, opened 2024-06-25
[Bugfix] "Per Page" Query Parameter | Dashboard Tables

Civolilah:
@beganovich @turbo124 The PR includes setting the "per_page" parameter through the endpoint to a fixed value of 50, instead of storing it in the preferences where it was always 10. Let me know your thoughts.


turbo124:
@beganovich => main please.

beganovich (approved):

PR #1864, opened 2024-06-25
Group tasks into two groups

beganovich:
This groups tasks into two groups.

image


turbo124:
image

Looks good, lets make columns fill width thou?

PR #1872, opened 2024-07-01
[Cleanup] Prevent Page Leaving Reimplementation

Civolilah:
@beganovich @turbo124 The PR includes the reimplementation of the "prevent page leaving" functionality from the Slack discussion. Let me know your thoughts.


turbo124:
@Civolilah were you able to find the issue here? just ping when we should attempt another review.

Civolilah:
> @Civolilah were you able to find the issue here? just ping when we should attempt another review.

@turbo124 I found the edge case issue and then refactored the solution to fix it. Please let me know if you are seeing the correct behavior in all cases now.

turbo124:
@Civolilah I think we are close.

One set of edge cases to resolve.

Need to be careful with the implementation of this when a user uses a More Actions command.

For example, if the user clicks Mark Sent. Then the change modal should not be displayed.
Alternatively, when clicking on Enter Payment, you would need to present the modal for any changes that are yet to be saved.

Civolilah:
@turbo124 I just merged the develop branch into this one to remove the tiny, unnecessary logic for the `back` button that we no longer have. I finished it, and the PR is ready for final review. Let me know your thoughts.

turbo124:
@beganovich this looks good from my testing. We'll need a way to disable this if unexpected edge case appear.

Civolilah:
> @beganovich this looks good from my testing. We'll need a way to disable this if unexpected edge case appear.

@turbo124 @beganovich I just implemented the disabling of this functionality if something goes wrong. It will be controlled through the .env variable `VITE_DISABLE_PREVENT_NAVIGATION_FEATURE`. If this is set to `true`, the functionality will not be available.

beganovich (approved):

beganovich:
@turbo124 feel free to merge after confirmation that this feature is _disabled_ by default (VITE_DISABLE_PREVENT_NAVIGATION_FEATURE=true).

beganovich (approved):

PR #1878, opened 2024-07-04
[Feature] Implementing "Referrals" Column | Users

Civolilah:
@beganovich @turbo124 The PR includes the implementation of displaying additional details for "Referrals". Screenshot:

Screenshot 2024-07-08 at 01 34 07

`Note`: We are missing the "referrals" translation keyword from the files, so if you agree, please just add it there.

Let me know your thoughts.

### DIFF
--- a/src/common/interfaces/user.ts
+++ b/src/common/interfaces/user.ts
@@ -11,6 +11,12 @@
 import { CompanyUser } from './company-user';
 import { Timestamps } from './timestamps';
 
+interface ReferralMeta {
+  pro: number;
+  free: number;
+  enterprise: number;
+}
+
 export interface User extends Timestamps {
   id: string;
   custom_value1: string;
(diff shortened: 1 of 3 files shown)


turbo124:
For this partial item, thimage

We only need to display the referral data for the current logged in user.

This should be included in the same location at the referral program URL.

Civolilah:
> For this partial item, thimage
> 
> We only need to display the referral data for the current logged in user.
> 
> This should be included in the same location at the referral program URL.

@turbo124 Ah, okay, my bad. Actually, I'm not able to see the `referral_meta` object in the currently logged-in user object, but I guess it is just not live yet? Screenshot:

Screenshot 2024-07-08 at 01 34 07

Let me know your thoughts.

turbo124:
I've tagged new releases, so this will be available for testing in v5.10.5

Civolilah:
> I've tagged new releases, so this will be available for testing in v5.10.5

@turbo124 Okay, great!

beganovich (approved):

PR #1886, opened 2024-07-10
[Cleanup] Removing Facebook Pixel Script From Login And Register Pages

Civolilah:
@beganovich @turbo124 The PR includes improving the logic for importing Facebook scripts. These scripts will be injected on all pages except for the `login` and `register` pages. This improvement has been achieved using the `react-helmet `package. Let me know your thoughts.


beganovich:
Please don't use Helmet. It's been abandoned for a while and will block our upgrade path to React 19.

turbo124:
this task is to remove it from all pages, no need for granularity, delete all references.

Civolilah:
> this task is to remove it from all pages, no need for granularity, delete all references.

@turbo124 Okay, so I've removed the scripts that inject Facebook scripts. They have been removed across the entire app.

beganovich (approved):

PR #1893, opened 2024-07-11
[Cleanup] Nordigen Reconnecting One Time Token Generation

Civolilah:
@beganovich @turbo124 The PR includes adding an additional prop (`institution_id`) to the payload while generating the `one_time_token` for the `Nordigen` reconnection when the upstream is `DISABLED`. Screenshot:

Screenshot 2024-07-11 at 17 16 17

Let me know your thoughts.

### DIFF
--- a/src/common/interfaces/bank-accounts.ts
+++ b/src/common/interfaces/bank-accounts.ts
@@ -28,4 +28,5 @@ export interface BankAccount {
   provider_name: string;
   updated_at: number;
   integration_type: string;
+  nordigen_institution_id: string;
 }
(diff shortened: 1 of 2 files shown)


Civolilah:
@turbo124 We already had an implementation for this purpose, so the same red triangle was displayed with the tooltip. Once the user clicks on the triangle, the `one_time_token` endpoint is called with the `context: 'nordigen'` and `platform: 'react'` in the payload. Now, I have added the `institution_id` set to the value from the `nordigen_institution_id` bank integration property. After receiving the `hash`, we open a new tab navigating to the `'/nordigen/connect/:hash'` page.

Please let me know if adding `institution_id` to the payload for this token is enough or if we need to do some additional work.

beganovich (approved):

PR #1897, opened 2024-07-12
Auto expanding description inputs for products table

beganovich:
This auto expands the description note for the products table. Also collapses on click and goes back on focus lost.


turbo124:
@beganovich looks great!

can we add this as an option within react_settings please, configurable in Settings > User Details > Preferences

PR #1899, opened 2024-07-13
Fixes for apply payments with refunded amounts

turbo124:
Implements fixes for the way in which apply payments are displayed when refunds have been made on a payment with remaining unapplied balance.


turbo124:
patch to main please.

PR #1917, opened 2024-07-18
[Cleanup] Making Currency Selector Dismissable && Fixing Recurring Expense Currency Converter

Civolilah:
@beganovich @turbo124 The PR includes adding dismiss functionality for the currency selector on the expense and recurring expense pages, along with minor fixes for the convert currency section logic for both entities. Screenshot:

Screenshot 2024-07-19 at 20 44 11

Let me know your thoughts.

### DIFF
--- a/src/pages/expenses/create/components/AdditionalInfo.tsx
+++ b/src/pages/expenses/create/components/AdditionalInfo.tsx
@@ -21,17 +21,18 @@ import dayjs from 'dayjs';
 import { useResolveCurrencySeparator } from '$app/pages/transactions/common/hooks/useResolveCurrencySeparator';
 import { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
-import { ExpenseCardProps } from './Details';
 import { useReactSettings } from '$app/common/hooks/useReactSettings';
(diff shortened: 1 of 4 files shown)


Civolilah:
@turbo124 I've checked production and the convert_currency toggle does nothing. I've fixed it here along with improving the dismissing part. I made the functionality exactly the same as on the expense pages. It might be great to push this into production after merging. Let me know your thoughts.

turbo124:
@Civolilah still WIP? i've tested this and looks ok?

Civolilah:
> @Civolilah still WIP? i've tested this and looks ok?

@turbo124 Yes, I noticed one more tiny bug through additional testing, so I moved it to the draft to check it out and eventually fix it. Let me check that one more time.

Civolilah:
@turbo124 In one or two cases, the logic for converting currency was relying on the main currency selected for the entity (which is not the case for the Flutter app). This logic caused an edge case bug that did not update the `foreign_amount` correctly. I fixed it for both entities. I've checked production, and we have the same issue there. Please let me know if it works correctly on your end as well. If so, it might be great to push it to production since we have a bug there. What do you think?

beganovich (approved):

PR #1924, opened 2024-07-24
[Bugfix] Label When Self-hosted Plan Is Paid

### LINKED ISSUE #1844: Wrong text for Whitelabel
Hi,
I recently purchased a whitelabel license for our instance, but the React interface still show's as "Free", while Flutter shows as "Licensed". I'm running the latest version available, as of now.
image

Civolilah:
@beganovich @turbo124 The PR includes adjusting the label on the `account_management` page for self-hosted users with a paid plan. The label will use the `"licensed"` translation keyword when `account.plan_expires` is not empty, or in other words, when `plan_expires` is populated.

`Note`: We are missing the "licensed" translation keyword from the files. If you agree with this addition, please add it to the files.

Let me know your thoughts.

Closes #1844

### DIFF
--- a/src/pages/settings/account-management/component/Plan.tsx
+++ b/src/pages/settings/account-management/component/Plan.tsx
@@ -41,7 +41,8 @@ export function Plan() {
         ) : (
           <span>
             {t(
-              account?.plan_expires !== ''
+              account?.plan_expires !== '' &&
+                !dayjs(account.plan_expires).isBefore(dayjs())
                 ? 'licensed'
                 : 'plan_free_self_hosted'
             )}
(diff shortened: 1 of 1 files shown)


turbo124:
@Civolilah populated won't be enough, you will need to also parse and ensure plan_expires >= now

Civolilah:
> @Civolilah populated won't be enough, you will need to also parse and ensure plan_expires >= now

@turbo124 Sure, I've added a comparison to the current date to ensure that the `plan_expires` date is the `same` as or `greater` than the current date. Let me know your thoughts.

beganovich (approved):

PR #1939, opened 2024-07-29
[Feature] Default Quantity While Creating Line Items

### LINKED ISSUE #1843: React web app in new version 5.9.4 always uses default quantity of zero
Hi!

We recently upgraded our self-hosted installation to v5.9.4. and I wanted to test the react web app.

In the product settings we usually disable "Show Product Quantity" and set the default quantity to one. (We don’t deal with product quantities and we don’t show them on the invoices).
This works fine in the old flutter web interface but when I switch to the new react web app, it still shows the product quantity when creating a new invoice and this quantity is always set to zero.

How can we get rid of product quantities again?
Or at least have a default value of always one?

Civolilah:
@beganovich @turbo124 The PR includes an improvement in setting the default quantity when creating a line item. Before this change, we had the default quantity set to 0 after creating the line item, but now it is always 1. Let me know your thoughts.

Closes #1843

### DIFF
--- a/src/pages/credits/common/hooks.tsx
+++ b/src/pages/credits/common/hooks.tsx
@@ -167,7 +167,11 @@ export function useCreditUtilities(props: CreditUtilitiesProps) {
           ...current,
           line_items: [
             ...current.line_items,
-            { ...blankLineItem(), type_id: InvoiceItemType.Product },
+            {
+              ...blankLineItem(),
+              type_id: InvoiceItemType.Product,
+              quantity: 1,
+            },
           ],
         }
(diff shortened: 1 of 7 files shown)


Civolilah:
@turbo124 The issue with the `unit_cost` column and non-updating PDF preview has been fixed, and it should now work as expected. Let me know your thoughts.

turbo124:
@beganovich for patch to main as a priority please.

beganovich (approved):

beganovich (approved):

PR #1940, opened 2024-07-29
POC Websockets.

turbo124:


beganovich:
POC looks good. It'd be better if we could make slight changes in passing data down thru React Context or Jotai atom in case we need to access the connection somewhere else or listen for an event.

src/index.tsx
```diff
@@ -33,7 +33,7 @@ import editorWorker from 'monaco-editor/esm/vs/editor/editor.worker?worker';
 import jsonWorker from 'monaco-editor/esm/vs/language/json/json.worker?worker';
 import cssWorker from 'monaco-editor/esm/vs/language/css/css.worker?worker';
 import htmlWorker from 'monaco-editor/esm/vs/language/html/html.worker?worker';
-import tsWorker from 'monaco-editor/esm/vs/language/typescript/ts.worker?worker';
+import tsWorker from 'monaco-editor/esm/vs/language/typescript/ts.worker?worker';2
```
beganovich, line 36: Probably misclick?

beganovich (commented):
Tiny clarification.

turbo124:
@beganovich 

agreed, this is just the working POC. we need to think about how to integrate this now

PR #1944, opened 2024-07-31
[Bugfix] Replaced Wrong Translation Function

Civolilah:
@beganovich @turbo124 The PR includes replacing the incorrect translation function. The client noticed this issue when adding tasks to an invoice, but upon investigation, I found a few more instances. All of these have been fixed. Let me know your thoughts.


turbo124:
@beganovich for merge to main as a priority please.

beganovich (approved):

PR #1957, opened 2024-08-06
[Bugfix] More Descriptive Message On Modal | Health Check | Gateways

Civolilah:
@beganovich @turbo124 The PR includes changing the approach to inform users about payment_gateway status issues when they click the `health_check` button. I caught the error message from the response and displayed it in a modal.

The client reported an issue where clicking the save button causes the API to return a 404 error, navigating them to a 404 page. We don't manually navigate to this page except when the API returns a 404. David, please let us know in which cases this endpoint returns a 404, this information might help us resolve the issue if one exists on the React side.

The client also mentioned not receiving a descriptive message on the health_check button (I guess he just receives the "Failed" message). That's why I included the message from the API's response, hoping it might provide more information about what's wrong.

The only case I can recreate is when I don't enter the api_key prop, which results in a message on the health_check indicating it needs to be filled. Screenshot:

Screenshot 2024-08-06 at 15 00 02

Let me know your thoughts.


turbo124:
@Civolilah we'll want to handle another edge case here, i don't think we should allow the use to click on health check unless the gateway is saved. if i create a new gateway and then click health check, i get 404

Civolilah:
> @Civolilah we'll want to handle another edge case here, i don't think we should allow the use to click on health check unless the gateway is saved. if i create a new gateway and then click health check, i get 404

@turbo124 Sure, so I implemented this logic: If you are on the /create page, you won't even have this button. However, if you are on the /edit page, you will have the button, but it will be disabled if you change anything from the initial fetch. Once you save the details, the button will be enabled again. Let me know your thoughts.

turbo124:
Thanks, this looks good. 

One thing I have just noticed on this page is that we only show the gateway name on the first page.

image

It would be ideal to have the gateway name in the breadcrumb, but if not possible, then in the heading of the card it would be second best

image

Civolilah:
> Thanks, this looks good.
> 
> One thing I have just noticed on this page is that we only show the gateway name on the first page.
> 
> image
> 
> It would be ideal to have the gateway name in the breadcrumb, but if not possible, then in the heading of the card it would be second best
> 
> image

@turbo124 Sure thing, I implemented it through the breadcrumbs. Screenshot:

Screenshot 2024-08-08 at 10 57 00

Let me know your thoughts.

beganovich (approved):

PR #1961, opened 2024-08-07
[Bugfix] Displaying Activities On Invoice Edit Page

Civolilah:
@beganovich @turbo124 The PR includes two fixes. First, we now use the same function for displaying activities on both the invoice edit page and the invoice slider. Second, I've added the resolution of two additional variables that we missed: `payment` and `payment_amount`. Screenshot:

Screenshot 2024-08-07 at 11 42 34

Let me know your thoughts.


turbo124:
@beganovich => main please.

beganovich (approved):

PR #1972, opened 2024-08-12
[Bugfix] Disabling Navigation For Activities

Civolilah:
@beganovich @turbo124 The PR includes a fix for activities with no history. Previously, when navigating to the preview page of an activity with no history, the API returned a 404 error and the user was redirected to the 404 page. I resolved this by disabling navigation if the history object does not have a populated `id` property (the activity element will not be clickable). Let me know your thoughts.


Civolilah:
@turbo124 The bug is also in production, so it might be great to merge this one into the `main` branch.

turbo124:
@beganovich please merge to main

beganovich (approved):

Issue #1966, opened 2024-08-11
Task details editor's "running" timer is visually unclear for the current time entry at the bottom of the list


nekohayo:
Pro SaaS v5's running timer UI, compared to what I remember from v4, is less clear when it comes to the currently "running" portion of a task. Two issues:

* It doesn't differentiate between the _total_ accumulated time for that task vs the _current_ timelog entry's running time; yet, most of the time, with tasks that can span weeks and months, I want to know how much time I've been spending on it _today_ and whether my timer makes sense or if I forgot to turn it off at some point.
* It doesn't help me quickly see that the last entry in the list is "running"

Currently it looks like this:

![task edit details running timer UI - current](https://github.com/user-attachments/assets/89e64a7a-ba1f-462f-8ab1-97fbbfe4af12)

Here is a proposed way to make this much clearer, within the list of timelog entries (notice how it also shows the timelog's timer count as a different number label than the total one at the top):

![task edit details running timer UI - mockup](https://github.com/user-attachments/assets/b47d81d5-4f2e-4c6f-a834-82eb07ad7b2e)

Pausing it would make the fields reappear, with the correctly saved values.

Ideally, there should be a "resume" button for it, allowing me to remove the end time and resume it "as if there had been no interruption" (I frequently stop tasks only to resume them a few minutes later because it turned out I was not actually done with it, so I want to be able to resume "without gap").

PR #1975, opened 2024-08-12
Indicator that task is running in tasks table

beganovich:
Ref. #1966.

image


turbo124:
need to ensure we don't have multiple timers running in the UI

turbo124:
@beganovich looks good!

PR #1979, opened 2024-08-14
[Bugfix] Documents Permissions

Civolilah:
In progress...


turbo124:
@Civolilah look good,

@beganovich => main please

beganovich (approved):

beganovich:
Unfortunately we cannot merge this one to production as is, since it targets some components that aren't in the main, yet (credits, quotes, recurring invoices).

@Civolilah please send PR that target document updates only for components that *are* in main.

Issue #1986, opened 2024-07-15
I would assume after I 'invoice project' from a specific project that it would auto-populate the invoice with this project.


theoryshaw:
I would assume after I 'invoice project' from a specific project that it would auto-populate the invoice with this project.
Perhaps i'm missing something. 

https://github.com/user-attachments/assets/fa349be4-413a-4746-b22f-330b63c2b4d8

turbo124:
@theoryshaw no, the client should be populated when the project is created. My guess is that the client record is not available at this point @beganovich @Civolilah please advise.

turbo124:
@theoryshaw one thing that may impact this is if the task inside the project has already been invoiced? you would not be able to re-invoice the project?

PR #1989, opened 2024-08-16
Fixes for empty client on invoicing the project

beganovich:
Ref. #1986.


turbo124:
@beganovich 

Thanks this looks good.

PR #1992, opened 2024-08-20
[Bugfix] Covering Edge Case For "Custom" Date Range | Dashboard Chart

Civolilah:
@beganovich @turbo124 The PR includes fixes for two minor issues on generating date ranges for the chart on the dashboard. The first one is related to the `last_30_days` date range, and the logic is improved to cover the edge case there. The second issue is more about improving UX when selecting a `custom` date range, and it has also been improved. I tested this very thoroughly, and it seems to me that it works perfectly fine. The issue was `ONLY` related to when `"MONTH"` period is selected on the left picker side. Let me know your thoughts.


Civolilah:
@turbo124 I created many more cases using the `custom` picker rather than dropdown fixed date range periods. So fixed date range periods are always correct and should remain that way. Testing very edge cases can be done through custom date ranges.

beganovich (approved):

PR #1997, opened 2024-08-22
[Bugfix] Recalculating Task Timer On Saving

### LINKED ISSUE #1965: Task details editor's "running" timer does not correct itself when changing the currently running task's start time
If you edit a task to start its running timer, and later in the day want to correct the start time for it (either to make it start earlier, or to make it start later because you got sidetracked or delayed by something else), editing the field and clicking "Save" does not correct the running timer.

For example, if the running timer is now at 4.5 hours, and you bump today's start time 2 hours later, the timer remains at 4.5 hours+ after saving the changes. Only a full page reload corrects it.

Tested with the v5 pro SaaS version on Firefox 128.

Civolilah:
@beganovich @turbo124 The PR includes an improvement to the task timer. It includes recalculation of the timer when, for example, the start time is changed, and when you save the task, the time will be updated/recalculated. Let me know your thoughts.

Closes #1965


Civolilah:
@turbo124 I believe that we can go a step further on this one and improve the UX more. I can recalculate the timer on any changes of the running time log, which is I believe better than on saving, if you would agree with me.

turbo124:
@Civolilah i am still seeing the 404 on this PR, should we change this to draft?

Civolilah:
> @Civolilah i am still seeing the 404 on this PR, should we change this to draft?

@turbo124 The Ben's PR, which resolves that issue, was not merged into develop until today. Right now we have those fixes on this PR also, so you should not experience that issue anymore.

turbo124:
@Civolilah looks good

beganovich (approved):

PR #2002, opened 2024-08-23
[Bugfix] Handling 422 On Reports Page

### LINKED ISSUE #1970: Reports time range dates fields don't interpret what I mean, and reports export button breaks forever after error message
Tested with the Pro v5 SaaS version, two issues in one here:

1. It does not "parse what I mean" when I simply type the month in the field. It already knows what I mean, after all it had that placeholder text in the background of the field! Why can't it understand that I meant to change only the month, and that if I left the year (or some other part of the date) incomplete then it could assume this year? (same for "this month" etc.)

2. As soon as you hit that error message, it's Game Over, the "Export" button will never work again until you reload the page and start over

https://github.com/user-attachments/assets/9491745c-990b-4ab7-95ad-a9b36dc8368f

Civolilah:
@beganovich @turbo124 The PR includes fixes for handling 422 errors when exporting reports. Screenshot:

Screenshot 2024-08-24 at 17 51 15

Let me know your thoughts.

Closes #1970

### DIFF
--- a/src/pages/reports/index/Reports.tsx
+++ b/src/pages/reports/index/Reports.tsx
@@ -239,14 +239,12 @@ export default function Reports() {
             setIsPendingExport(false);
           });
       })
-      .catch((error: AxiosError<ValidationBag | Blob>) => {
-        setIsPendingExport(false);
-
+      .catch((error: AxiosError<ValidationBag>) => {
         if (error.response?.status === 422) {
-          console.log(error);
-
           setErrors(error.response.data as ValidationBag);
         }
+
+        setIsPendingExport(false);
       })
       .finally(() => {
         if (showCustomColumns) {
--- a/src/pages/reports/index/Reports.tsx
+++ b/src/pages/reports/index/Reports.tsx
@@ -240,16 +240,12 @@ export default function Reports() {
           });
       })
       .catch((error: AxiosError<ValidationBag | Blob>) => {
+        setIsPendingExport(false);
+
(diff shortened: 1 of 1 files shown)


src/pages/reports/index/Reports.tsx
```diff
@@ -239,18 +239,12 @@ export default function Reports() {
...
+      .catch((error: AxiosError<ValidationBag>) => {
         if (error.response?.status === 422) {
-          if (report.payload.send_email) {
-            setErrors(error.response.data as ValidationBag);
-          }
-
-          if (!report.payload.send_email) {
-            const blob = error.response.data as Blob;
-
-            blob.text().then((text) => setErrors(JSON.parse(text)));
```
Civolilah, line 251: @beganovich I'm not sure I can find a case where resolving the blob is needed and then putting it into an errors state. Whatever I did, I got the standard error object from the API, and this handling causes an issue with not displaying the message. Please let me know if there is a case where this is still needed.

beganovich: I am sure it was set with a reason. If it still works without, then we good.

Civolilah: @beganovich I just retested it, and the behavior is exactly the same. As I can see here, error handling is separated based on the `send_email` boolean value, which I tested, and the response from the API is not different, so it does not require different handling. I believe that this approach is correct as a solution.

beganovich (approved):

PR #2009, opened 2024-08-25
[Bugfix] Nested Modals Focusing Field Issue

Civolilah:
@beganovich @turbo124 The PR includes fixes for focusing the input field on the password confirmation modal in the about_modal. The only way to fix this issue was by installing version `2.0.0-alpha.4` of the `@headlessui/react` package, where they fixed this issue as mentioned here (`https://github.com/tailwindlabs/headlessui/discussions/2983`). After installing it and setting the `initialFocus` prop, we no longer have this issue. I tested some other more complex modals, and installing this new version did NOT change anything, but we will need to pay ATTENTION to modals in further use of the app to see if there are any edge cases that come with installing the new package version. Let me know your thoughts.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "dependencies": {
         "@azure/msal-browser": "^3.0.1",
         "@emotion/styled": "^11.10.5",
-        "@headlessui/react": "^1.7.4",
+        "@headlessui/react": "^2.0.0-alpha.4",
         "@headlessui/tailwindcss": "^0.1.3",
         "@hello-pangea/dnd": "^16.2.0",
         "@monaco-editor/react": "^4.4.6",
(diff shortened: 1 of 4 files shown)


turbo124:
@beganovich thoughts on this approach?

Another approach we may want to consider is never presenting two modals on the same page?

Civolilah:
> @beganovich thoughts on this approach?
> 
> Another approach we may want to consider is never presenting two modals on the same page?

@turbo124 Yes, another approach is to never display nested modals when the second, third, etc. have a field that has to be focused.

I think that there is probably another approach, but I believe that one breaks the UX a little bit. If we open a second modal, we can try automatically closing the first one so there's always only one modal open on the page. I didn't try this because it seemed like bad UX from my perspective, but from a technical perspective, maybe that can solve the issue.

beganovich:
If a package upgrade fixes the bug, I think that's the way to go.

Never allowing two modals is also an option, gotta make sure we already don't have UI that does this.

beganovich (approved):

PR #2012, opened 2024-08-27
[Bugfix] Invoicing Expense | Query Invalidation

Civolilah:
@beganovich @turbo124 The PR includes fixes for query invalidation when saving the invoice. We did not have correctly invalidated the /expenses route, so the table was not updating. I have added query invalidation for expenses once the invoice is saved. Let me know your thoughts.


Civolilah:
@turbo124 This is a production bug, so I believe it would be great to push this fix to main.

beganovich (approved):

PR #2033, opened 2024-09-07
Simplify Statics interface

ttytm:
The PR removes redundancy. Applying the same principles already used for `Country` and `Currency` typed fields in the `Statics`

### DIFF
--- a/src/common/interfaces/statics.ts
+++ b/src/common/interfaces/statics.ts
@@ -10,13 +10,14 @@
 
 import { Country } from './country';
 import { Currency } from './currency';
+import { DateFormat } from './date-format';
 
 export interface Statics {
   banks: Bank[];
   countries: Country[];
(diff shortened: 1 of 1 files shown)


src/common/interfaces/statics.ts
```diff
@@ -35,13 +36,6 @@ export interface Bank {
   config: string;
 }
 
-export interface DateeFormat {
-  id: string;
-  format: string;
-  format_moment: string;
-  format_dart: string;
-}
-
```
ttytm, line 44: Quick reference to the updated type

https://github.com/invoiceninja/ui/blob/fbb22e5363feabd1bb28154bf5111ddb951d7766/src/common/interfaces/date-format.ts#L11-L16

beganovich (approved):
Thanks :+1:

PR #2048, opened 2024-09-13
[Bugfix] Currency Decimal And Thousand Operators Adjustments

Civolilah:
@beganovich @turbo124 The PR includes the implementation of the `use_comma_as_decimal_place` property field in the Localization page, replacing number-type input fields from the `react-debounce-input` package with fields from the `react-number-format` package to achieve comma separators and allow setting precision for them, as specified in reactSettings. Let me know your thoughts.


src/components/forms/InputField.tsx
```diff
@@ -78,86 +106,129 @@ export function InputField(props: Props) {
...
-          onChange={(event) => {
-            if (
-              props.element === 'textarea' &&
-              reactSettings.preferences.auto_expand_product_table_notes
-            ) {
-              const scrollHeight = event.target.scrollHeight + 2;
-
-              if (scrollHeight < 200) {
-                event.target.style.height = scrollHeight + 'px';
+        {props.type === 'number' ? (
```
beganovich, line 109: This is getting a bit noisy. Is it an easy win to split the components? Eg. NumberInputField?

Civolilah: > This is getting a bit noisy. Is it an easy win to split the components? Eg. NumberInputField?

@beganovich Sure thing, I also think it's best to change it now and make maintenance easier for us. I've made the necessary changes, and you can take a look to see if you agree with them. However, I'll need some time to test these significant changes in our app to avoid any edge case bugs. I'll return the PR to "ready for review" once I finish the full testing.

beganovich (approved):

beganovich:
@turbo124 looks good to me. if you have time, give it another QA pass :+1:

PR #2054, opened 2024-09-18
[Bugfix] Displaying Modal With Message When Invoice Is Locked

Civolilah:
@beganovich @turbo124 The PR includes the implementation of improved UX when different lock_invoices options are selected. I made the logic which will check if the invoice should be locked and notify the user on the invoice edit page if the invoice is already locked, not allowing the user to change anything there, with a message on a modal explaining why the invoice is locked. Screenshot:

Screenshot 2024-09-18 at 22 58 19

Let me know your thoughts.


Civolilah:
@turbo124 As you recommended on the Jira ticket, the implementation has been slightly changed with an `is_locked` query parameter on the invoice edit, along with the banner implementation if the invoice is locked. Screenshot:

Screenshot 2024-09-22 at 23 56 01

Let me know your thoughts.

beganovich (approved):

PR #2055, opened 2024-09-18
[Cleanup] Replacing Dropdowns With Searchable Dropdowns

Civolilah:
@beganovich @turbo124 The PR includes changing three dropdowns (country selector, US state selector, and EU country selector) into our searchable dropdowns. Screenshot:

Screenshot 2024-09-18 at 23 38 01

Let me know your thoughts.

### DIFF
--- a/src/components/EUCountrySelector.tsx
+++ b/src/components/EUCountrySelector.tsx
@@ -8,23 +8,25 @@
  * @license https://www.elastic.co/licensing/elastic-license
  */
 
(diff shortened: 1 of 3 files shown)


Civolilah:
@turbo124 The title of the Jira ticket states that we should change all dropdowns to searchable ones, but also there are two subtasks specifically for those three dropdowns. I have currently changed those three, please let me know if I should change any other dropdowns.

beganovich (approved):

PR #2072, opened 2024-09-25
[Feature] Inbound Mailbox Fields Implementation | Expense Settings

Civolilah:
In progress...

### DIFF
--- a/src/common/interfaces/company.interface.ts
+++ b/src/common/interfaces/company.interface.ts
@@ -79,6 +79,15 @@ export interface Company {
   smtp_verify_peer: boolean;
   e_invoice?: EInvoiceType;
   use_comma_as_decimal_place?: boolean;
+  notify_vendor_when_paid?: boolean;
+  expense_mailbox_active?: boolean;
+  inbound_mailbox_allow_company_users?: boolean;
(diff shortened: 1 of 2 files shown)


Civolilah:
@turbo124 Currently, fields are only available for company-level settings and self-hosted users. Please let me know if I'm incorrect.

turbo124:
@Civolilah 

Looks good.

To confirm:

1. Self hosted only.
2. expense_mailbox_active should be the first setting in the list with a line break starting this section
3. only display the additional properties IF expense_mailbox_active = true

Civolilah:
> @Civolilah
> 
> Looks good.
> 
> To confirm:
> 
> 1. Self hosted only.
> 2. expense_mailbox_active should be the first setting in the list with a line break starting this section
> 3. only display the additional properties IF expense_mailbox_active = true

@turbo124 Okay, so the functionality is available only for COMPANY level settings for the SELF-HOSTED users. Also, a divider has been added to separate this section, and other fields will be available once the `expense_mailbox_active` field has been set to true, which is the first field in this section. Let me know your thoughts.

beganovich (approved):

PR #2074, opened 2024-09-27
[Feature] Tooltip On Quantity Field For Line Items On Invoices And Quotes

Civolilah:
In progress...


turbo124:
@Civolilah I've just realised this one is redundant as we display stock quantity in the product itself 

image

PR #2080, opened 2024-09-30
Broadcasting: Banner for "invoice updated" event

beganovich:
This adds a banner informing users about changes in the background of invoices.

image


turbo124:
@beganovich we may need a killswitch on this one

beganovich:
@turbo124 do we need killswitch on the backend or only in react?

turbo124:
@beganovich that is a good point. we can just disable the broadcast_driver on backend.

PR #2105, opened 2024-10-07
[Bugfix] Fixing Edge Case Issue With Precision Of Numbers And Adding Fallback Value

Civolilah:
@beganovich @turbo124 The PR includes fixing an edge case bug of setting default precision when converting numbers where the precision has been disabled (we do not have a specific number, precision is equal to the user's input). Also, I've added a fallback value of an empty string for fields that missed it and disabled precision and separators for fields that do not need them. Screenshot:

Screenshot 2024-10-07 at 20 24 44

Let me know your thoughts.


turbo124:
=> main please.

beganovich (approved):

PR #2106, opened 2024-10-07
[Cleanup] Changing Selectors For Payments And Quotes

Civolilah:
@beganovich @turbo124 The PR includes changing current selectors with new ones for Payments and Quotes. Screenshot:

Screenshot 2024-10-07 at 21 52 31

Let me know your thoughts.


turbo124:
@Civolilah should i be able to typeahead with these? Or are they only pure selects?

Civolilah:
> @Civolilah should i be able to typeahead with these? Or are they only pure selects?

@turbo124 These are pure selects, because we've already made searchable selects with the same design for selectors where we thought it would be great to have a search option. If you want to make all of them with a search option, it's not a problem; it's literally just passing one prop. Let me know your thoughts.

turbo124:
@Civolilah search option is best of both worlds, please change to that.

Civolilah:
> @Civolilah search option is best of both worlds, please change to that.

@turbo124 Okay, so right now those fields are also made searchable. Please let me know your thoughts.

turbo124:
image

can we ensure the x is only present when there is a value to be cleared?

Civolilah:
> image
> 
> can we ensure the x is only present when there is a value to be cleared?

@turbo124 Okay, sure, so the clear button is only available if there is a selected value. Screenshot:

Screenshot 2024-10-20 at 18 34 56

Let me know your thoughts.

beganovich (approved):

PR #2112, opened 2024-10-08
GoCardless OAuth2: Scope testing to company

beganovich:
To enable this in production, we need to pass `VITE_GOCARDLESS_OAUTH_TESTING_COMPANY` on the build. Ensure the value is `hashed_id` since we compare the company ID with the one from the API response.


turbo124:
this has passed QA, unsure what is needed from here?

beganovich:
this is old PR, fn has already been merged :+1:

PR #2114, opened 2024-10-09
[Bugfix] Import Template Modal Nested Modals Issue

Civolilah:
@beganovich @turbo124 The PR includes fixes for an issue with nested modals when saving import templates. Let me know your thoughts.

### DIFF
--- a/src/components/import/ImportTemplateModal.tsx
+++ b/src/components/import/ImportTemplateModal.tsx
@@ -52,6 +52,7 @@ export function ImportTemplateModal(props: Props) {
   const handleOnClose = () => {
     setTemplateName('');
     setIsSaveTemplateModalOpen(false);
+    navigate(`/${entity}s`);
   };
 
   const shouldOpenTemplateModal = () => {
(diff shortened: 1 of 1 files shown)


src/components/import/ImportTemplateModal.tsx
```diff
@@ -225,7 +226,12 @@ export function ImportTemplateModal(props: Props) {
 
           <Button
             behavior="button"
-            onClick={() => setIsSaveTemplateModalOpen(true)}
+            onClick={() => {
+              setIsTemplateModalOpen(false);
+              setTimeout(() => {
+                setIsSaveTemplateModalOpen(true);
+              }, 310);
```
beganovich, line 233: Is there reason why 310?

Civolilah: @beganovich The only reason for it is actually that I want to make sure everything is unmounted from the first modal. We have some transitions of 200 and 300 milliseconds in the Modal base component, so I want to make sure that everything is correctly unmounted before opening a second modal. That is the only reason.

beganovich (approved):

PR #2116, opened 2024-10-09
[Cleanup] Refactoring Totals Calculation Logic | Purchase Orders

Civolilah:
@beganovich @turbo124 The PR includes just a quick refactoring of the totals calculation logic for purchase orders to align with other entities and to avoid unnecessary calculations. Let me know your thoughts.


Civolilah:
@turbo124 Can you please just confirm that calculations for this entity work as expected?

beganovich (approved):

PR #2119, opened 2024-10-10
[Bugfix] Logic And Modal Preferences Issue | Dashboard

Civolilah:
@beganovich @turbo124 The PR includes two fixes: one for the preferences logic and another for the currencies dropdown in the preferences modal. Screenshot:

Screenshot 2024-10-10 at 19 17 46

Let me know your thoughts.

### DIFF
--- a/src/common/hooks/usePreferences.tsx
+++ b/src/common/hooks/usePreferences.tsx
@@ -62,7 +62,7 @@ interface SaveOptions {
 }
 
 export function usePreferences() {
-  const user = useInjectUserChanges();
+  const user = useInjectUserChanges({ overwrite: false });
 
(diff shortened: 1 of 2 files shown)


Civolilah:
@turbo124 The issue was with the logic of updating the preferences currency if the selected currency from the preferences modal's dropdown was not available. The steps to recreate the bug:

- Select a currency in the preferences modal that does not exist in the dropdown list on the dashboard outside the modal, then save it
- Once you've saved the currency, whenever you go somewhere else and return to the dashboard, it will appear as if "All" currencies are selected

turbo124:
Thanks!

@beganovich => main please

beganovich (approved):

PR #2127, opened 2024-10-14
[Feature] Implementing "Bulk Updates" Actions For Expenses && Recurring Invoices

Civolilah:
@beganovich @turbo124 The PR includes the implementation of the "bulk_updates" bulk action for `Expenses` and `Recurring Invoices`. The only available field is the tax selector for tax columns. Screenshot:

Screenshot 2024-10-15 at 07 53 09

`Note`: @turbo124 Before testing, please make sure that the `bulk_updates` property in the `statics` query is correctly updated for those two entities.

Currently, I have implemented a solution calling those columns "tax_1", "tax_2", and "tax_3".

Let me know your thoughts.


turbo124:
@Civolilah could you use translations tax1,tax2,tax3 for the labels please?

turbo124:
image

when selecting a tax, i do not get the second option for the tax to be selected.

Civolilah:
> image
> 
> when selecting a tax, i do not get the second option for the tax to be selected.

@turbo124 Okay, so the solution has been adjusted using the old data. Now, as I can see in the `bulk_updates`, it returns `tax1`, `tax2`, and `tax3`. According to that change, the solution has been adjusted and it should display selectors properly. Screenshot:

Screenshot 2024-10-20 at 16 25 52

Let me know your thoughts.

turbo124:
instead of newValue, can you please pass snake case new_value

Civolilah:
> instead of newValue, can you please pass snake case new_value

@turbo124 Ah okay, so this was my mistake. Right now it's updated to be "new_value" instead of "newValue", as it was for the clients endpoint already. It looks like it's working great on my end. Let me know your thoughts.

src/common/queries/expenses.ts
```diff
@@ -90,23 +90,36 @@ export function useExpensesQuery(params: ExpensesParams) {
...
+const successMessages = {
+  bulk_update: 'updated_records',
+};
+
 export function useBulk() {
   const queryClient = useQueryClient();
   const invalidateQueryValue = useAtomValue(invalidationQueryAtom);
 
-  return (
+  return async (
```
beganovich, line 101: Am I missing something or is there a reason these fns are marked as async? 
No awaits inside?

beganovich (commented):
Looks good, tiny info needed.

Civolilah:
@beganovich @turbo124 Dave, could this PR please go through one more QA iteration? When we merged the PR related to tax selector value uniqueness (name + rate), it wasn't updated here, which caused an issue. I've fixed it now, and it should work as expected.

beganovich (approved):

PR #2134, opened 2024-10-16
[Cleanup] Labels For Non-Existing Taxes In Company | Expenses

Civolilah:
@beganovich @turbo124 The PR includes displaying taxes as labels on the expense page if the tax is not created in the company but the expense has been imported via XML file. Screenshot of the solution when taxes have names:

Screenshot 2024-10-16 at 18 25 48

Screenshot of the solution when taxes do not have names:

Screenshot 2024-10-16 at 18 50 28

Let me know your thoughts.


Civolilah:
@turbo124 The approach for selecting values in the tax selector has been changed. Instead of selecting taxes by name in the selector, we will now have both name and rate, which will create a unique value for a specific tax. This is in line with what we discussed on Slack. Let me know your thoughts.

beganovich (approved):

PR #2145, opened 2024-10-21
[Bugfix] Time Formatting

### LINKED ISSUE #2142: [BUG] Task-time-details faulty
Using Docker on version 5.10.41

Only (hopefully) in the React-web-app task-time-details are faulty in invoices after creating an invoice out of a task:
grafik

1. Localization is set to 24 hours, but it's formated into AM/PM
2. Sometimes the sum of the times is 0 when it should be > 0

Civolilah:
@beganovich @turbo124 The PR includes formatting of times in the app when we display them. Wherever I found a fixed time format, I replaced it with a dynamic one relying on the `military_time` company settings property. Screenshot:

Screenshot 2024-10-21 at 19 00 34

Let me know your thoughts.

Closes #2142

### DIFF
--- a/src/common/helpers.ts
+++ b/src/common/helpers.ts
@@ -15,6 +15,7 @@ import { route } from '$app/common/helpers/route';
 import entityState from './constants/entity-state';
 import { request } from './helpers/request';
 import { useCurrentCompanyDateFormats } from './hooks/useCurrentCompanyDateFormats';
+import { useCompanyTimeFormat } from './hooks/useCompanyTimeFormat';
 
 export function isHosted(): boolean {
   return import.meta.env.VITE_IS_HOSTED === 'true';
@@ -62,6 +63,22 @@ export function date(date: number | string, format: string) {
   }
 
   return dayjs(date).format(format);
+}
+
+export function useFormatTime() {
+  const { timeFormat } = useCompanyTimeFormat();
+
+  return (date: number | string, format?: string) => {
+    if (date === 0 || date === '' || date === undefined) {
+      return '';
+    }
+
+    if (typeof date === 'number') {
+      return dayjs.unix(date).format(format ?? timeFormat);
+    }
+
+    return dayjs(date).format(format ?? timeFormat);
+  };
 }
 
 export function useParseDayjs() {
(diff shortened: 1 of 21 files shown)


src/common/helpers.ts
```diff
@@ -64,6 +65,22 @@ export function date(date: number | string, format: string) {
   return dayjs(date).format(format);
 }
 
+export function useFormatTime() {
```
beganovich, line 68: Isn't this a duplicate of `useDateTime`?

Civolilah: @beganovich Actually, it's not a duplicate because `useFormatTime` is dedicated only to formatting time values, while `useDateTime` formats the full `date + time`. However, since you mentioned it, I believe it's better to have it under `useDateTime` so we can make it more global. So, I've modified the `useDateTime` hook to accept time-only formatting. Let me know your thoughts.

src/common/hooks/useDateTime.ts
```diff
@@ -37,16 +31,18 @@ export function useDateTime() {
...
-      return dayjs
-        .utc(date)
-        .tz(timeZone || companyTimeZone)
-        .format(finalFormat);
+      return dayjs.unix(date).format(finalFormat);
     }
 
-    return dayjs
-      .utc(date)
-      .tz(timeZone || companyTimeZone)
```
Civolilah, line 49: @beganovich After our quick discussion in the group about time zones, I'm now aware that we don't need to add timezone here. So, I've removed it.

beganovich: But this hook is used in different places? Is the behavior same after your change?

Civolilah: @beganovich Currently, the hook is only used in three places for formatting the `next_send_datetime` column value in Recurring Invoices. While we have many similar column formatting instances throughout the tables, this particular one formats both date and time.

The behavior will change as it will no longer use the company timezone. However, this aligns with our existing formatting logic, as we don't use company timezones for any other date or time column formatting in the application. The `dayjs` will take the timezone from the browser and format it into that timezone, which is how we handle all other formatting.

Basically change aligns logic with the rest of the app.

beganovich (approved):

PR #2174, opened 2024-11-09
[Cleanup] Importing "location" From react-router-dom Pacakge

Civolilah:
@beganovich @turbo124 The PR imports the `useLocation` hook from `react-router-dom` package instead of using the default `location` class previously used in "Documents" components, to avoid unexpected behavior. Let me know your thoughts.


turbo124:
@Civolilah this fixes the issue that i am seeing.

Civolilah:
> @Civolilah this fixes the issue that i am seeing.

@turbo124 Okay, that's great! Then I believe it happens very rarely on some browser versions, but now we're using react-router-dom package and it should be consistent with the rest of the app. Can we mark this as "QA Passed"?

beganovich (approved):

PR #2180, opened 2024-11-10
[Cleanup] Bulk Email Quotes Action | New Options

Civolilah:
Work in progress..


Civolilah:
@turbo124 Could you add the `email_count_quotes` translation keyword with a `:count` variable, similar to what we have for invoices, to maintain consistency?

beganovich (approved):

PR #2185, opened 2024-11-11
[Cleanup] Making Payment Means Form Available For All E-Invoice Types

Civolilah:
@beganovich @turbo124 The PR includes adjusting the logic so the 'Payment Means' form is available for all e-invoice types with ONLY one criterion: `enable_e_invoice` must be turned on. Screenshot:

Screenshot 2024-11-11 at 18 44 40

Let me know your thoughts.

### DIFF
--- a/src/components/e-invoice/PaymentMeans.tsx
+++ b/src/components/e-invoice/PaymentMeans.tsx
@@ -385,63 +385,66 @@ export const PaymentMeans = forwardRef<PaymentMeansFormComponent, Props>(
     useEffect(() => {
       form.setFieldValue(
         'payment_means.0.code',
-        get(company.e_invoice, 'Invoice.PaymentMeans.0.PaymentMeansCode.value')
+        get(
+          company.e_invoice,
+          'Invoice.PaymentMeans.0.PaymentMeansCode.value'
+        ) || '1'
       );
       form.setFieldValue(
         'payment_means.0.iban',
         get(
           company.e_invoice,
           'Invoice.PaymentMeans.0.PayeeFinancialAccount.ID.value'
-        )
+        ) || null
       );
       form.setFieldValue(
         'payment_means.0.bic_swift',
         get(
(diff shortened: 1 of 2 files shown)


src/components/e-invoice/PaymentMeans.tsx
```diff
@@ -380,7 +378,7 @@ export const PaymentMeans = forwardRef<PaymentMeansFormComponent, Props>(
         ? PAYMENT_MEANS_FORM_ELEMENTS[
             form.values.payment_means[0]
               .code as keyof typeof PAYMENT_MEANS_FORM_ELEMENTS
-          ]
+          ] || []
```
Civolilah, line 381: @beganovich When the component is loading for the first time, I experienced page breaking because the `code` prop had a value that did not match any elements from the array. As a result, the elements variable was `undefined` and caused the page to break, so I added a fallback to an empty array for any edge cases.

turbo124: @Civolilah we will need to set a default of "1" for the code please if none is present.

image

Otherwise user will never be able to save

Civolilah: @turbo124 Sure thing, the code has been set to `'1'` by default if nothing is populated, and I also disabled the possibility of clearing the value (setting to empty string) for this selector so we can avoid having an empty string as a value from the React side. Let me know your thoughts.

turbo124:
Still does not pass validation for me? 

{"entity":"company","payment_means":[{}]}

Civolilah:
> Still does not pass validation for me?
> 
> {"entity":"company","payment_means":[{}]}

@turbo124 Okay, yeah, I made a fresh seed of the data and succeeded in recreating your case. I've adjusted the part of the code that was overriding initial values. It should be fixed for you now - please let me know if you see anything else that needs fixing.

PR #2186, opened 2024-11-11
[Cleanup] E-Invoice Entity Validation | Invoice Edit Page

Civolilah:
@beganovich @turbo124 The PR includes adjusting the validation flow and removing unnecessary banners that we have already implemented. For the invoice edit page only, we call the `/api/v1/einvoice/validateEntity` endpoint, which returns an object with a `"passes"` property if successful. If this property is set to `false`, we will need to display a banner with messages. Additionally, when "passes" is false, we will DISABLE the "Save" button when the user is on the "E-Invoice" tab. The messages will be displayed in the same way we display error message banners throughout the rest of the app to maintain consistency. Screenshot:

Screenshot 2024-11-12 at 00 05 25

The data is currently mocked from the ticket because we need to wait for a release that will update this endpoint's response to align with the ticket's specifications.

Let me now your thoughts.


Civolilah:
@turbo124 The endpoint will only be called if `enable_e_invoice` is set to true and `PEPPOL` is selected as the e_invoice_type.

Civolilah:
@turbo124 After your quick guidelines in the ticket, it was much clearer for me to implement the adjusted solution here. So, a few things have been changed/implemented here:

- The result for client/company is taken from a 422 error which is returned from the API and displayed in the error banner
- The 'E-Invoice' tab now displays the number of errors if any are returned
- I've implemented the `check_invoice` button to check the invoice state independently, after which a modal will open with the final result from that endpoint

Screenshots:

Screenshot 2024-11-13 at 19 07 50

Screenshot 2024-11-13 at 19 18 57

In the private chat, I mentioned the issue that I experienced with the response when checking the invoice entity. It returns a 500 error with a slightly strange response, which seems like something is wrong on the API side.

`Important note`: The solution is adjusted so the invoice check endpoint also catches 422 error also, but I assumed that errors are returned as an ARRAY of STRINGS just as you gave as an example in the ticket (first example).

Let me know your thoughts.

Civolilah:
@turbo124 As we discussed on Slack, a few things have changed:

- The "Required" text has been added to the validation banner along with a link for editing the specific entity
- The button's translation keyword has been changed to "check".
- The errors for invoice entity are displayed outside of the modal, but if there are no errors, only a "Success" toaster is displayed

Screenshot:

Screenshot 2024-11-14 at 17 44 58

Let me know your thoughts.

beganovich (approved):

beganovich (approved):

PR #2189, opened 2024-11-12
E-invoices: Few improvements

beganovich:
This includes the following improvements for invoicing:
- Licence checks are skipped when in development, so we don't run in rate limiting issue.
- Additional legal identifiers form is now cleared upon successful submit.


turbo124:
image

When attempting to add new identifier.

PR #2193, opened 2024-11-13
[Feature] Default Gateway Functionality

Civolilah:
@beganovich @turbo124 The PR includes the implementation of setting the default gateway on the Client show page. Screenshot:

Screenshot 2024-11-14 at 19 11 26

`Note`: We are missing the "set_as_default" translation keyword. If you agree, please add it to the translation files.

Let me know your thoughts.


turbo124:
@Civolilah please use translation save_as_default

Civolilah:
> @Civolilah please use translation save_as_default

@turbo124 The "set_as_default" translation has been changed to "save_as_default". Let me know your thoughts.

turbo124:
@Civolilah

After setting the payment method as default, the UI never changes.. ie the badges need to change their statuses.

We may also want to add a different shade of color to the button action vs badge to differentiate.

Civolilah:
> @Civolilah
> 
> After setting the payment method as default, the UI never changes.. ie the badges need to change their statuses.
> 
> We may also want to add a different shade of color to the button action vs badge to differentiate.

@turbo124 Yeh, the source of gateway tokens is in the client object itself, so client invalidation has been added for updating badges when the default gateway changes. I also updated the styling to clearly distinguish between default (non-clickable) and interactive badges. Screenshot:

Screenshot 2024-11-18 at 17 57 08

Let me know your thoughts.

beganovich:
My two cents, unless requested, please avoid super custom styles such as that fully rounded one. Across the app we use standard rounding. If you absolutely love it, then please adjust x-padding to give it more space and make sure border is compatible with dark mode.

Civolilah:
> My two cents, unless requested, please avoid super custom styles such as that fully rounded one. Across the app we use standard rounding. If you absolutely love it, then please adjust x-padding to give it more space and make sure border is compatible with dark mode.

@beganovich Hmm, I checked the style of other badges and noticed some of them are fully rounded. That's why I chose this style, and after trying less rounded variants, I think this looks better. But, I'll adjust the padding as you suggested.

turbo124:
image


Looks good, minor point. I think the two action buttons are  too close to each other, this could cause users to misclick...

Perhaps the default button should move to the bottom left to avoid this. the Default badge itself should stay.

Civolilah:
> image
> 
> Looks good, minor point. I think the two action buttons are too close to each other, this could cause users to misclick...
> 
> Perhaps the default button should move to the bottom left to avoid this. the Default badge itself should stay.

@turbo124 Okay, sure. So, I've adjusted styling just as you explained. Let me know your thoughts.

beganovich (approved):

PR #2218, opened 2024-11-20
[Cleanup] Implementing Warning Banner For Payments If Amout Is Negative

Civolilah:
@beganovich @turbo124 The PR implements a warning banner that appears when a negative amount is entered. Screenshot:

Screenshot 2024-11-20 at 19 01 40

Let me know your thoughts.

### DIFF
--- a/src/components/Banner.tsx
+++ b/src/components/Banner.tsx
@@ -9,16 +9,17 @@
  */
 
 import classNames from 'classnames';
-import { ReactNode } from 'react';
+import { CSSProperties, ReactNode } from 'react';
 
 interface Props {
   variant: 'orange' | 'red';
   children: ReactNode;
   className?: string;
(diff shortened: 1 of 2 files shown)


turbo124:
The banner appears after the user has already saved the payment. 

The banner should appear under IF the amount is negative using either debounce/onBlur After the payment is saved, this warning is too late.

Civolilah:
> The banner appears after the user has already saved the payment.
> 
> The banner should appear under IF the amount is negative using either debounce/onBlur After the payment is saved, this warning is too late.

@turbo124 Okay, so I've changed the approach. Instead of using an onBlur event to update the field's value, I've implemented a debounce override. Now, when the value changes on the Create page, if it's negative, the banner will appear after a 300ms delay. Let me know your thoughts.

PR #2220, opened 2024-11-21
Improvements for tax settings

beganovich:
This includes couple changes to tax settings, including automatic setting of tax values when calculate taxes is on, in-app guide and hiding the invoice tax rates.


beganovich:
Updated :+1:

PR #2230, opened 2024-11-25
[Cleanup] Formatting next_send_datetime Property By Timezone | Recurring Invoices

Civolilah:
@beganovich @turbo124 The PR includes formatting the `next_send_datetime` property for Recurring Invoices in three locations (all that I found):

- Dashboard Recurring Invoices table
- Recurring Invoices slider
- Recurring Invoices main table

Let me know your thoughts.

### DIFF
--- a/src/common/hooks/useCompanyTimeZone.ts
+++ b/src/common/hooks/useCompanyTimeZone.ts
@@ -11,6 +11,7 @@
 import { useStaticsQuery } from '$app/common/queries/statics';
 import { useEffect, useState } from 'react';
 import { useCurrentCompany } from './useCurrentCompany';
+import { Timezone } from '../interfaces/statics';
 
 export function useCompanyTimeZone() {
   const company = useCurrentCompany();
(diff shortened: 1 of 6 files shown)


Civolilah:
@turbo124 When researching setting values across different levels, we consider the hierarchy: first client, then group, and finally company level. The video only mentioned client and company levels, but should group level be included in these checks as well?

turbo124:
@Civolilah looks good, we will end up reusing this logic in quite a few places, so if it isn't yet in a hook.....

Civolilah:
> @Civolilah looks good, we will end up reusing this logic in quite a few places, so if it isn't yet in a hook.....

@turbo124 Yes, we have `useDateTime`, `useGetSetting`, and `useGetTimezone` hooks. Everything is global, so we just need to use the `useDateTime` hook and specify whether we want formatting with timezone or not. We can additionally pass specific formats or timezones. So yeah, it's very easy to reuse this logic.

beganovich (approved):

PR #2244, opened 2024-12-02
[Cleanup] Updating Payload For Designs Default Endpoint

Civolilah:
@beganovich @turbo124 The PR includes updating the payload for the `/api/v1/designs/set/default` endpoint by adding `settings_level` and `client_id/group_settings_id` properties. Let me know your thoughts.


turbo124:
you are using group as settings prop level, but this is invalid:

company,client,group_settings

Civolilah:
> you are using group as settings prop level, but this is invalid:
> 
> company,client,group_settings

@turbo124 Ah, my bad - I did not notice "_settings" for the group level. The logic has been adjusted so that if it is group level, the `setting_level` will be set as "group_settings"; otherwise, it will be "company/client". Let me know your thoughts.

beganovich (approved):

PR #2247, opened 2024-12-02
[Bugfix] Missing Mapping Of Custom Fields When Converting Or Adding To Invoice

### LINKED ISSUE #2246: Missing custom field data from invoices
This may be a compatibility thing between flutter and react - however invoicing projects will not have a date set even when they're set in the individual tasks

Civolilah:
@beganovich @turbo124 The PR adds missing mappings of custom fields when converting or adding different entities to an invoice. The fixed cases include: adding an expense to an invoice, invoicing an expense, and invoicing a project. 

Closes #2246 

Let me know your thoughts.

### DIFF
--- a/src/pages/expenses/common/components/AddToInvoiceAction.tsx
+++ b/src/pages/expenses/common/components/AddToInvoiceAction.tsx
@@ -113,6 +113,10 @@ export function AddToInvoiceAction(props: Props) {
                 expense.tax_amount3,
                 expense.tax_rate3
               ),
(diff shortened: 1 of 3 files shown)


turbo124:
@Civolilah 


For tasks, if there are custom values, these are mapped to the line item of each invoice entry, so insteading of injecting into invoice.custom_valueX

as you iterate through the line items of tasks, you inject into the product.custom_valueX level.

beganovich (approved):

PR #2258, opened 2024-12-09
Notifications: Generic message

beganovich:
This is a frontend addition for: https://github.com/invoiceninja/invoiceninja/pull/10361


turbo124:
Selfhost able to use this also? Or, is there a way to be selective in target audience?

turbo124:
I think a one time modal asking for their preference for these notifications can be offered for self hosted. 

Only listening to public channel of course if that is able to be done at a granular level?

beganovich:
Notifications are now opt-in for self-hosted user with one-time popup. Please let me know how it looks :+1:

PR #2259, opened 2024-12-09
Resizable tables

### LINKED ISSUE #2235: Option to adjust columns width (Like in Excel)
Invoice Ninja 5
Hosted
Web browser version

---
Some companies have very long legal names, taking way too much space and making it hard to see all the selected columns, even when I reduced the page size to 80% in Chrome.

image

Resizing should be "remembered" between visits (cookies?)
Resizing settings should be independent of types like Tasks, Quotes, Invoices, Expenses, etc.

Something like this:
image

Thank you for your consideration

beganovich:
Resolves #2235.

Notes:
— Uniquely resizeable tables
— Automatic resize for every data table
— Persisted changes per device

https://github.com/user-attachments/assets/153d5e9a-2d24-4ace-a6ae-aa32a0f1526e


turbo124:
intuitively for me, i want to double click on that divider and the column resize all the way down to the up/down chevrons.

It looks like we have a minimum column width being enforced?

turbo124:
@beganovich is there a fixed minimum size for the columns? i think this needs to be further reduced.

turbo124:
:fire: :fire: :fire: :fire:

PR #2262, opened 2024-12-10
[Bugfix] Removing Default Quick Actions

Civolilah:
@beganovich @turbo124 The PR includes a fix for displaying default quick actions when no preferences are stored for a specific entity. Let me know your thoughts.


turbo124:
@Civolilah 

I'm not sure about this. We shouldn't force this one on users, this should be opt in only.

Civolilah:
> @Civolilah
> 
> I'm not sure about this. We shouldn't force this one on users, this should be opt in only.

@turbo124 So, this is how we initially imagined the solution for quick actions - with two default actions - so I just fixed that here. However, if you prefer NOT to have this part of the logic anymore, I can remove it. Let me know what you prefer.

> @Civolilah
> 
> I'm not sure about this. We shouldn't force this one on users, this should be opt in only.

@turbo124 So, this is how we initially imagined the solution for quick actions - with two default actions - so I just fixed that here. However, if you prefer NOT to have this part of the logic anymore, I can remove it. Let me know what you prefer.

turbo124:
@Civolilah 

Are you saying that we already set some default actions now without this PR?

Civolilah:
> @Civolilah
> 
> Are you saying that we already set some default actions now without this PR?

@turbo124 Exactly, in the very first PR for the Invoice entity we had them implemented, but I made a small logical mistake which caused an issue when the second entity was added to the implementation. That is how we had imagined the solution for the Invoice entity.

beganovich (approved):

PR #2263, opened 2024-12-10
[Bugfix] Unnecessary By Default Checked "All" Checkbox On Tables

Civolilah:
@beganovich @turbo124 The PR includes fixes for an unnecessary checked "all" checkbox on tables. Currently, if we have no data in a specific table and then create one record, after navigating to any other page and returning to the table where the record was created, the "all" checkbox is checked by default, which is unnecessary. This fix addresses only this specific case. Let me know your thoughts.


turbo124:
I could not recreate this one myself, but assume the fix is appropriate

beganovich (approved):

PR #2264, opened 2024-12-10
[Feature] Implementing Delete Functionality For Gateway Tokens

Civolilah:
@beganovich @turbo124 The PR implements the functionality to delete gateway tokens. Screenshot:

Screenshot 2024-12-11 at 15 13 15

Let me know your thoughts.

### DIFF
--- a/src/components/dropdown/Dropdown.tsx
+++ b/src/components/dropdown/Dropdown.tsx
@@ -30,6 +30,9 @@ interface Props extends CommonProps {
   label?: string | null;
   cardActions?: boolean;
   cypressRef?: string;
+  customLabel?: ReactNode;
+  minWidth?: string;
+  maxWidth?: string;
 }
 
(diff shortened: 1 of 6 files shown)


Civolilah:
@turbo124 From my perspective, a delete button fits much better in this context compared to a trash icon.

Civolilah:
@turbo124 The solution has been adjusted as we discussed on Slack. We now have arrows on buttons that will display the delete option, similar to how we handle additional save options for some entities. Let me know your thoughts.

turbo124:
resize_borked

I think we need to rethink where we have this dropdown so that it is always visible

Civolilah:
@turbo124 As we discussed on Slack, the width of cards has been adjusted for the Client show page. Let me know your thoughts.

beganovich (approved):

PR #2265, opened 2024-12-11
Option to use system fonts

beganovich:
This allows users to use system fonts.

Settings > User details > Preferences

image


turbo124:
:fire:

PR #2271, opened 2024-12-14
[Cleanup] Removing Toaster For /quota Endpoint Errors

Civolilah:
@beganovich @turbo124 The PR removes the logic for displaying toaster errors on 401 and 403 errors from the **/quota** endpoint.. Let me know your thoughts.

### DIFF
--- a/src/common/helpers/request.ts
+++ b/src/common/helpers/request.ts
@@ -13,8 +13,9 @@ import { defaultHeaders } from '$app/common/queries/common/headers';
 import { ValidationBag } from '../interfaces/validation-bag';
 import { toast } from './toast/toast';
 import { $refetch } from '../hooks/useRefetch';
-import { checkJsonObject } from '../helpers';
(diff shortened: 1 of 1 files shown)


src/common/helpers/request.ts
```diff
@@ -39,19 +38,6 @@ client.interceptors.response.use(
     return response;
   },
   (error: AxiosError<ValidationBag>) => {
-    if (
```
Civolilah, line 42: @beganovich Is this logic only for the /quota endpoint?

beganovich: No, we use it for all einvoice endpoints. Can you please do double-checks to make sure we exclude quota?

Civolilah: @beganovich Okay, so I ensured that the toaster for the /quota endpoint is no longer displayed.

beganovich (approved):

PR #2275, opened 2024-12-17
Fixes for socket preferences

beganovich:
This fixes the issue with popup showing on guest pages.


turbo124:
@beganovich unsure how the last change made it into dev, we will need to merge this asap, so i can retag a new verison.

PR #2291, opened 2025-01-03
[Feature] Implementing "Previous/Next" Icons | Recurring Invoices

Civolilah:
@beganovich @turbo124 The PR includes the implementation of "previous/next" arrows for Recurring Invoices. Screenshot:

Screenshot 2025-01-03 at 19 35 21

Let me know your thoughts.


turbo124:
functionality is good, we just need  (if possible) to place next/previous on hover.

We should also deploy this across all entities to keep the UI consistent.

Civolilah:
> functionality is good, we just need (if possible) to place next/previous on hover.
> 
> We should also deploy this across all entities to keep the UI consistent.

@turbo124 Sure thing - both changes have been made. The arrows will be displayed when hovering over the breadcrumbs section, and the functionality has been applied to all entities. Let me know your thoughts.

Civolilah:
@turbo124 As we discussed, this functionality will be displayed by default in the breadcrumbs section instead of on hover, while maintaining the currently implemented tooltip. Screenshot:

Screenshot 2025-01-06 at 10 53 45

Let me know your thoughts.

beganovich (approved):

PR #2295, opened 2025-01-07
[Bugfix] Input Value Changes On Product Selector

Civolilah:
@beganovich @turbo124 The PR includes fixes for changing the input value on the product selector. Let me know your thoughts.


src/pages/invoices/common/hooks/useResolveInputField.tsx
```diff
@@ -325,6 +325,7 @@ export function useResolveInputField(props: Props) {
         <ProductSelector
           key={`${property}${resource?.line_items[index][property]}`}
           onChange={(value) =>
+            value.resource &&
```
beganovich, line 328: @Civolilah can you elaborate this line change? We gotta be extra careful with such changes.

Civolilah: @beganovich This change is not required, so I've removed it. I added it while trying to find the source of the issue. However, it would not have caused an issue because the `onChange` function is dedicated to changing values in the dropdown by selecting the whole resource - if there is no resource, we cannot take prop values from it. For the product selector, we have a specific case where the entered input value needs to be set as `product_key` without specific selection. So, while this has been removed, the issue is resolved by properly using the `onInputValueChange` function.

PR #2300, opened 2025-01-08
[Cleanup] Implementing New Approach For Saving Table Filters

Civolilah:
@beganovich @turbo124 The PR includes changes for saving table filter preferences. Currently, only main tables will be responsible for saving them, and all other tables will use these same saved settings.

`Note`: A small part of the code has been added to delete all previously saved filter preferences that used the old approach. After the first save of the new filter preferences, the react_settings object will be significantly smaller and clearer with only saved preferences for current approach.

 Let me know your thoughts.


Civolilah:
> In progress...

turbo124:
@beganovich QA is more on implementation for this PR.

beganovich (approved):

PR #2315, opened 2025-01-17
[Bugfix] "Line Total" Column On Line Items Table

Civolilah:
@beganovich @turbo124 The PR includes changes to the 'Line Total' column - if the column is included in product/task variables, it will be displayed, otherwise, it will be hidden. Additionally, its position will match the position specified in the variables. Let me know your thoughts.

### DIFF
--- a/src/pages/invoices/common/hooks/useProductColumns.ts
+++ b/src/pages/invoices/common/hooks/useProductColumns.ts
@@ -175,14 +175,6 @@ export function useProductColumns() {
       }
     });
 
-    // In some cases, the user may add the line_total variable in a different position than the last.
-    // We want to ensure that it is completely excluded from the variables so that it can be added in the last position.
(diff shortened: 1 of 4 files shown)


Civolilah:
@turbo124 I've implemented your recommendation to push label columns to the end of the line items table and fixed some additional issues related to this functionality.

I discovered that the `$task.tax_amount` variable was missing from the dropdown when configuring Task Columns under Invoice Design.

Additionally, I found that for both Product and Task columns, the "tax" and "tax_amount" columns were always present in the dropdown. However, comparing this to the Flutter app, this shouldn't always be the case. These variables should ONLY be displayed in the dropdown when `enabled_item_tax_rates` is greater than 0, indicating that one or more taxes are enabled for line items.

All of these issues have been fixed and adjusted.

Let me know your thoughts.

beganovich (approved):

PR #2319, opened 2025-01-21
[Bugfix] Debounce Subject And Body Changing Value Instead of onBlur

Civolilah:
@beganovich @turbo124 The PR includes changes to trigger the API (generate PDF) on debounce when values change, instead of on the onBlur event for Email pages. Let me know your thoughts.


src/components/forms/InputField.tsx
```diff
@@ -104,13 +104,15 @@ export function InputField(props: Props) {
           )}
           placeholder={props.placeholder || ''}
           onBlur={(event) => {
-            event.target.value =
-              event.target.value === '' && props.type === 'number'
-                ? '0'
-                : event.target.value;
+            if (!props.changeOverride) {
```
Civolilah, line 107: @beganovich We forgot to avoid unnecessarily triggering the onValueChange function if this prop isn't set to "true". So, basically the function was being called twice - I just added this check to avoid calling it onBlur.

beganovich:
@turbo124 When you get a chance please QA this one, so I can patch the production.

beganovich:
Parent https://github.com/invoiceninja/ui/pull/2309.

beganovich (approved):

PR #2321, opened 2025-01-21
[Bugfix] Adding Maximum Clients Validation Message On Client Create Modal

Civolilah:
@beganovich @turbo124 The PR includes adding a validation message for maximum clients if it is returned from the API under the "Name" client field. Screenshot:

Screenshot 2025-01-22 at 00 08 26

 Let me know your thoughts.

### DIFF
--- a/src/pages/invoices/common/components/ClientCreate.tsx
+++ b/src/pages/invoices/common/components/ClientCreate.tsx
@@ -168,7 +168,7 @@ export function ClientCreate({
                   label={t('name')}
                   value={client?.name || ''}
                   onValueChange={(value) => handleChange('name', value)}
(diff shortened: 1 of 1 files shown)


Civolilah:
@turbo124 I remember that we worked on improving the clients limit message a few months ago. We added additional explanatory text with a link alongside the message that is returned from the API. However, it looks like we missed the translation keyword for the additional text we wanted to add there. Screenshot:

Screenshot 2025-01-22 at 00 10 27

The `clients_limit` translation keyword is missing.

turbo124:
Does this cover the modal (when creating a client from the invoice itself?)

beganovich (approved):

PR #2322, opened 2025-01-21
[Bugfix] Fixes For DateRange Selector Width

Civolilah:
@beganovich @turbo124 The PR includes width fixes for the date range selector. Screenshot:

Screenshot 2025-01-22 at 00 21 46

Let me know your thoughts.


Civolilah:
> I am seeing an edge case here,
> 
> image
> 
> when clicking on the date icon, it sometimes triggers the sort column of the adjacent column (Number)

@turbo124 Okay, so I was not able to recreate it, but I made some adjustments to prevent this case from happening. Let me know your thoughts.

turbo124:
#2326

PR #2323, opened 2025-01-22
[Bugfix] Auto-Fill Line Item When Search Term Is Entered In Product Selector

Civolilah:
@beganovich @turbo124 The PR includes fixing an issue where line items were not auto-filling when a search term was entered into the product selector. The conflict was between the onBlur event on the input field and selecting an option by clicking. Let me know your thoughts.


turbo124:
=> main please

beganovich (approved):

PR #2330, opened 2025-01-26
[Cleanup] Adjusting Company Switcher Per New Design

Civolilah:
@beganovich @turbo124 The PR includes the redesign of the company switcher using the existing color scheme and font sizes but with a new design. Screenshot:

Screenshot 2025-01-29 at 18 57 41

Let me know your thoughts.


Civolilah:
@turbo124 The only color I used from Figma is for the icons of add_company, logout, and other actions. It looks really great to me in both light and dark mode - better than any other from the color scheme. All other colors are used from the existing scheme.

turbo124:
After minimising the sidebar, how do we re-expand it? cc @beganovich

Civolilah:
> After minimising the sidebar, how do we re-expand it? cc @beganovich
> 
> ok I found it:
> 
> image
> 
> Problem. it is not very intuitive for me (perhaps we need to relocate this to bottom so that it is always in context.
> 
> Also, i note a _huge_ delay when clicking to reopen the sidebar
> 
> image
> 
> I think this is related to the API call for updating the react_settings for this, we should not need to await this.

@turbo124 The icon has been replaced and the expanding/collapsing sidebar speed has been improved. Let me now your thoughts.

src/common/hooks/useInjectUserChanges.ts
```diff
@@ -31,7 +31,7 @@ export function useInjectUserChanges(options?: Options) {
   const changes = useUserChanges();
 
   useEffect(() => {
-    if (changes && options?.overwrite === false) {
```
Civolilah, line 34: @beganovich I don't think this ever worked correctly. David reported that the speed of collapsing/expanding sidebar is really slow - the reason for that is exactly this logic. The changes object is constantly being injected again because this if statement is almost never true.

By default, the changes object used in this if statement is an empty object ({}), it is never null/undefined. So, that part of the condition is always true, while the second part is equal to false only if the overwrite is passed through options. So, instead of comparing types also, we just want to check if the overwrite is set to true.

The overwrite is never passed as a true value, but also I don't think we ever need to overwrite changes over the changes we currently have.

turbo124:
image


both change buttons need to be in the same location to keep the UX consistent, place them _both_ in the footer section.

Civolilah:
> image
> 
> both change buttons need to be in the same location to keep the UX consistent, place them _both_ in the footer section.

@turbo124 Ah, sure thing! It has been replaced!

turbo124:
image


If no company logo present we should a character, can we please ensure this sizes proportionately to fill the circle.

Civolilah:
> image
> 
> If no company logo present we should a character, can we please ensure this sizes proportionately to fill the circle.

@turbo124 Sure thing, the font size in avatars has been increased. Screenshot:

Screenshot 2025-01-31 at 18 37 19

Let me know your thoughts.

beganovich:
This doesn't look that good imho. Can we make icons smaller and do some spacing to see how it looks? 

image

Civolilah:
> This doesn't look that good imho. Can we make icons smaller and do some spacing to see how it looks?
> 
> image

@beganovich Ah sure, I didn't notice this one because I didn't have that many icons. The sizes of icons have been decreased and spacing has been increased. Screenshots:

Screenshot 2025-02-06 at 00 35 08

Screenshot 2025-02-06 at 00 44 15

Let me know your thoughts.

Civolilah:
> Quick change needed on this one, if you don't have profile picture, and use invoice ninja default one, this is how it looks. Please make use of object-fit (cover) in CSS to fix this.
> 
> image

@beganovich Ah sure thing, it has been fixed. Screenshot:

Screenshot 2025-02-11 at 19 18 20

Let me know your thoughts.

package.json
```diff
@@ -38,6 +38,7 @@
     "pusher-js": "^8.4.0-rc2",
     "randexp": "^0.5.3",
     "react": "^18.2.0",
+    "react-avatar": "^5.0.3",
```
beganovich, line 41: We definitely need to kick out this library. It wasn't updated in years and it will definitely block our path to React 19.

As replacement we can use ui-avatars.com, but we should be careful here as well.

@turbo124 should we use 3rd party website to generate nice avatars for us (due to privacy concerns).

Just to be clear `react-avatar` isn't fully offline either.

src/components/CompanySwitcher.tsx
```diff
@@ -97,6 +116,27 @@ export function CompanySwitcher() {
...
   }, [currentCompany]);
 
+  if (isMiniSidebar) {
+    return (
+      <>
+        <img
+          className={classNames(
+            'rounded-full border overflow-hidden aspect-square',
+            {
+              'object-cover': logo?.includes('invoiceninja-logo@light'),
```
beganovich, line 126: I am not sure this is correct way to handle this, we need same approach for every avatar. Both rounded, 1:1 aspect or 16:9.

Can we apply same this for everyone and test different sizes?

beganovich (changes requested):

src/components/CompanySwitcher.tsx
Civolilah: @beganovich Okay, I made the same approach for all avatars to match the Figma design with the circle and logo. The best preview will be for 1:1 pictures, of course. Screenshot:

Screenshot 2025-02-12 at 18 45 11

However, if we want to have a consistent design, 16:9 pictures for example will be less visible, but I don't think there is a general approach to make all of them look great when they are displayed within the same avatar shape. This is how it looks with a 16:9 picture. Screenshot:

Screenshot 2025-02-12 at 18 50 01

It is not fully visible of course, but I'm not sure if it's even possible when the avatar is consistent and circular. Also, in my opinion, the avatar is a small space to make it visible for all sizes, and I don't think that an avatar is actually a place where it must be fully visible?

Let me know your thoughts.

beganovich: Not to worry, optimize for 1:1 aspect ration, for 16:9 we can cut the content.

beganovich:
Let's dismiss the idea of dynamic avatars with 3rd party sites or packages.

Can we default to a circular Invoice Ninja logo for users who don't have avatars?

Civolilah:
> Let's dismiss the idea of dynamic avatars with 3rd party sites or packages.
> 
> Can we default to a circular Invoice Ninja logo for users who don't have avatars?

,

Civolilah:
> Let's dismiss the idea of dynamic avatars with 3rd party sites or packages.
> 
> Can we default to a circular Invoice Ninja logo for users who don't have avatars?

@beganovich Hmm, okay, I think it looks better when the letter avatar is displayed instead of the default Invoice Ninja logo. Screenshot:

Screenshot 2025-02-13 at 19 07 26

Do you think this would be a good solution? Here's how it would look with this change.

beganovich:
Please use this logo:

https://github.com/invoiceninja/ui/blob/main/public/logo180.png

Civolilah:
> Please use this logo:
> 
> https://github.com/invoiceninja/ui/blob/main/public/logo180.png

@beganovich Sure thing, I made it as a fallback logo and it looks much better now. Screenshot:

Screenshot 2025-02-14 at 16 22 39

Let me know your thoughts.

PR #2356, opened 2025-02-08
[Cleanup] Redesigning 

Civolilah:


beganovich:
This doesn't look like Figma one completely, here's updated version.

Before and after.

image
image

cc @turbo124 for QA.

turbo124:
image


Looks like this for me, i assume once the save button is updated, as long as the proportions are matching...

beganovich:
I guess we can change it afterwards, if it doesn't end up matching with rest of buttons.

beganovich (approved):

PR #2366, opened 2025-02-15
Add ability for chunked file uploads

turbo124:
Provides chunked upload support, only for company backups.

Please review, requires latest v5-develop API changes.

### DIFF
--- a/src/components/import/UploadCompanyImport.tsx
+++ b/src/components/import/UploadCompanyImport.tsx
@@ -1,0 +1,721 @@
+/**
+ * Invoice Ninja (https://invoiceninja.com).
+ *
+ * @link https://github.com/invoiceninja/invoiceninja source repository
+ *
+ * @copyright Copyright (c) 2022. Invoice Ninja LLC (https://invoiceninja.com)
+ *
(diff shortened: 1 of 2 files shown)


beganovich (approved):
This seems to be working fine, no complaints really.

One thing that crossed my mind is, I see this is pretty much copy-paste and then modified version of UploadImport, right?

I wonder if we gonna have this chunked upload concept in other imports, if so, is it worth extracting the changes right now, because at the moment this will leave us with two exactly same looking components (excluding upload progress) with a bit of changes.

turbo124:
@beganovich 

You are correct. The UploadImport was very overloaded, it looked like this handled uploads for the entire UI.

There was no clean way to inject the chunking abstraction into that class. Also, chunking requires API modifications also, so moving this into its own makes sense.

beganovich:
@turbo124 

Do you think we gonna have more cases like this or this is only one?

turbo124:
I think chunking is a very solid solution for some aspects that we may see in the future, so i think the answer here i syes.

beganovich:
I guess it's good to merge now as-is, as soon as we hit next chunk upload requirement, we can refactor and extract.

PR #2370, opened 2025-02-18
Fixes for multiple sockets connection

beganovich:
This fixes multiple sockets connections and implements clean up concept.


turbo124:
testing this, sockets do not work at all now.

PR #2380, opened 2025-02-25
[Feature] Client Locations && Tab Reorganization

Civolilah:
@beganovich @turbo124 The PR includes reorganization of client create and edit page tabs with implementation of locations functionality. Screenshots:

Screenshot 2025-02-28 at 01 05 09

Screenshot 2025-02-28 at 01 05 22

Let me know your thoughts.


turbo124:
Please add validation for country

image

Also, on new create, the fields need to currentLocation?.name currently it is missing ? operator

Please move the phone number to the last input field.

Also, it appears we are missing the selector on the invoice/quote/credit/recurring invoice page where we can select a client location:

image

Civolilah:
> Please add validation for country
> 
> image
> 
> Also, on new create, the fields need to currentLocation?.name currently it is missing ? operator
> 
> Please move the phone number to the last input field.
> 
> Also, it appears we are missing the selector on the invoice/quote/credit/recurring invoice page where we can select a client location:
> 
> image

@turbo124 All of these things should be adjusted right now. Sorry if I missed something, as I needed to work with mocked data. Let me know your thoughts.

turbo124:
@Civolilah 

Still no selector on the invoice page to select the location.

turbo124:
User should be able to change / remove locations. 

We'll want to have functionality like Tax Selector for this so location can be removed.

beganovich:
Looks good, but I think design-wise it's not that good.

1. Cards are too wide, almost optimized for mobiles, instead of desktops. Can we prefer the following layout:

image

2. For modal, I think width is also too wide, why not prefer default size as for rest of modals with input fields?

3. Confirmation modal is empty. Can we please add a simple sentence that explains what's going to happen? 

@turbo124 please let me know if you agree with proposed changes.

turbo124:
@beganovich sure sounds good

Civolilah:
> Looks good, but I think design-wise it's not that good.
> 
> 1. Cards are too wide, almost optimized for mobiles, instead of desktops. Can we prefer the following layout:
> 
> image
> 
> 2. For modal, I think width is also too wide, why not prefer default size as for rest of modals with input fields?
> 3. Confirmation modal is empty. Can we please add a simple sentence that explains what's going to happen?
> 
> @turbo124 please let me know if you agree with proposed changes.

@turbo124 @beganovich All mentioned things have been adjusted. Screenshot:

Screenshot 2025-03-06 at 01 36 12

Please let me know if I missed anything.

beganovich (approved):

PR #2389, opened 2025-03-01
[Cleanup] "Send Time" Confirmation Modal | Email Settings

Civolilah:
@beganovich @turbo124 The PR includes the implementation of a "Send Time" confirmation modal if Send Time is changed, with a message to sync it with Recurring Invoices and Reminders. If the user confirms that they want to sync, I've added "?sync_send_time=true" to the "/api/v1/companies" query. Screenshot:

Screenshot 2025-03-01 at 19 28 03

`Note`: Modal contains plain message text, but I replaced it with the "sync_send_time" translation keyword. If you agree with it, please just add it in the translation files.

Let me know your thoughts.


Civolilah:
@turbo124 As the API does not have this query parameter logic implemented, please let me know if you want to change anything in the implementation.

turbo124:
@Civolilah 

Thanks, we'll need to add a toggle to the confirmation modal. The Sync Send Time will need to be optionally turned on.

We'll need to use the value of the ?query parameter  `?sync_send_time=true` or `?sync_send_time=false`

Civolilah:
> @Civolilah
> 
> Thanks, we'll need to add a toggle to the confirmation modal. The Sync Send Time will need to be optionally turned on.
> 
> We'll need to use the value of the ?query parameter `?sync_send_time=true` or `?sync_send_time=false`

@turbo124 Okay, I've adjusted the solution by adding a toggle in the modal. This is how I imagined the solution with a toggle. Screenshot:

Screenshot 2025-03-02 at 19 35 15

On the screenshot I used plain text, but the text is replaced with "sync_send_time" and "sync_send_time_help".

Let me know your thoughts.

src/pages/settings/gateways/common/components/GatewaysTable.tsx
```diff
@@ -167,7 +167,7 @@ export function GatewaysTable(params: Params) {
 
   useEffect(() => {
     if (updateCompany) {
-      handleCompanySave(updateCompany);
+      handleCompanySave({ excludeToasters: true });
```
beganovich, line 170: I wonder in which cases we want toasters? It seems like we're disabling tons of them, I wonder if opposite should be default?

Civolilah: @beganovich So, we are disabling the toaster for the company endpoint in cases when another "priority" endpoint is called on that page. This priority endpoint already includes great UX with its own toaster, so having the company endpoint toaster would potentially break the UX. In this specific case that you commented on, we're disabling the toaster when users drag and drop (dnd) gateways (prioritizing payments), so users can simply drag and drop while the details are saved in the background.

PR #2399, opened 2025-03-04
[Bugfix] Double-Querying

Civolilah:
@beganovich @turbo124 The PR includes fixes for double-querying caused by setting values from preferences after calling the query with default parameter values. Let me know your thoughts.

### DIFF
--- a/src/common/hooks/useDataTablePreferences.ts
+++ b/src/common/hooks/useDataTablePreferences.ts
@@ -173,9 +173,7 @@ export function useDataTablePreferences(params: Params) {
         setStatus(['active']);
       }
 
-      setTimeout(() => {
-        setArePreferencesApplied(true);
-      }, 50);
+      setArePreferencesApplied(true);
     }
   }, [isInitialConfiguration]);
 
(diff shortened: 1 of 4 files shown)


turbo124:
@Civolilah this is better, any idea why we are still seeing the POST to documents when viewing Client Overview?

image
image

Civolilah:
> @Civolilah this is better, any idea why we are still seeing the POST to documents when viewing Client Overview?
> 
> image image

@turbo124 I just found the reason for it. The `useTabs` hook called it to display the number of documents in the tab. We don't need this at all, because client.documents provides it for us. I've removed it - please let me know if you also no longer see this endpoint being called.

src/common/hooks/useDataTablePreferences.ts
```diff
@@ -168,6 +172,10 @@ export function useDataTablePreferences(params: Params) {
       } else {
         setStatus(['active']);
       }
+
+      setTimeout(() => {
+        setArePreferencesApplied(true);
```
beganovich, line 177: Do you mind explaining why are we using time-based checks here..? I am always super cautious when it comes to things like this because if timer is too low race condition can mess it up.

Thanks.

Civolilah: @beganovich Yes, I understand your point. The only reason I did this is to ensure that this state is updated as the last one, and all re-renderings are done before setting this state to true. But it will work great without this tiny timeout, I just removed it and tested. So, it has been removed!

beganovich (approved):

Issue #2393, opened 2025-03-03
uMatrix shows massive page links to Google


itsaw:
## Version 
v5.11.48

## Environment 
LXC

![Image](https://github.com/user-attachments/assets/cbc86bbe-3f30-410e-bdec-958423f9fc4b)

As shown in the screenshot below, my self-hosted instance is trying to communicate massively with Google.

We had already addressed this problem once with version 4, especially with regard to the fonts.
However, the fact that images and YouTube are now also appearing is a disaster for me as a data protection officer.

According to the GDPR, this is a massive infringement.

turbo124:
We will need more information....

Which pages are you seeing this on?

itsaw:
Here is a small picture gallery (Firefox private window, uMatrix)

Settings - Tax settings
![Image](https://github.com/user-attachments/assets/a50fc82c-5234-4b5c-b576-97b08544e0ba)

Looks like the video at 
![Image](https://github.com/user-attachments/assets/7245e56c-132f-40f9-98ad-454193a09557)
already starts access



But I also have to revise some things, it is apparently only the Youtube access due to the video

turbo124:
@beganovich my guess is that this will be tricky to stop preloading on?

beganovich:
Two cases here.

1. Requests to Google due to loading of Google Sign-in library
2. Embedding the tax guide video (from Google/YouTube). https://invoiceninja.github.io/en/taxes/

I guess we can get rid of YouTube by uploading guide to Vimeo, possibly, but that's still 3rd party request. Next step would be not embedding the thing at all, which IMHO would decrease UX. 

As a replacement, perhaps we can introduce privacy toggle which would configure some elements to be disabled, eg. Google log in, embedded videos, etc, on account basis.

itsaw:
Therefore, on the basis of the GDPR, no personal data may be transmitted to third-party providers unless consent has been obtained in advance.

Here I think that the corresponding switches, also for administrators by default, would probably be a pretty clean way.

In addition to the logins & videos, this also applies to the fonts, which should actually be hosted locally (enable upload?)

turbo124:
@beganovich we shouldn't be loading google signin for self hosters thou, right? I believe we closed that one off.

@itsaw, I'll update the documentation website to not use external resources, however the content will still be delivered from invoicing.co

In regards to other third party resources, the designs do pull in google fonts from their URLs. If this was an issue you would need to create custom designs that did not use fonts, or pulled from a preferred location.

[...]

PR #2400, opened 2025-03-04
Dynamically load Google module

beganovich:
This dynamically loads and imports Google login, resulting in less cookies being set.

https://github.com/invoiceninja/ui/issues/2393


turbo124:
closes #2393

masroor1967 (approved):

PR #2413, opened 2025-03-08
[Feature] Implementing Column Picker For "Transactions"

### LINKED ISSUE #2287: [Feature Request] Customize or add field to Transaction list
**What version of Invoice Ninja are you running? ie v4.5 / v5**
v5.11.7

**What environment are you running?**
Docker

**Have you searched existing issues/requests?**
Yes found none

**Screenshots**
image

**Additional context**
Added the field "Participant name" or give the ability to customize columns like in "Invoices" or many other places.
As it can be seen on my screenshot it is not easy to se what is what from this bank/gocardless as description is blank, only the sender or recipient is given in the "Participant name" field - So I need to edit all transactions to se and copy into description field to be able to reconcile transactions.
But best would be to customize columns as I personal would like to have account displayed also.

Civolilah:
@beganovich @turbo124 The PR includes the implementation of column picker as we have for many other tables for "Transactions".

Closes #2287 

Let me know your thoughts.


turbo124:
selecting date created  does nothing.

beganovich (approved):

PR #2420, opened 2025-03-10
[Bugfix] Notes Colors And Logic Fixes

Civolilah:
@beganovich @turbo124 The PR includes fixes for notes colors for both light and dark mode and a fix for the logic on client notes. Let me know your thoughts.

### DIFF
--- a/src/pages/clients/common/components/Settings.tsx
+++ b/src/pages/clients/common/components/Settings.tsx
@@ -41,7 +41,11 @@ export default function Settings() {
   const handleChange = (property: string, value: string | number | boolean) => {
     const $client = cloneDeep(client)!;
 
-    if (property !== 'currency_id' && value === '') {
+    if (
+      property !== 'currency_id' &&
+      $client.settings?.[property] !== undefined &&
+      value === ''
+    ) {
       delete $client.settings?.[property];
     } else {
       set($client, property, value);
--- a/src/pages/clients/common/hooks/useClientColumns.tsx
+++ b/src/pages/clients/common/hooks/useClientColumns.tsx
@@ -35,7 +35,7 @@ import {
   extractTextFromHTML,
   sanitizeHTML,
 } from '$app/common/helpers/html-string';
-import { useColorScheme } from '$app/common/colors';
+import classNames from 'classnames';
 
 export const defaultColumns: string[] = [
   'name',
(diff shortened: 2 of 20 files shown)


src/pages/clients/common/hooks/useClientColumns.tsx
```diff
@@ -326,11 +325,12 @@ export function useClientColumns() {
           tooltipElement={
             <div className="w-full max-h-48 overflow-auto whitespace-normal break-all">
               <article
-                className="prose prose-sm"
+                className={classNames('prose prose-sm', {
+                  'prose-invert': !reactSettings?.dark_mode,
```
Civolilah, line 329: @beganovich Color scheme cannot cover all cases here. When text is bolded, the best visibility is when the `prose-invert` class is used for example. But, we need to use reversed logic for tooltips, so on the light mode of the app tooltip is dark so the prose-invert class should be used, however on the dark mode of the app, tooltips are light so it should not be used. Only if we show this outside of the tooltip should we use the prose-invert class when the dark mode of the app is turned on.

src/pages/clients/common/components/Settings.tsx
```diff
@@ -41,7 +41,11 @@ export default function Settings() {
   const handleChange = (property: string, value: string | number | boolean) => {
     const $client = cloneDeep(client)!;
 
-    if (property !== 'currency_id' && value === '') {
+    if (
+      property !== 'currency_id' &&
+      $client.settings?.[property] !== undefined &&
+      value === ''
+    ) {
       delete $client.settings?.[property];
```
Civolilah, line 49: @beganovich I just took a look at this logic. I did not delete the old `AdditionalInfo` component and there is this logic of deleting property also. Is there any specific reason why the prop should be completely deleted from the client.settings object? I mean it is just changing the value...

turbo124: @Civolilah unsetting the prop is very important.

When a prop is SET at the group or client level it will override the company level prop.

If the prop does not exist, then it signifies that the property has not been set at the client/group level.

why is this important? At settings level you can have bool / null values as valid values, so there is no way to determine which property is "active" at this level.

turbo124:
@beganovich for merge to main please as priority after review

src/pages/clients/common/components/Settings.tsx
Civolilah: @turbo124 Ah okay, thanks for the explanation! I got it right now!

beganovich (approved):

PR #2425, opened 2025-03-11
[Bugfix] HTML Tags | Reports Preview

Civolilah:
@beganovich @turbo124 The PR includes fixes for "notes" columns on Report previews. So, instead of having text wrapped in those tags, we have extracted text now and displayed plain text. Let me know your thoughts.


turbo124:
@Civolilah 

we'll need to cover any input field which may be impacted.

Public Notes
Terms
Private Notes

Across all entities... invoices,quotes, credits,purchase orders, vendors,clients etc

Civolilah:
> @Civolilah
> 
> we'll need to cover any input field which may be impacted.
> 
> Public Notes Terms Private Notes
> 
> Across all entities... invoices,quotes, credits,purchase orders, vendors,clients etc

@turbo124 Ah okay, now all necessary props should be covered in this context. Let me know your thoughts.

beganovich (approved):

PR #2432, opened 2025-03-13
[Bugfix] Bluring Not Triggering When Dragging | Products Table

Civolilah:
@beganovich @turbo124 The PR includes fixes for input fields not changing value when the onBlur event isn't triggered before dragging in the products table. Let me know your thoughts.


turbo124:
Appears to correct the issue, however now i am seeing some flickering  after changing positions of a line with drag and drop. We'll want to ensure the implementation has not introduced another issue.

Civolilah:
> Appears to correct the issue, however now i am seeing some flickering after changing positions of a line with drag and drop. We'll want to ensure the implementation has not introduced another issue.

@turbo124 I totally understand your point and what you're talking about, but actually this implementation didn't cause that flickering that you mentioned. If you check the current production version, this small thing already happened. However, I made an improvement to the implementation and ensured nothing will go wrong. Let me know your thoughts.

turbo124:
@beganovich qa passes, if the flickering on drag and drop on the number input is normal, this is good to go.

beganovich (approved):
Thanks :+1:

PR #2455, opened 2025-03-24
Reports - Projects

turbo124:
Add project reports into the UI.

@beganovich, this required some changes to how we handle different file types, previously we only supported CSV downloads, in this PR i've added the ability to inspect the payload to differentiate between PDF / CSV.


beganovich (approved):
Thanks :+1:

PR #2460, opened 2025-03-26
[Cleanup] Data Table Actions Dropdown Redesign

Civolilah:
@beganovich @turbo124 The PR includes redesigning the action buttons on tables. Instead of buttons, there will be an interactive "dots" icon with changes to the popover, including updates to border radius, dividers, border colors, etc. Screenshot:

Screenshot 2025-03-27 at 16 48 29

Let me know your thoughts.


Civolilah:
@turbo124 Currently, I am leaving the icons that we already have applied to each action in the app, so just the styling of the dropdown has been improved per design. The PR would be huge if we changed each icon in one PR, so step by step, as we work on any bug or functionality, I will change the icons.

Also, I am leaving the accent color here. Please let me know if you want to switch to the color of icons from the Figma.

turbo124:
@Civolilah we discussed we will not change from the Actions menu button to the three dots.

beganovich:
Should we close this one?

Civolilah:
> Should we close this one?

@beganovich @turbo124 Actually, the designer made some small changes within the popover of the dropdown menu, on elements there, etc. I like those small changes, and even though we don't want the dots icon, I kept those additional small changes. So, I've removed the unnecessary icon and left just those changes. Let me know your thoughts.

src/components/datatables/Actions.tsx
```diff
@@ -359,6 +359,7 @@ export function Actions(props: Props) {
       border: `1px solid ${colors.$19}`,
       zIndex: 10,
       width: '16rem',
+      boxShadow: 'none',
```
Civolilah, line 362: @beganovich Just a small missed fix for the unnecessary box shadow on the menu.

beganovich (approved):

PR #2466, opened 2025-03-31
[Cleanup] General Table Styling Adjustment

Civolilah:
@beganovich @turbo124 The PR includes final styling of the data table along with adjusting the border color on the dashboard to match Figma. Screenshots:

Screenshot 2025-04-01 at 19 18 22

Screenshot 2025-04-01 at 19 19 04

Let me know your thoughts.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -123,7 +123,7 @@ export function App() {
   };
 
   useEffect(() => {
-    document.body.style.backgroundColor = colorScheme.$2;
+    document.body.style.backgroundColor = colorScheme.$23;
     document.body.style.colorScheme = colorScheme.$0;
   }, [colorScheme]);
 
--- a/src/common/colors.ts
+++ b/src/common/colors.ts
@@ -52,6 +52,8 @@ export const $1 = {
   $20: '#323236', // Dropdown element hover background color
   $21: '#1f2e41', // Divider color
   $22: '#a1a1aa', // Label color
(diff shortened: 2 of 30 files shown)


turbo124:
this looks good to me, @beganovich could you run your eyes over it also please?

beganovich:
Couple issues with this:

1. Not related to this PR, but background colour doesn't seem to match, creating issues with contrast. 
2. Contrast is very low. I've checked colours, seems to match, but the problem is font weight. Make sure to match that both in table PRs, and please fix it for the badges too.
3. Table roundness.

Civolilah:
> Couple issues with this:
> 
> 1. Not related to this PR, but background colour doesn't seem to match, creating issues with contrast.
> 2. Contrast is very low. I've checked colours, seems to match, but the problem is font weight. Make sure to match that both in table PRs, and please fix it for the badges too.
> 3. Table roundness.

@beganovich Thanks for the advice. Yes, the background color did not match, so I made the color match and it improved contrast. But between background color fields and table, the issue was the wrong border color. I checked Figma and found that the border color was wrong (it just needed to be a little bit darker.), so I've fixed it also. Table roundness is also fixed. Let me know if I fixed anything.

beganovich (approved):
Looks good, thanks :+1: 

I'd imagine we might get some users to complain about contrast and muted/pastel colors, but let's roll and see if they like it.

PR #2482, opened 2025-04-10
[Cleanup] "Quick Actions" Modal Redesign

Civolilah:
@beganovich @turbo124 The PR includes a redesign of the quick actions modal according to Figma's design. Screenshot:

Screenshot 2025-04-10 at 02 12 36

`Note`: We are missing the "edit_quick_actions" translation keyword in the files. If you agree, please add it to the files.

Let me know your thoughts.


turbo124:
@Civolilah this looks really good, the only change i would suggest here is making the entire label also draggable, otherwise the user only has a small section where this can be activated.

image


also, please revert the translation to the existing one.

Civolilah:
> @Civolilah this looks really good, the only change i would suggest here is making the entire label also draggable, otherwise the user only has a small section where this can be activated.
> 
> image
> 
> also, please revert the translation to the existing one.

@turbo124 Yes, I agree with you. The logic for drag-and-drop is fixed on the "Quick Actions" modal as well as on the "Edit Columns" modal since we have very similar UX in both places, making the whole labels draggable. Also, the old label has been restored as the title for the "Quick Actions" modal. Let me know your thoughts.

turbo124:
@Civolilah 

this looks good. just a query, in my environment, when i now click to edit an invoice, i am seeing the invoice slider also being presented, unsure if this is just an aberration with dev build, or edge case when click on the invoice number from the list view.

Civolilah:
> @Civolilah
> 
> this looks good. just a query, in my environment, when i now click to edit an invoice, i am seeing the invoice slider also being presented, unsure if this is just an aberration with dev build, or edge case when click on the invoice number from the list view.

@turbo124 I tested it and couldn't recreate the issue, but I've added a small logic change that ensures the invoice slider doesn't open when a link is clicked, instead navigating directly to the invoice edit page. I tested with plain text (no link) and when I clicked on the plain text, the invoice slider didn't open. Let me know your thoughts.

PR #2486, opened 2025-04-13
[Cleanup] "Online Payments" Table Redesign

Civolilah:
@beganovich @turbo124 The PR includes a redesign of the "Online Payments" table using the new Figma design, as this table is custom-created due to the dnd system implementation. Screenshot:

Screenshot 2025-04-13 at 19 22 51

Let me know your thoughts.


beganovich (approved):
Looks good, thanks :+1: 

@turbo124 I wonder if we should drag-n-drop whole row or use that handle on the right exclusively (this is how we do it on invoices page)?

Civolilah:
> Looks good, thanks 👍
> 
> @turbo124 I wonder if we should drag-n-drop whole row or use that handle on the right exclusively (this is how we do it on invoices page)?

@beganovich Actually, the whole row is draggable. I believe that the icon is just there to indicate that dnd functionality is available. Also, at the top right corner there is an icon with a tooltip to additionally explain that dragging actually prioritizes online payments.

beganovich:
Yeh, it was a question should we do it or keep it as invoices table.

Civolilah:
@turbo124 @beganovich The approach for dragging has been changed here. Now, dragging is only available using the icon on the right side. Let me know your thoughts.

beganovich (approved):
Thanks :+1:

PR #2488, opened 2025-04-13
Improve TinyMCE functionality

turbo124:
Improves toolbar layout + functionality for font/font size selections.


beganovich (approved):
Thanks :+1:

PR #2492, opened 2025-04-14
[Cleanup] Combobox Redesign

Civolilah:
@beganovich @turbo124 The PR includes a redesign of the combobox in general according to the Figma design. Screenshots:

Screenshot 2025-04-15 at 00 33 05

Screenshot 2025-04-15 at 00 33 17

Let me know your thoughts.

### DIFF
--- a/src/components/forms/Combobox.tsx
+++ b/src/components/forms/Combobox.tsx
@@ -429,18 +429,22 @@ export function Combobox<T = any>({
                   hoverColor: colors.$20,
                 }}
                 key={option.id}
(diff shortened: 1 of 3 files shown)


turbo124:
@Civolilah 

Can we ensure - especially for the client combo - that the result container spans the full width, the previous iteration looked like this:

image

this PR truncates..

image

Civolilah:
> @Civolilah
> 
> Can we ensure - especially for the client combo - that the result container spans the full width, the previous iteration looked like this:
> 
> image
> 
> this PR truncates..
> 
> image

@turbo124 Sure thing, so the logic is adjusted and the combo only has a minimum width but not a maximum width. It will depend on how long the label is. Let me know your thoughts.

beganovich (approved):
Thanks :+1:

PR #2499, opened 2025-04-17
[Cleanup] Line Item Product Table Redesign

Civolilah:
@beganovich @turbo124 The PR includes line item product table redesign according to Figma design. Screenshot:

Screenshot 2025-04-17 at 18 40 21

Let me know your thoughts.

### DIFF
--- a/src/components/forms/Combobox.tsx
+++ b/src/components/forms/Combobox.tsx
@@ -342,7 +342,11 @@ export function Combobox<T = any>({
         </p>
       ) : null}
 
-      <div className="relative mt-1">
+      <div
+        className={classNames('relative', {
+          'mt-1': inputOptions.label,
+        })}
+      >
         <div
           className={classNames(
             'relative w-full cursor-default overflow-hidden rounded-md border text-left sm:text-sm',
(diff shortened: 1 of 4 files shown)


turbo124:
Dropdowns are not working, i cannot see the content from them:

image

Civolilah:
> Dropdowns are not working, i cannot see the content from them:
> 
> image

@turbo124 The conflict in the customization of the select field. The issue has been fixed. Let me know your thoughts.

turbo124:
image


We may have to rethink the overall logic here. As you can see, with multiple custom fields the line items quickly become unmanageable.

We _may_ need to entertain so method that allows us onFocus() to present the field in its full width, either via Modal - or - by dynamically resizing..

@beganovich do you have any input on this?

beganovich:
Two ways I'd go about this.

a) multi-line table rows (cards)
b) horizontally scrollable table with line total and x pinned.

image

We may need to reconstruct the tables a bit for option b, but we have to do same for a.

Other than that, can't do much, this will be always issue when too many columns.

PR #2500, opened 2025-04-18
[Bugfix] Line Items Table Fields Width

Civolilah:
@beganovich @turbo124 The PR includes adding more width specifications for fields in the line item table to avoid issues with widths there. Let me know your thoughts.


beganovich (approved):
Thanks :+1:

PR #2507, opened 2025-04-24
[Cleanup] Client Selector && Two Field Cards Redesign

Civolilah:
@beganovich @turbo124 The PR includes a redesign of the top of the invoice page according to the Figma design. Screenshot:

Screenshot 2025-04-24 at 18 24 17

Let me know your thoughts.

### DIFF
--- a/src/components/CopyToClipBoardIconOnly.tsx
+++ b/src/components/CopyToClipBoardIconOnly.tsx
@@ -10,7 +10,8 @@
 
 import { toast } from '$app/common/helpers/toast/toast';
 import { MouseEvent } from 'react';
-import { MdOutlineContentCopy } from 'react-icons/md';
+import { CopyToClipboard } from './icons/CopyToClipboard';
+import { useColorScheme } from '$app/common/colors';
 
 interface Props {
   text: string;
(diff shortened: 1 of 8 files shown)


Civolilah:
> Couple of issues here:
> 
> image
> 
> I am not sure about the change in vertical heights of cards matching... in the new design.. lots of blank space.
> 
> We need to add back the copy link icon functionality
> 
> image
> 
> Which does brings up the last issue if there are many contacts on an invoice, the vertical height of that card is excessive.
> 
> image
> 
> I think we should apply css flex on the email / view portal / copy link icons so that we can save some space along here, @beganovich thoughts?

@turbo124 The UI here has been adjusted per your instructions. The issue with not having a copy icon is actually what I mentioned on Slack (https://invoiceninja.slack.com/archives/C05CZU50PU5/p1745510613429089), so the check is added there also. Screenshot:

Screenshot 2025-04-25 at 16 20 39

Let me know your thoughts.

beganovich:
Perhaps we can save a ton of space by not showing email always? How about ℹ️ which on hover displays email? I doubt it's useful to be shown all the time. 

Let's see what Dave says.

turbo124:
I think we can save some space by changing to a grid like this:

image

email can be on the same line as contact name, but with truncate css to take care of any x overflow

Civolilah:
> I think we can save some space by changing to a grid like this:
> 
> image
> 
> email can be on the same line as contact name, but with truncate css to take care of any x overflow

@turbo124 Sure thing. The UI has been redesigned a little bit here not to be so long and to make a little bit more space. Screenshot:

Screenshot 2025-04-26 at 19 02 37

Let me know your thoughts.

beganovich (approved):

PR #2510, opened 2025-04-25
[Bugfix] Removing "View In Portal" Action When Link Is Not Available

Civolilah:
@beganovich @turbo124 The PR includes fixes on existing develop for removing unnecessary `view_in_portal` action on create pages because the link is not available there. Let me know your thoughts.


beganovich (approved):
Thanks :+1:

PR #2523, opened 2025-05-04
Fixes for current hours formatting

### LINKED ISSUE #2522: Project total hours incorrect total hours calculation
Appears as if the total hours is being rounded to the nearest hour, but it shows to 2dp. which means that it's actually showing the incorrect number of hours. 

Should display 30.25 instead of 30.00
Image

Image

This one is in the app
Image

turbo124:
closes #2522


beganovich (approved):
Thanks :+1:

PR #2529, opened 2025-05-06
[Cleanup] Payment Pages Redesign

Civolilah:
@beganovich @turbo124 The PR includes a redesign of Payment pages according to the design. Screenshot:

Screenshot 2025-05-07 at 18 54 29

Let me know your thoughts.

### DIFF
--- a/src/components/CountrySelector.tsx
+++ b/src/components/CountrySelector.tsx
@@ -9,7 +9,7 @@
  */
 
 import { useCountries } from '$app/common/hooks/useCountries';
-import { SearchableSelect } from './SearchableSelect';
+import { SelectField } from './forms';
 
 export interface GenericSelectorProps<T = string> {
   value: T;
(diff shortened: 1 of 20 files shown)


turbo124:
image


thanks, can we realign this field please.

Civolilah:
> image
> 
> thanks, can we realign this field please.

@turbo124 Yes, I agree with you that this part of the page is a little bit weird in terms of design. I played with this area a little bit, and this is my suggestion. Screenshot:

Screenshot 2025-05-08 at 14 07 27

Let me know your thoughts.

turbo124:
what does dark mode look like? also unsure about the background color of the card

Civolilah:
> what does dark mode look like? also unsure about the background color of the card

@turbo124 If you would like to keep the background of cards the same as the payment main card, I would like to suggest a very similar approach to what the designer made on sliders. Screenshot:

Screenshot 2025-05-09 at 16 35 15

Let me know your thoughts.

beganovich (approved):
Thanks :+1:

PR #2538, opened 2025-05-13
[Cleanup] Adding Client Contacts UX Improvement

Civolilah:
@beganovich @turbo124 The PR includes adding smooth scroll after the add_contact button is clicked to avoid manual user scrolling if there are more than 2 contacts added. Video:

https://github.com/user-attachments/assets/5ea2f9a2-0c0b-47e3-85e9-af93dd4c5132

Let me know your thoughts.


beganovich (approved):
Thanks :+1:

PR #2540, opened 2025-05-14
feat: added a net amount column to the quotes overview

Lars-:
This PR is related to this discussion: https://github.com/invoiceninja/invoiceninja/discussions/10954#discussioncomment-13081588


turbo124:
@beganovich thoughts on calculated fields in list views? 

performance impacts on large lists?

beganovich:
@turbo124 looks good to me, I don't think it'll affect performance too much. I think we already have some "calculated" fields with names (merging first and last name).

PR #2545, opened 2025-05-15
[Cleanup] Transaction Pages Redesign

Civolilah:
work in progress...

### DIFF
--- a/src/components/TabGroup.tsx
+++ b/src/components/TabGroup.tsx
@@ -9,7 +9,13 @@
  */
 
 import classNames from 'classnames';
-import React, { ReactElement, ReactNode, useEffect, useState } from 'react';
+import React, {
+  CSSProperties,
+  ReactElement,
+  ReactNode,
+  useEffect,
+  useState,
+} from 'react';
 import { useColorScheme } from '$app/common/colors';
 import styled from 'styled-components';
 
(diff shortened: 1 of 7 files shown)


turbo124:
@Civolilah 

I find the slider is not as responsive now when clicking on a row to make it slide out. 

We may need to align the logic on this to match other lists. 

image

Civolilah:
> when

@turbo124 Yes, very good catch. It was a quick fix. Let me know your thoughts.

src/pages/settings/expense-categories/components/CreateExpenseCategoryModal.tsx
```diff
@@ -100,7 +100,10 @@ export function CreateExpenseCategoryModal(props: Props) {
       onClose={() => props.setVisible(false)}
       stopPropagationInHeader
     >
-      <div onClick={(event) => event.stopPropagation()}>
+      <div
+        className="flex flex-col space-y-4"
+        onClick={(event) => event.stopPropagation()}
```
beganovich, line 105: What are we stopping here? Div shouldn't have any default action attached?

beganovich (approved):
Looks good, please see comment :+1:

src/pages/settings/expense-categories/components/CreateExpenseCategoryModal.tsx
Civolilah: @beganovich Yes, we can move this to RouterLink component and avoid having a div here. The purpose of stopping propagation is to prevent unnecessary actions such as opening the slider.

beganovich (approved):

PR #2546, opened 2025-05-15
Had trials in app

turbo124:
Refactor to allow in app trials, rather than pushing to CP.

Introduces some new props to help determine eligibility for a trial and provides the correct CTAs.

Also introduces accepting a CC during the trial signup phase.

All internal Manage Plan links now point to the /settings/account_management route


beganovich (approved):
Looks good :+1: 

We're missing couple translation strings, eg.

```
At the end of your 14 day trial your card will be charged $12/month. Cancel anytime.
```

But if you think that's OK; feel free to merge.

PR #2561, opened 2025-05-25
Plan expiry update

turbo124:
pushes the expiry date to the end of day to ensure plan are not prematurely marked as expired.


beganovich (approved):
Thanks :+1:

PR #2565, opened 2025-05-27
[Bugfix] Account Management Route

Civolilah:
@beganovich @turbo124 The PR includes fixes for condition on account management route. Let me know your thoughts.


beganovich (approved):
Thanks :+1:

PR #2575, opened 2025-06-04
Add net amount column invoices recurring invoices

Lars-:
- Add net_amount column to default columns in both invoice and recurring invoice views
- Display amount minus taxes for better visibility of net values

This PR is related to this discussion: https://github.com/invoiceninja/invoiceninja/discussions/10954#discussioncomment-13081588


turbo124:
This looks to make the net_amount a default column on the table, It would be preferable to make it optional.

Lars-:
@turbo124 Done!

beganovich (approved):
Thanks :+1:

PR #2578, opened 2025-06-05
[Bugfix] Overflow Issue For Dashboard Tables

### LINKED ISSUE #2576: Overflow for overdue invoices
If you have too many overdue invoices, on the dashboard, it doenst make a scrollable box it just keeps overflowing down the page.

Civolilah:
@beganovich @turbo124 I was able to recreate this issue once in production and once locally, but after that I was not able to recreate it and detect the reason for it. However, I noticed unnecessary logic for height calculation there, so I removed it and ensured that the height is fixed and overflow is impossible. 

Closes #2576 

Let me know your thoughts.


beganovich (approved):
Thanks :+1:

PR #2582, opened 2025-06-06
[Bugfix] Datatable Performance && Add Tasks  On Invoice Action Fix

Civolilah:
@beganovich @turbo124 The PR includes performance improvements for re-rendering rows in the Data Table when it is displayed, along with small UX improvements when data is loading. Let me know your thoughts.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -19,7 +19,7 @@ import { RootState } from './common/stores/store';
 import dayjs from 'dayjs';
 import { useResolveDayJSLocale } from './common/hooks/useResolveDayJSLocale';
 import { useResolveAntdLocale } from './common/hooks/useResolveAntdLocale';
-import { atom, useAtomValue, useSetAtom } from 'jotai';
+import { atom, useSetAtom } from 'jotai';
 import { useSwitchToCompanySettings } from './common/hooks/useSwitchToCompanySettings';
 import { useLocation, useNavigate, useParams } from 'react-router-dom';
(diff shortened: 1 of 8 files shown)


beganovich (commented):
Looks good :+1: 

Would you be kind and just make a list of changes that were made? Just a bit easier to understand.

Civolilah:
> Looks good 👍
> 
> Would you be kind and just make a list of changes that were made? Just a bit easier to understand.

@beganovich Sure thing!

Changes that were made:

- Changed the logic of the useColorScheme atom to avoid additional re-rendering because we had an atom there
- Along with the useColorScheme hook change, I adjusted what is in the App.tsx component
- We also had unnecessary re-rendering under the useDataTablePreferences hook because of updating Redux state, this is avoided by updating state on unmount of this hook/component
- Under DataTable, I implemented Tr memoization with a condition that the resource in that row needs to change to trigger re-rendering (used isEqual)
- Created logic to only re-render checkboxes when rows are being selected
- Created logic for opacity indication that shows rows are rendering
- Created memoization for selectedResources according to selected ID changes
- Avoided using refs for checkboxes under DataTable
- Added strokeWidth prop to Checkbox component
- Added onClick functionality to Th component to make it possible to check the checkbox by clicking anywhere in the th
- Resolved issue with AddTasksOnInvoice action (issue with selecting setSelected at an incorrect moment)

beganovich (approved):
Thanks :+1:

PR #2584, opened 2025-06-11
[Cleanup] Double Requesting | Function Reference | Recurring Invoices

Civolilah:
@beganovich @turbo124 The PR includes updated logic to avoid creating unnecessary references for the save function on Recurring Invoices. Let me know your thoughts.

### DIFF
--- a/src/pages/recurring-invoices/RecurringInvoice.tsx
+++ b/src/pages/recurring-invoices/RecurringInvoice.tsx
@@ -18,7 +18,7 @@ import { ResourceActions } from '$app/components/ResourceActions';
 import { Spinner } from '$app/components/Spinner';
 import { useAtom, useAtomValue } from 'jotai';
 import { cloneDeep } from 'lodash';
-import { useEffect, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Outlet, useParams } from 'react-router-dom';
 import { v4 } from 'uuid';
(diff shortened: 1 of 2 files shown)


turbo124:
Prior to merging this, we'll want to test the disable save button functionality _during_ a request and ensure the root cause here is not the ability for the end user to double tap the save button.

Civolilah:
> Prior to merging this, we'll want to test the disable save button functionality _during_ a request and ensure the root cause here is not the ability for the end user to double tap the save button.

@turbo124 Done!

turbo124:
@Civolilah I thought we were only going to implement isFormBusy() and not the useCallBack() at this stage to isolate the correct fix for this issue?

Civolilah:
> @Civolilah I thought we were only going to implement isFormBusy() and not the useCallBack() at this stage to isolate the correct fix for this issue?

@turbo124 I do not think that can be a solution if the issue is what Claude reported. The `useCallback` is used for updating function references only when needed, but `isFormBusy` is used when the save/update functions are called to disable the button. So, these serve two different purposes here: function reference management and button disabling during API calls.

Civolilah:
> @Civolilah sure, I understand the different functionalities here.
> 
> But, I think we still want to isolate each update here so we understand the exact root cause of duplicate request issue.
> 
> ie. double click vs additional references misbehaving.

@turbo124 Okay, sure thing. So, I've removed useCallBack and keep isFormBusy for first iteration. Please let me know if it still works as expected.

turbo124:
@Civolilah this appears to work for the create page, but it also appears that we need to deploy this across every single onSave() action including Edit etc. Can we please update this pr and deploy across all entities please.

Civolilah:
> @Civolilah this appears to work for the create page, but it also appears that we need to deploy this across every single onSave() action including Edit etc. Can we please update this pr and deploy across all entities please.

@turbo124 I tried to find as many places as possible where the "busy form" functionality is missing, which means that we are disabling the triggered save button until the call finishes. Let me know your thoughts.

PR #2587, opened 2025-06-11
Updates for reports to support item filters

turbo124:
Adds support for filtering all item reports by product keys


turbo124:
@beganovich after review could we please push this to main?

PR #2597, opened 2025-06-17
[Feature] Adding "Draft" Filter Option | Recurring Invoices

### LINKED ISSUE #2586: Bug/Missing filter options in status dropdown
Hi,

After the new REACT UI in v5.12.0 there is missing statuses from the filtering dropdowns...
I think I have seen the same issue mulitple places but not sure where at the moment, (so believe that it is something that goes through all status filtering dropdowns), Ex. here in recurring invoices... Draft is not an option.

Image

Used if I don't what to se paused and completed but all that will be future invoicing.

Civolilah:
@beganovich @turbo124 The PR includes adding additional "Draft" filter option for Recurring Invoices. Screenshot:

Screenshot 2025-06-17 at 21 03 42

Closes #2586 

Let me know your thoughts.

### DIFF
--- a/src/pages/recurring-invoices/common/hooks.tsx
+++ b/src/pages/recurring-invoices/common/hooks.tsx
@@ -925,6 +925,12 @@ export function useRecurringInvoiceFilters() {
 
   const filters: SelectOption[] = [
     {
+      label: t('draft'),
+      value: 'draft',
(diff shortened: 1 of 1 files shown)


beganovich (approved):
Thanks :+1:

PR #2601, opened 2025-06-18
[Bugfix] Keys Conflict On DataTable (Duplicating Columns On Quotes Table)

Civolilah:
@beganovich @turbo124 The PR includes fixes for conflicts with keys in the data table. Let me know your thoughts.

### DIFF
--- a/src/components/DataTable.tsx
+++ b/src/components/DataTable.tsx
@@ -571,7 +571,21 @@ export function DataTable<T extends object>(props: Props<T>) {
       setCurrentPage(1);
     }
 
-    setAreRowsRendered(true);
+    if (perPage === '10') {
+      setAreRowsRendered(true);
+    }
+
+    if (perPage === '50') {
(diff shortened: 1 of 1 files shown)


src/components/DataTable.tsx
```diff
@@ -571,7 +571,21 @@ export function DataTable<T extends object>(props: Props<T>) {
...
+    if (perPage === '50') {
+      setTimeout(() => {
+        setAreRowsRendered(true);
+      }, 50);
+    }
+
+    if (perPage === '100') {
+      setTimeout(() => {
+        setAreRowsRendered(true);
+      }, 150);
```
Civolilah, line 587: @beganovich Very small timeouts are added just for the purpose of UX improvement. I removed them in one of the previous versions, but I see right now that we will need just a small one. So, basically the time that is taken for rendering table rows - it is only needed when per page is 50 or 100.

PR #2602, opened 2025-06-18
[Bugfix] Adding Additional Memoization Parameter For Correct Updating

Civolilah:
@beganovich @turbo124 The PR includes adding an additional memo value for comparison to ensure correct updates when columns are changing. Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #2604, opened 2025-06-21
[Cleanup] UX Improvement When Selecting Permissions

Civolilah:
@beganovich @turbo124 The PR includes few UX improvements when selecting permissions:

- If the "Admin" permission is selected, all other permissions will appear "checked" and "disabled" without the possibility of unchecking them until the "Admin" permission is unchecked.
- If the "Admin" permission is checked, all permissions will be removed from the array, and only the `is_admin` boolean value will be set to true. The UI will be correctly updated.
- If 13/14 permissions from one type (for example, "View") are checked, and then the last one is checked, it will remove all of them from the array, and only the "view_all" permission will be added along with a proper UI update.

Let me know your thoughts.

### DIFF
--- a/src/pages/settings/users/edit/components/Permissions.tsx
+++ b/src/pages/settings/users/edit/components/Permissions.tsx
@@ -46,6 +46,14 @@ export function Permissions(props: Props) {
     'recurring_expense',
   ];
 
+  const isPermissionDisabled = () => {
+    if (user?.company_user?.is_admin) {
+      return true;
+    }
+
+    return false;
+  };
+
   const handleAdministratorToggle = (value: boolean) => {
     setUser(
       (user) =>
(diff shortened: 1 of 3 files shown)


src/pages/settings/users/edit/components/Permissions.tsx
```diff
@@ -119,6 +132,28 @@ export function Permissions(props: Props) {
...
+          currentPermission !== 'disable_emails'
+      ).length;
+
+      if (permissionsFromTypeLength === permissions.length) {
+        currentPermissions = currentPermissions.filter(
+          (currentPermission) =>
+            !currentPermission.startsWith(permissionType) ||
+            currentPermission === 'view_reports' ||
+            currentPermission === 'view_dashboard' ||
+            currentPermission === 'disable_emails'
```
beganovich, line 150: These seems to repeat? Do you think it's worth extracting into separate array so we don't have to update both places?

beganovich (approved):
Looks good, just tiny clarification :+1:

beganovich (approved):
Thanks :+1:

PR #2607, opened 2025-06-23
[Bugfix] Automatic Scroll When Live Preview Is Triggered 

Civolilah:
@beganovich @turbo124 The PR improves the logic to not have automatic scrolling when changes are made in an entity and live_preview should be triggered. Let me know your thoughts.


beganovich (approved):
Thanks :+1:

PR #2611, opened 2025-06-26
[Bugfix] Product Selector Filtering | Reports

Civolilah:
@beganovich @turbo124 The PR includes product selector filtering fixes along with a little UX improvement. Let me know your thoughts.


Civolilah:
> @beganovich for merge to main please after review.

@turbo124 I just tested production and the researching issue does not exist there. The only issue that exists on main is the issue with comma-separated values. So, I will create a separate PR for fixes on main right now. Later, when we sync develop and main, the solution will be the same because on develop it is a redesigned selector.

PR #2612, opened 2025-06-26
[Bugfix] Cost Field | Products

Civolilah:
@beganovich @turbo124 The PR includes transforming the "Cost" field for Products into a decimal number input field. Let me know your thoughts.


turbo124:
@Civolilah this still is not working, I believe the fix is just to use the display value rather than attempting to parse it?

PR #2618, opened 2025-06-27
Change status for sending einvoice

turbo124:


turbo124:
@beganovich for merge to main please after review

PR #2621, opened 2025-06-28
[Feature] Default Item Tax Rates

Civolilah:
@beganovich @turbo124 The PR includes implementation of default item tax rates functionality. Screenshot:

Screenshot 2025-06-28 at 15 18 43

`Note`: I've used "default_item_tax_rate" on the Tax Settings page. If you agree with it, please just add it to the translation files.

Let me know your thoughts.

### DIFF
--- a/src/pages/invoices/common/hooks/useResolveInputField.tsx
+++ b/src/pages/invoices/common/hooks/useResolveInputField.tsx
@@ -1,5 +1,3 @@
- 
-
 /**
  * Invoice Ninja (https://invoiceninja.com).
  *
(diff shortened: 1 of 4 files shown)


Civolilah:
@turbo124 The functionality is only available for company-level settings. Please let me know if this should be available for group/client-level settings.

turbo124:
@Civolilah 

This functionality should only be enforced if NO invoice level taxes are implemented... otherwise we may see double taxes like this:

image

Civolilah:
> @Civolilah
> 
> This functionality should only be enforced if NO invoice level taxes are implemented... otherwise we may see double taxes like this:
> 
> image

@turbo124 Yes, I found the issue and fixed it. Just one important confirmation: while selecting a product in the line item, the tax will still be applied to the line item if it is populated on the product level. That is actually the logic from current production. If there is anything that needs to be adjusted, let me know.

PR #2625, opened 2025-07-02
[Cleanup] Updated At Refresh Endpoint Query Parameter

Civolilah:
@beganovich @turbo124 The PR includes adding an `updated_at` query parameter to each `/refresh` endpoint in the app with the current timestamp as the value. Let me know your thoughts.


turbo124:
Looks good. @beganovich could we please merge to main after review.

PR #2629, opened 2025-07-04
Improve Live Preview logic

turbo124:
Live PR attempts to mitigate chromes native PDF view from intercepting focus inappropriately.

uses debounce logic on mousedown and keydown events in combination with debounce on resource changes AND intersection observer logic on the preview window itself.


turbo124:
@beganovich for merge to main please. we'll want to  ensure we can roll this one back in case of any undetected edge cases.

PR #2632, opened 2025-07-07
Remove Client Portal links for view / create users

turbo124:
Removes access to client portal links for view/create permission users.


turbo124:
@beganovich for merge to main please after review.

PR #2644, opened 2025-08-05
Allow tab to be shown when paid or partially paid

turbo124:
Shows the payment tab for paid AND partially paid.


beganovich (approved):
Thanks 👍

PR #2649, opened 2025-08-09
[Bugfix] Task Timer

### LINKED ISSUE #2489: Task row timer clock drift (on hidden window / when the browser suspends animations?)
I run `app.invoicing.co` in GNOME Web / Epiphany 47.4, as a web app, on Fedora 41.

However, I noticed that when I am inside a task (i.e. `https://app.invoicing.co/#/tasks/some_hash/edit`) and have the timer running, the timecodes drift compared to the system clock, even without suspending/resuming the computer; simply by me switching to another workspace or minimizing the window, so it's not the same as my previously reported issue #1744.

Demonstration video where you can see that the InvoiceNinja timer drifts / lags compared to a normal stopwatch (`gnome-clocks`) that measures the system time:

https://github.com/user-attachments/assets/4b75758e-e7fa-4da8-a3fd-5c67d628d702

As you can see here, my task timer indicates `03:08:00` when I focus out to another workspace, and when I come back exactly 60 seconds later, it indicates `03:08:33` (approximatively), and when I refresh the page a few moments later, it jumps from `03:08:42` to `03:09:27`...

This was observed mainly on the bottom row timer. The big timer at the top does not seem to be affected, somehow (but maybe I didn't look properly).

---

Note that recent versions of WebKitGTK [stop the animation clock](https://bugs.webkit.org/show_bug.cgi?id=285167) (to save power) when the window is not visible, so I don't know if that might be related. If so, maybe part of solving this durably would be handling the visibility change events as suggested in https://github.com/invoiceninja/ui/issues/1871#issuecomment-2781702456 to actually stop doing any CPU wakeups as long as the window is not visible, and resynchronize everything once it becomes visible again :)

Civolilah:
@beganovich @turbo124 The PR includes a fix for the issue with task timers on the time logs table on the task edit page. The issue was related to recalculating time when the user refocuses the tab of our app or returns from another screen.

Closes #2489 

Let me know your thoughts.

### DIFF
--- a/src/pages/tasks/common/components/DurationClock.tsx
+++ b/src/pages/tasks/common/components/DurationClock.tsx
@@ -8,7 +8,7 @@
  * @license https://www.elastic.co/licensing/elastic-license
  */
 
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { isTaskRunning } from '../helpers/calculate-entity-state';
 import { Task } from '$app/common/interfaces/task';
 import { PauseCircle } from 'react-feather';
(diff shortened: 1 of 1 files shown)


beganovich (approved):
Thanks 👍

PR #2654, opened 2025-08-13
Client statement permissions

turbo124:
Provide email statement option for clients with Edit Client permissions.


beganovich (approved):
Thanks 👍

PR #2655, opened 2025-08-13
Update user translation to Assigned user

### LINKED ISSUE #2653: "User" column title is unclear (can be mistaken for "Assigned user", and mismatches the rest of the UI)
As mentioned in #2652, the "User" column in tasks list views lacks precision in its naming.

For a while I thought the UI was broken, until I sent a support request that said this:

> When visiting https://app.invoicing.co/?#/tasks and enabling the "User" column, tasks that I (re)assigned to another user in my Enterprise plan all show up as being my user in the listing table's column, even though the user is successfully set within the task

The reply I got was:

> In the column section there are two user options:
> 
> 1. User (the user who created the records)
> 2. Assigned User (the user who was assigned to the task)
> 
> It sounds like you'll want to enable the assigned user column rather than the user column

That is a great reply that solved my immediate need, but from a UX standpoint I would highly suggest you rename the "User" column (in the table headers and in the columns customizer) to be more specific. Everywhere in the tasks editing UI the assigned user field is just called "User".

I think "Task creator" would be better terminology to use in the columns-related UIs for that field, and in the editing UI you might want to replace "User" by "Assigned user". Be specific and consistent everywhere.

turbo124:
closes #2653

### DIFF
--- a/src/pages/clients/edit/components/Details.tsx
+++ b/src/pages/clients/edit/components/Details.tsx
@@ -102,7 +102,7 @@ export function Details(props: Props) {
       )}
 
       {users && (
-        <Element leftSide={t('user')}>
+        <Element leftSide={t('assigned_user')}>
           <SelectField
             value={props.client?.assigned_user_id}
             onValueChange={(value) => handleChange('assigned_user_id', value)}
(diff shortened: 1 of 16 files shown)


beganovich (approved):
Thanks 👍

PR #2656, opened 2025-08-13
Add user,assigned_user includes for project tasks

### LINKED ISSUE #2652: "User" and "Assigned user" columns' values are blank or incorrect when viewed from within a Client's or Project's tasks list view
With the v5 SaaS Enterprise version on app.invoicing.co, I am observing the disparity below.

This is correct, it shows the user (which should be named "Creator" or "Task creator" for clarity?) and the "Assigned user":

Image

…but if you go to the "Projects" section, enter a project's view, in its tasks list, for the same tasks (ex: no. 125 and 126) you see "Unknown user" as the task creator, and nothing at all as the "Assigned user" (even though those tasks absolutely do have an assigned user):

Image

Same problem if you visit the "Client" details page and click its "Tasks" tab:

Image

turbo124:
closes #2652

### DIFF
--- a/src/pages/clients/show/pages/Tasks.tsx
+++ b/src/pages/clients/show/pages/Tasks.tsx
@@ -44,7 +44,7 @@ export default function Tasks() {
     <DataTable
       resource="task"
       endpoint={route(
-        '/api/v1/tasks?include=status,client&client_id=:id&sort=id|desc',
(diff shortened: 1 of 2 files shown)


beganovich (approved):
Thanks 👍

PR #2657, opened 2025-08-13
[Bugfix] User Selector Read Only | Projects

Civolilah:
@beganovich @turbo124 The PR includes fixes for removing the `readOnly` prop from the user selector on the project edit page. Screenshot:

Screenshot 2025-08-14 at 01 54 17

Let me know your thoughts.


beganovich (approved):
Thanks 👍

Issue #2660, opened 2025-08-12
Custom Field switch status on client details doesnt update


tomtom30:
## Setup
- Version: v5.12.8

- Environment: Docker

## Checklist
- Can you replicate the issue on our v5 demo site https://demo.invoiceninja.com or https://react.invoicing.co/demo?
    YES
- Have you searched existing issues?
    YES
- Have you reported this to Slack/forum before posting?
    NO
- Have you inspected the logs in storage/logs/laravel.log for any errors?
    NO

## Describe the bug
The custom switch state is not updated on the client details description.

### Steps To Reproduce
Create a custom field switch and update the value (API or manual) and back to client description to look the status ( yes / no )

### Expected Behavior
The value is the same everywhere.

Thanks
Thomas

turbo124:
Looks like this issue is with the interface not displaying the correct value when toggled off.

PR #2662, opened 2025-08-15
Minor fixes for header checks of CSV file

turbo124:
Allows less column headers than total columns with data.


turbo124:
closes #2660

PR #2672, opened 2025-08-22
[Feature] Filtering In Email History Section

Civolilah:
@beganovich @turbo124 The PR includes adding filtering to the "Email History" card/section on the Client overview page. Currently, I implemented React filtering without calling the API, as the API always returns all emails. To filter emails, I check if the filter is contained in the `subject`, `recipients`, or `entity`, or if a date is entered, I check the dates in the events array to see if it includes any of those dates.

`Note`: I used "search_emails" translation keyword, so if you agree please just add it into files.

Let me know your thoughts.


Civolilah:
@turbo124 Should this be added only on the Client show page, or everywhere we display the Email History list?

PR #2677, opened 2025-08-27
Update email configuration to provide SES hosted as an option

turbo124:
Adds SES as a sending option for hosted users.


beganovich (approved):
Thanks 👍

PR #2687, opened 2025-09-02
Fixes for SES including settings the from address in the UI

turbo124:
Adds additional configuration props for AWS SES mailer


beganovich (approved):
Thanks 👍

PR #2689, opened 2025-09-02
[Feature] Bulk Update Action | Tasks

Civolilah:
@beganovich @turbo124 The PR includes the implementation of 'bulk_update' action functionality for tasks. Let me know your thoughts.

### DIFF
--- a/src/common/hooks/useBulkUpdatesColumns.ts
+++ b/src/common/hooks/useBulkUpdatesColumns.ts
@@ -16,9 +16,14 @@ export function useBulkUpdatesColumns() {
 
   const [bulkUpdates, setBulkUpdates] = useState<Record<string, string[]>>();
 
+  console.log(statics?.bulk_updates);
+
   useEffect(() => {
     if (statics?.bulk_updates) {
-      setBulkUpdates(statics.bulk_updates);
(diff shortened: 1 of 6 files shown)


Civolilah:
> ```
>       >
>         <div className="flex flex-col space-y-5">
>           <SelectField
>             label={t('column')}
>             value={column}
>             onValueChange={(value) => {
>               setColumn(value);
>             }}
>             withBlank
>             customSelector
>           >
>             {bulkUpdatesColumns?.[props.entity] <------ this line here bugs out
>               .filter((currentCol) => showColumn(currentCol))
>               .map((currentColumn) => (
>                 <option key={currentColumn} value={currentColumn}>
>                   {getColumnTranslation(currentColumn)}
>                 </option>
>               ))}
>           </SelectField>
> ```
> 
> @Civolilah on selecting a checkbox, the white screen of death appears.

@turbo124 Hmm, I think the reason for that is not having the task property in the bulk_updates response. I had the same issue and I added it manually, then removed it after implementation. So, just make sure that the bulk_updates prop in the static query contains the task entity.

beganovich (approved):
Thanks 👍

PR #2697, opened 2025-09-05
[Cleanup] Adjusting Max Width For Products Selector

Civolilah:
@beganovich @turbo124 The PR includes adjusting the max width for the product selector. So, if there are long notes, for example, the product selector will be significantly wider than we had it earlier, but if there is no need for that (if there are no long notes), the size will stay as default. Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #2701, opened 2025-09-08
[Bugfix] Gateways Confirmation Modal

Civolilah:
@beganovich @turbo124 The PR includes a fix for the confirmation dialog when deleting gateways on client show page. Let me know your thoughts.

### DIFF
--- a/src/pages/clients/show/components/Gateways.tsx
+++ b/src/pages/clients/show/components/Gateways.tsx
@@ -16,7 +16,7 @@ import { GatewayLogoName, GatewayTypeIcon } from './GatewayTypeIcon';
 import { useCompanyGatewaysQuery } from '$app/common/queries/company-gateways';
 import { useEffect, useState } from 'react';
 import { CompanyGateway } from '$app/common/interfaces/company-gateway';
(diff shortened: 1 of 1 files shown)


src/pages/clients/show/components/Gateways.tsx
```diff
@@ -115,17 +110,30 @@ export function Gateways(props: Props) {
...
   }, [deleteGatewayTokenId]);
 
   return (
     <>
-      <ConfirmActionModal
-        onClick={() => handleDeleteGatewayToken()}
-        onClose={() => setDeleteGatewayTokenId('')}
-        disabledButton={isFormBusy}
-      />
+      <Modal
```
Civolilah, line 119: @beganovich Quick confirmation before your review. The reason I removed `ConfirmActionModal` is that when the component uses an atom, all other props are also overwritten if the same component is rendered elsewhere on the page. In this case, we use it for bulk deleting invoices, so it overwrites the disabled and other props. I discovered this behavior for the first time after encountering this bug :)

beganovich (approved):
Thanks 👍

PR #2713, opened 2025-09-16
[Bugfix] Net Amount Calculation | Expenses | Recurring Expenses

Civolilah:
@beganovich @turbo124 This PR includes fixes for net amount calculations on expenses and recurring expenses. Let me know your thoughts.


turbo124:
image

not sure why we have this label appear when tax by amounts is present, it shows a %... I don't think we should have this label at all.

beganovich (approved):
Thanks 👍

PR #2727, opened 2025-09-29
[Bugfix] 404 API Error

Civolilah:
@beganovich @turbo124 This PR includes fixes for behavior when the API returns a 404. Instead of navigating to the 404 page, we now display a not_found toaster. Let me know your thoughts.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -163,8 +163,8 @@ export function App() {
   }, [darkMode, resolvedLanguage]);
 
   useEffect(() => {
-    window.addEventListener('navigate.invalid.page', () =>
-      navigate('/not_found')
+    window.addEventListener('display.not.found.toaster', () =>
+      toast.error('not_found')
     );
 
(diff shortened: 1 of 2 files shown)


Civolilah:
> @Civolilah the logic here is still not correct:
> 
> We still need to maintain the /not_found route if the user attempts to access a route which does not exist in routes file of the REACT application ie:
> 
> image
> The toast should only be shown when the API response returns a 404

@turbo124 I think that you experienced a blank screen just because of the local environment. Even in the current production it works as expected, so the API is not even triggered - this route just does not exist in our route system. This is from current production:

Screenshot 2025-09-30 at 20 49 58

This is locally on this branch:

Screenshot 2025-09-30 at 20 50 14

But for 404 errors from API, the toaster will be displayed.

Let me know your thoughts.

beganovich (approved):
Thanks 👍

PR #2738, opened 2025-10-07
[Feature] Creating Feedback Component

Civolilah:
work in progress....


turbo124:
@Civolilah thanks, looks good.

I think we should only show this on the dashboard page.

turbo124:
@Civolilah looks good. 

Just not sure on background color..

image


@beganovich thoughts on a good combination for background vs selection colors?

beganovich (approved):
Thanks 👍

PR #2741, opened 2025-10-08
Dynamic navigation

beganovich:
This enables dynamic status of navigation item and their visibility selection.

We can now hide or show navigation items from anywhere using helper:

```tsx
$visibility('/documents/users', true);
$visibility('/documents/blueprints', true);
```

The first parameter is href of the sidebar item. Helper works with both sub and regular items.

Please see demo:

https://github.com/user-attachments/assets/a7be0e6c-660a-4eb8-9d85-e4c68ff28d4f


beganovich:
@turbo124 

For the demo purposes, I've set documents and subitems visibility to false as default as well as simulated async visiblity toggle: https://github.com/invoiceninja/ui/blob/DOC-362/src/components/layouts/common/navigation.ts#L377

Please remove them and set the default values for documents as you wish.

PR #2748, opened 2025-10-13
[Feature] New Task Filters

### LINKED ISSUE #2696: Ability to filter the Tasks listview table by assigned user, client or project, in addition to the search's filtering
With the v5 Enterprise SaaS version, one little shortcoming I have with the interface at https://app.invoicing.co/#/tasks is that I can't easily drill down to a particular combination of client+project+assignee + optionally search terms.

Let's say you have an assigned user "Alice", a task description "Increase banana sales tenfold", client "Chiquita", project "2025-09 back to school event".
* If you search for "banana" you'll find it
* If you search "chiquita" you'll find it
* If you search "Alice" you probably won't find it
* If you search "banana alice chiquita", or any combination of the words in the fields you certainly won't find it.

So basically, short of having the ability to do "multiple partial strings keywords in any order in any fields" search, I think what would be useful would be to have more more filtering dropdown comboboxes available above the table in addition to the "Filter" searchentry, instead of just "Lifecycle" and "Status":

Image

Then I could filter one or two of those comboboxes (for client, project, assigned user?) and _then_ apply a text search filter to further drill down.

This would be useful to be able to then select multiple results and apply some batch operations to them as part of https://github.com/invoiceninja/invoiceninja/issues/10383

Civolilah:
@beganovich @turbo124 The PR includes implementation of filter columns for the Tasks table. We now have three additional filters on this table: "Assigned User", "Project", and "Client". Once values are selected through the filter popup, we will send user_ids, project_ids, or client_ids as a comma-separated list of IDs. Screenshot:

Screenshot 2025-10-14 at 20 53 27

Closes #2696 

Let me know your thoughts.

### DIFF
--- a/src/common/events.ts
+++ b/src/common/events.ts
@@ -1,3 +1,4 @@
 export type Events = {
   'bulk.completed': void;
+  'date_range_picker.clear': void;
 };
--- a/src/components/DataTable.tsx
+++ b/src/components/DataTable.tsx
@@ -182,6 +182,7 @@ export type ResourceAction<T> = (resource: T) => ReactElement;
 
 export type PerPage = '10' | '50' | '100';
 
+export const dateRangeAtom = atom<string>('');
 export const dataTableSelectedAtom = atom<Record<string, string[]>>({});
 export const filterColumnsValuesAtom = atom<Record<string, string[]>>({});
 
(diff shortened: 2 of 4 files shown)


turbo124:
@Civolilah 

For the date filter, if i remove the filter values (clicking x) the query does not refresh.



Please change cancel text to clear

image

Finally, when we have filters in place, we should present a button adjacent to Columns with the Number of filters active with a Clear Filters title to wipe all filters in place.

Civolilah:
@turbo124 I think I've covered everything you mentioned. Please let me know if you see anything else that needs to be covered. Screenshot:

Screenshot 2025-10-16 at 20 32 13

Let me know your thoughts.

Civolilah:
> @Civolilah when clearing the filters, the date filter persists

@turbo124 Ah sorry, the date range logic is a bit separated, so I forgot to include it there. It should be included now as well. Let me know your thoughts.

Issue #2733, opened 2025-10-04
Tax not rounding up correctly


alexgauvin:
I have to open up [this issue](https://github.com/invoiceninja/ui/issues/2472) again. I still have some totals not rounding up as expected.

Here's a few screenshots

**Invoice footer on edit page**
<img width="647" height="588" alt="Image" src="https://github.com/user-attachments/assets/d8d467a8-b1f0-4e26-84c4-d7a506dd328c" />

**PDF view from same invoice**
<img width="391" height="294" alt="Image" src="https://github.com/user-attachments/assets/86d48fe4-8601-4d54-829a-9508cac356d9" />

**Payment screen for same invoice**
<img width="1457" height="650" alt="Image" src="https://github.com/user-attachments/assets/bb53a44f-d609-49bd-af1f-a12ef2d8035a" />

I still believe the problem is related to the way IN rounds the taxes.
I'd like to remind you that the taxes in Canada/Quebec are 5% and 9.975%.

Summing up all invoices in Excel for same period gives me $ 21,961.26 and in Invoice Ninja, I get $ 21,961.31
$ 0.05. Not much, but still a difference.

Thanks

turbo124:
Can you share a screenshot of your excel calculation logic.


5c is way off

alexgauvin:
As mentionned in https://github.com/invoiceninja/ui/issues/2472, the miscalculation seems to show up only on invoice edit page total footer, but correct on PDF view.

Here is a screenshot portion of Sheets file. Logic is simple = sums of columns. I keep this for validation purposes.

<img width="544" height="450" alt="Image" src="https://github.com/user-attachments/assets/e51172fe-654f-40fb-8ff0-910ec8b94888" />


And here's the same invoice viewed in Flutter and ReactUI

<img width="811" height="477" alt="Image" src="https://github.com/user-attachments/assets/2ff57763-0b0d-4939-a7ef-b8870f3336af" />

turbo124:
For these invoices that are out of balance, does the amount / balance appear correct when viewing the invoice list?

How are you applying payments to these invoices? are you manually marking them as paid, or entering a payment amount?

alexgauvin:
> For these invoices that are out of balance, does the amount / balance appear correct when viewing the invoice list?

Yes, the only places where the amount are wrong it's on the Edit invoice page, under the subtotal section and in the dashboard Recent transactions payments total That's where I can observe the $0.05 difference)

[...]

PR #2751, opened 2025-10-15
Use epsilon in calculations for invoice tax totals

turbo124:
Introduces the addition of an epsilon value to account for float point math when calculating dynamic values from the invoice sum class,


turbo124:
related to #2733

beganovich (approved):
Thanks :+1:

beganovich:
@turbo124 do we push this straight into production?

PR #2756, opened 2025-10-19
[Cleanup] UI/UX Refactoring Payment Pages

Civolilah:
@beganovich @turbo124 This PR includes refactoring of the UX for applying invoices on the payment page. Screenshot:

Screenshot 2025-10-21 at 13 09 54

Let me know your thoughts.

### DIFF
--- a/src/components/DataTable.tsx
+++ b/src/components/DataTable.tsx
@@ -688,9 +688,15 @@ export function DataTable<T extends object>(props: Props<T>) {
   }, []);
 
   useEffect(() => {
-    emitter.on('deselect.resource', (id: string) =>
-      setSelected((current) => current.filter((v) => v !== id))
-    );
+    const handleDeselectResource = (id: string) => {
+      setSelected((current) => current.filter((v) => v !== id));
+    };
+
+    emitter.on('deselect.resource', handleDeselectResource);
+
+    return () => {
+      emitter.off('deselect.resource', handleDeselectResource);
+    };
   }, []);
 
   useEffect(() => {
--- a/src/pages/payments/apply/Apply.tsx
+++ b/src/pages/payments/apply/Apply.tsx
@@ -20,7 +20,6 @@ import { FormikProps, useFormik } from 'formik';
 import { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useParams } from 'react-router-dom';
-import { v4 } from 'uuid';
 import { useNavigate } from 'react-router-dom';
 import { useFormatMoney } from '$app/common/hooks/money/useFormatMoney';
 import { useSaveBtn } from '$app/components/layouts/common/hooks';
@@ -187,7 +186,7 @@ export default function Apply() {
 
                   selectedResources.forEach((resource: Invoice) => {
                     newInvoices.push({
-                      _id: v4(),
+                      _id: window.crypto.randomUUID(),
                       amount: calcApplyAmount(resource.balance, newInvoices),
                       credit_id: '',
                       invoice_id: resource.id,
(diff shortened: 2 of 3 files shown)


Civolilah:
> @Civolilah looks good.
> 
> What happens if there are thousands of invoices, do we only pull back a limited set? if we sort by date, does it requery the results based on sort|date|asc|desc?

@turbo124 So, right now we pull 1000 invoices here, and yes, just as for regular tables, sorting by any column works, so I just pulled 1000 invoices and removed pagination. I assumed it was enough. What do you think?

Civolilah:
> @Civolilah
> 
> Thanks for the update. Can we also use this format format for the create payment page too?

@turbo124 Sure thing, I've refactored the create page as well. Screenshot:

Screenshot 2025-11-07 at 19 46 17

Let me know your thoughts.

turbo124:
@Civolilah 

image


When using the apply credit option, the credit is not selected (nor visible for selection)

Civolilah:
> Tracking revision in RU-2052

@turbo124 I gave comment on: https://invoiceninja.atlassian.net/browse/RU-2052 and the implementations are also pushed to this PR. Let me know your thoughts.

turbo124:
@Civolilah 

image


if i istart filtering, and the filter returns 0 results... i think get this, which is irreversible


image

Civolilah:
> @Civolilah
> 
> image
> if i istart filtering, and the filter returns 0 results... i think get this, which is irreversible
> 
> image

@turbo124 Ah sure, that's my bad, but if we want to have this empty state that is not a classic empty state in the table, we must be aware that the filter field must be displayed. Because we can't determine the reason there are 0 results, is it the search or are there just no results. Screenshot:

Screenshot 2025-11-23 at 19 38 57

Let me know your thoughts.

Civolilah:
@turbo124 As we discussed on the Slack error messages under table fields have been adjusted here. Please let me know if you see any issues or if you have any additional suggestions.

Civolilah:
> @Civolilah i am seeing issues with the currency formatting of the amounts here. They are not consistent for what should be displayed for the client?

@turbo124 Yes, you noticed that well, we didn't have the correct formatting. I fixed it just now, and it looks great on my end. Screenshot:

Screenshot 2025-12-10 at 19 46 01

Let me know your thoughts.

src/pages/payments/apply/Apply.tsx
```diff
@@ -177,105 +172,54 @@ export default function Apply() {
...
+              resource="invoice"
+              endpoint={`/api/v1/invoices?include=client&payable=${payment?.client_id}&per_page=100&sort=date|desc&per_page=1000`}
+              columns={columns}
+              onSelectedResourcesChange={(selectedResources) => {
+                if (selectedResources.length > 0) {
+                  const newInvoices: any[] = [];
+
+                  selectedResources.forEach((resource: Invoice) => {
+                    newInvoices.push({
+                      _id: v4(),
```
beganovich: I think it's safe for us to move to `window.crypto` in future for this since browser APIs can generate UUIDs now.

beganovich (changes requested):

src/components/DataTable.tsx
```diff
@@ -660,6 +687,22 @@ export function DataTable<T extends object>(props: Props<T>) {
     emitter.on('bulk.completed', () => setSelected([]));
   }, []);
 
+  useEffect(() => {
+    emitter.on('deselect.resource', (id: string) =>
```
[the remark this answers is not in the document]
Civolilah: @beganovich Done!

src/pages/payments/apply/Apply.tsx
Civolilah: @beganovich Replaced!

PR #2767, opened 2025-10-25
[Bugfix] Fixing Issue For Add To Invoice | Expense

Civolilah:
@beganovich @turbo124 This PR fixes the issue with converting amounts from expense currency to invoice currency when they differ in the "Add to Invoice" expense action. Let me know your thoughts.

### DIFF
--- a/src/pages/expenses/common/components/AddToInvoiceAction.tsx
+++ b/src/pages/expenses/common/components/AddToInvoiceAction.tsx
@@ -81,21 +81,9 @@ export function AddToInvoiceAction(props: Props) {
     setVisibleModal(true);
   };
 
-  const getConvertedAmount = (expense: Expense, invoice: Invoice) => {
+  const getCurrentAmount = (expense: Expense, invoice: Invoice) => {
     if (expense.currency_id !== invoice.client?.settings.currency_id) {
-      const expenseCurrency = resolveCurrency(expense.currency_id);
-      const invoiceCurrency = resolveCurrency(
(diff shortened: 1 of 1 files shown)


turbo124:
ping @Civolilah for update here please.

Civolilah:
> ping @Civolilah for update here please.

@turbo124 So, I think we just misunderstood each other on this one. What I did in the logic is basically convert the expense currency to the invoice's client currency. So, basically, if the expense is in EUR and that expense is added to an invoice that has a client with USD, it will be converted into USD currency per the exchange rates.

So, I'm a little bit confused about how this can even help us with converting/foreign amount in this case. Right now, a user can add an expense to any invoice, which means that the currency used under additional_info as the converting currency can be totally different from the currency on the invoice, and it still requires conversion at the moment of adding the expense on invoice.

Sorry If I got something wrong here.

turbo124:
@Civolilah If expense.currency_id == client.currency_id => use expense.amount

otherwise use expense.foreign_amount

You do not need to perform any calculations here. just a simple logic check on client.currency_id => expense.currency_id

Civolilah:
> @Civolilah If expense.currency_id == client.currency_id => use expense.amount
> 
> otherwise use expense.foreign_amount
> 
> You do not need to perform any calculations here. just a simple logic check on client.currency_id => expense.currency_id

@turbo124 Sure thing, the solution has been updated just as you explained. Please let me know if it works for you as expected.

turbo124:
:muscle:

PR #2770, opened 2025-10-28
Map Signatories

turbo124:
When converting a template (with placeholder props) we need to map the signatories to clients/users so that the template variable resolve correctly.

This PR implements a wizard solution where signatories can be assigned prior to the Templates PDF generates in order for variables to render in the document.


beganovich (approved):
Looks good to me, ping for conflicts @turbo124.

PR #2772, opened 2025-10-28
[Bugfix] Send Button | Builder Page 

Civolilah:
@beganovich @turbo124 The PR includes adding a condition for the "Send" button on the "Builder" page. The button will be displayed only if there are signatories added. Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #2776, opened 2025-11-02
[Cleanup] Confirmation Modals | Document Actions

Civolilah:
@beganovich @turbo124 This PR includes the implementation of confirmation modals for "Clone", "Void", and "Clone to template" document actions. Please let me know if there are any other actions that need confirmation modals.

### DIFF
--- a/src/pages/documents/common/components/CloneAction.tsx
+++ b/src/pages/documents/common/components/CloneAction.tsx
@@ -28,6 +28,10 @@ export function CloneAction({ document }: Props) {
         onClose={() => setIsModalVisible(false)}
       >
         <div className="flex flex-col space-y-6">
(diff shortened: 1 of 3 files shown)


turbo124:
@Civolilah 

looks good, but feels like we are missing supplementary text for the action itself.

image

Civolilah:
> @Civolilah
> 
> looks good, but feels like we are missing supplementary text for the action itself.
> 
> image

@turbo124 Sure thing, I've added a confirmation message for each modal. Screenshot:

Screenshot 2025-11-04 at 19 11 51

So, I've added "clone_document_description", "make_template_description", and "void_document_description" translation keywords, so if you agree, please just add those translation keywords to the files.

Let me know your thoughts.

PR #2778, opened 2025-11-03
Signatory swaps

beganovich:
This implements swapping signatories in Invoice Ninja.


beganovich:
Turns out there was a bug upstream, see https://github.com/DocuNinja/builder2.0/pull/66.

PR #2796, opened 2025-11-14
Handle selecting client after creation

beganovich:
This auto preselects client after creation.


turbo124:
@beganovich this isn't working for me? no console errors either?

beganovich:
There was a flaw in logic, please see now :+1:

PR #2801, opened 2025-11-18
[Cleanup] Displaying Custom Surcharge Fields

Civolilah:
@beganovich @turbo124 The PR includes cleanup logic for custom surcharge fields on invoice pages. So, instead of expecting to display them once labels are entered and they are included in totals, we will only expect labels to be entered to align the logic with the rest of the custom fields. Let me know your thoughts.


beganovich (approved):
Thanks :+1:

PR #2805, opened 2025-11-19
[Cleanup] Mark Paid Action Change Route

Civolilah:
@beganovich @turbo124 The PR includes changes for the "Mark Paid" Invoice action route. Instead of using the `bulk` route, we use the PUT route with the `mark_paid=true` query parameter. Let me know your thoughts.


Civolilah:
> @Civolilah please ensure that this is changes from the list view also.
> 
> image

@turbo124 Sure the same logic has been applied on table also.

beganovich (approved):
Thanks 👍

PR #2817, opened 2025-11-25
Fix/eliminate redundant client vendor api calls

turbo124:
Continues to work on issues where the client resolver is called directly which causes issue with lesser permissioned users.

adds the ability to pass the relation if present, to prevent the client resolver from needing to make an API for the client record itself.


turbo124:
@beganovich for merge to main please. may need to be carefully assessed.

PR #2835, opened 2025-12-03
2fa enrollment QR code fix

TechNoNerd87:
Fixes issues where the QR code will not scan in dark mode.


turbo124:
@TechNoNerd87 can you rebase so that only your changes come through please.

PR #2840, opened 2025-12-06
[Bugfix] Table Column Preferences Issue

Civolilah:
@beganovich @turbo124 The PR includes fixes for saving table column preferences. In the video that Dave posted in the ticket, I noticed that this property had somehow become an array for this user, which is very weird because we never handle it that way. However, I was able to recreate the issue by manually setting this property to an empty array (as it appeared for the client who experienced the issue) and implemented a fix accordingly. Everything works perfectly fine on my end now. Let me know your thoughts.


turbo124:
@beganovich please merge to main post review. thks

PR #2824, opened 2025-11-28
[Bugfix] Duplicating Invoice Bulk Actions

Civolilah:
@beganovich @turbo124 This PR includes fixes for duplicating archive/delete/restore bulk actions. Let me know your thoughts.

PR #2842, opened 2025-12-07
[Bugfix] Custom Bulk Actions Duplication

Civolilah:
@beganovich @turbo124 This PR includes fixes for duplicating archive/delete/restore bulk actions. 

Formatting removal for this PR: #2824 

Let me know your thoughts.

### DIFF
--- a/src/pages/invoices/Invoice.tsx
+++ b/src/pages/invoices/Invoice.tsx
@@ -172,8 +172,7 @@ export default function Invoice() {
                 disableSaveButton={
                   (invoice &&
                     (invoice.status_id === InvoiceStatus.Cancelled)) ||
-                  isFormBusy || 
-                  invoice.is_locked
+                  isFormBusy
                 }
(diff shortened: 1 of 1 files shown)


turbo124:
@Civolilah with this PR, i no longer see the delete option.

Civolilah:
> A deleted invoice should have the option to be restored?
> 
> image image image
> The restore option only appears from the list view when selecting the row option menu.
> 
> Restore not available from bulk actions, not directly from edit screen.
> 
> No special e-invoicing levels are active here.

@turbo124 Okay, so the reason you didn't see the "Restore" action in bulk actions was due to a wrong condition. But on the edit page, the condition for disabling the actions dropdown was also wrong, which is why you couldn't open it. So, as we discussed on Slack, I matched the logic from the single action table page and it works great for me now. Screenshot:

Screenshot 2025-12-09 at 10 21 10

Please let me know if it also looks good on your end.

beganovich (approved):
Thanks 👍

PR #2847, opened 2025-12-11
[Bugfix] Money Formatting Per Client | Create Payment Page

Civolilah:
@beganovich @turbo124 The PR includes fixes for formatting money values on the create payment page. Let me know your thoughts.

### DIFF
--- a/src/common/hooks/useReactSettings.ts
+++ b/src/common/hooks/useReactSettings.ts
@@ -35,6 +35,7 @@ export interface Preferences {
     default_view: 'day' | 'week' | 'month';
     range: string;
     currency: number;
+    include_drafts: boolean;
   };
   datatables: {
     clients: {
@@ -97,6 +98,7 @@ export const preferencesDefaults: Preferences = {
     default_view: 'month',
     currency: 1,
     range: 'this_month',
+    include_drafts: false,
   },
   datatables: {
     clients: {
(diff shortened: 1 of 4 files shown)


src/pages/payments/create/Create.tsx
```diff
@@ -204,6 +208,14 @@ export default function Create() {
...
   const onSubmit = useSave({ setErrors, setIsFormBusy, isFormBusy });
 
+  useEffect(() => {
+    if (payment?.client_id) {
+      clientResolver
+        .find(payment.client_id)
+        .then((client) => setResolvedClient(client));
+    }
+  }, [payment?.client_id]);
+
```
turbo124: I think we need to align this type of call with a recent change that we implemented around resolving clients.


```ts
  useEffect(() => {
    // Use the client data that's already included in the invoice relation
    // instead of fetching it separately (which requires view_client permission)
    if (resource?.client) {
      setClient(resource.client);
    } else if (resource?.client_id) {
      // Only fetch if client not already included
      clientResolver
        .find(resource.client_id)
        .then((client) => setClient(client))
        .catch(() => {
          // Silently fail if user doesn't have view_client permission
        });
    }
  }, [resource?.client_id]);
```

For this type of query, we'll always have the client record when it is selected. It is important for several reasons, particularly around permissions and preventing 401's

Civolilah: @turbo124 Yes, you are right, we can mostly take the client object from the selector. I think I've significantly reduced the number of times we call `clientResolver`. Please let me know how it looks from your end.

beganovich (approved):
Thanks 👍

PR #2861, opened 2025-12-19
[Bugfix] Tiptap Empty Text Value

Civolilah:
@beganovich @turbo124 The PR includes fixes to avoid having an empty <p> tag when no text is entered, ensuring an empty string is used as the value instead. Let me know your thoughts.


turbo124:
@beganovich please merge to main post review. thanks!

PR #2865, opened 2025-12-21
[Feature] Quote Rejected Notification

Civolilah:
@beganovich @turbo124 The PR includes adding "Quote Rejected" notifications in configuration. Screenshot:

Screenshot 2025-12-21 at 15 37 15

`Note`: I used the "quote_rejected" translation key, so if you agree, please add it to the translation files as we're missing it.

Let me know your thoughts.


turbo124:
Settings > User Managament | User - Notifications tab has not been updated.

PR #2869, opened 2025-12-22
[Cleanup] Hiding PEPPOL Cards For Free Users

Civolilah:
@beganovich @turbo124 The PR includes hiding unnecessary PEPPOL cards and fields when the user becomes free. Let me know your thoughts.

### DIFF
--- a/src/pages/settings/e-invoice/EInvoice.tsx
+++ b/src/pages/settings/e-invoice/EInvoice.tsx
@@ -83,9 +83,9 @@ export function EInvoice() {
   const [t] = useTranslation();
 
   const shouldShowPEPPOLOption = (skipPlanCheck: boolean = false) => {
-    if (import.meta.env.DEV) {
-      return true;
-    }
+    // if (import.meta.env.DEV) {
+    //   return true;
+    // }
 
     const isPlanActive =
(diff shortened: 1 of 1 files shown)


src/pages/settings/e-invoice/EInvoice.tsx
```diff
@@ -500,15 +518,18 @@ export function EInvoice() {
...
-        currentEInvoice={company?.e_invoice || {}}
-        entity="company"
-      />
+      {!freePlan() && (
+        <PaymentMeans
+          ref={eInvoiceRef}
+          currentEInvoice={company?.e_invoice || {}}
+          entity="company"
+        />
+      )}
```
turbo124: this should show for free users.

Civolilah:
> @Civolilah this feature mainly should be focussed on allowing a "Free" user to disconnect their entity from the PEPPOL network.
> 
> Currently, if they are free, we hide all PEPPOL related cards. This PR should make the disconnect CTA available to a user where company.legal_entity_id is not null / or != 0

@turbo124 I think I found what the problem was in the logic that prevented the "Preference" component from displaying on the UI along with the needed button to disconnect from the PEPPOL network. Can you please check right now if you see the correct UI displayed?

PR #2879, opened 2026-01-02
[Bugfix] Applying Description When Invoicing Task

Civolilah:
@beganovich @turbo124 The PR includes fixes for the description not being applied when invoicing a task. Let me know your thoughts.


turbo124:
@Civolilah 


When using the Add To Invoice action, the description does not carry over.

Civolilah:
> @Civolilah
> 
> When using the Add To Invoice action, the description does not carry over.

@turbo124 Ah sure thing, I forgot that we do the same thing in that action as well. I've done it, let me know if it works great on your end too.

PR #2886, opened 2026-01-06
[Bugfix] Custom Fields Trailing Spaces

Civolilah:
@beganovich @turbo124 The PR includes fixes for data table custom columns when empty space has been entered at the end of a custom field label, ensuring it will be displayed in the table if it is selecetd. Let me know your thoughts.


src/common/helpers/data-table.ts
```diff
@@ -0,0 +1,4 @@
+export const normalizeColumnName = (columnName: string): string => {
+  if (!columnName) return '';
+  return columnName.trimEnd().replace(/\/+$/, '');
```
turbo124: @Civolilah we don't need to .replace() just the trimEnd() ?

PR #2892, opened 2026-01-08
[Bugfix] EInvoice Tabs

Civolilah:
@beganovich @turbo124 The PR includes logic fixes for e_invoice tabs for invoices and recurring_invoices. The tabs will now be visible if EInvoice has been ENABLED (`enable_e_invoice === true`). Under the Invoices tab, the `EInvoice card and Status` card will ONLY be visible if `e_invoice_type is PEPPOL AND acts_as_sender === true`, OR IF `verifactuEnabled === true` (I hope I assumed correctly that we should have a fallback to this condition).

For Recurring Invoices, the tab logic is the same, but we didn't have the first two cards under the tab at all, so I've added an EInvoice card which will be displayed under the same conditions as for Invoices, enabling users to validate invoices in the same way as for Invoices. Screenshot:

Screenshot 2026-01-08 at 08 10 01

Let me know your thoughts.

### DIFF
--- a/src/pages/recurring-invoices/edit/hooks/useTabs.tsx
+++ b/src/pages/recurring-invoices/edit/hooks/useTabs.tsx
@@ -46,7 +46,7 @@ export function useTabs(params: Params) {
     {
       name: t('e_invoice'),
       href: route('/recurring_invoices/:id/e_invoice', { id }),
-      enabled: company?.settings.enable_e_invoice === true,
+      enabled: company?.settings.enable_e_invoice === true || company?.settings.e_invoice_type === 'PEPPOL',
     },
     {
       name: t('documents'),


Civolilah:
@turbo124 If you can, please confirm that we should have a verifactuEnabled fallback for the condition.

```
(company?.settings.e_invoice_type === 'PEPPOL' &&
      company?.tax_data?.acts_as_sender) || varifactuEnabled
```
      
Let me know your thoughts.

Civolilah:
> @Civolilah I believe we show a dedicated Verifactu Component rather than the EInvoice component?

@turbo124 Ah sure thing, we have another tab for Verifactu. I've removed the fallback condition, please let me know if the fix works for you.

PR #2905, opened 2026-01-13
[Cleanup] Link Labels Invoices/Expenses

Civolilah:
@beganovich @turbo124 The PR includes cleanup of labels for invoices and expenses on transactions. Since we can have only one linked expense (one expense_id), we display only "Link" as the label, but for invoices we will display a template of "Invoice #1, Invoice #2, etc." Screenshot:

Screenshot 2026-01-13 at 19 07 36

Let me know your thoughts.


Civolilah:
> image
> I think there is an issue with the additional $ in the string for the invoice #?

@turbo124 Ah sure thing, my bad. It has been removed.

turbo124:
@beganovich it "should", but due to datastorage types, it is not practical. But again, the expense # means very little in this particular context we implement similar functionality when there are linked records such as quote=>invoice invoice=>quote

beganovich:
Got it, thanks guys!

PR #2932, opened 2026-01-28
[Feature] Adding created_between Date Range Filter

Civolilah:
@beganovich @turbo124 The PR includes the implementation of a new table filter across all tables with the query parameter "created_between". Screenshot:

Screenshot 2026-01-28 at 16 45 51

Let me know your thoughts.

### DIFF
--- a/src/components/DataTable.tsx
+++ b/src/components/DataTable.tsx
@@ -634,11 +634,13 @@ export function DataTable<T extends object>(props: Props<T>) {
     },
     [setDateRangeEntries]
   );
-  const getDateRangeEntryForColumn = (
-    columnId: string
-  ): DateRangeEntry | undefined => {
-    return dateRangeEntries.find((entry) => entry.column === columnId);
-  };
+
+  const getDateRangeEntryForColumn = useCallback(
+    (columnId: string): DateRangeEntry | undefined => {
+      return dateRangeEntries.find((entry) => entry.column === columnId);
+    },
+    [dateRangeEntries]
+  );
 
   const getFooterColumn = (columnId: string) => {
     return footerColumns.find((footerColumn) => footerColumn.id === columnId);


turbo124:
image

for projects, i am seeing the filter on due-date and nothing on Date Created. the query parameter has ?created_between=

turbo124:
for consistency, products date range filter on due date should be 

date_range=due_date,2026-01-21,2026-01-31

Civolilah:
> for consistency, products date range filter on due date should be
> 
> date_range=due_date,2026-01-21,2026-01-31

@turbo124 Sure, so for "Projects" and the `date_range` query parameter, we have also included a column in the pattern: `date_range=due_date,startDate,endDate`. Screenshot:

Screenshot 2026-02-02 at 08 17 11

Let me know your thoughts.

turbo124:
The filter fires when the first date is selected (with no end date) 

the filter should only filter when the start/end dates have both been selected.

Civolilah:
> This only works the first time a date range is set.
> 
> if i then clear the date range, and set a new one, the api call does not trigger......



> This only works the first time a date range is set.
> 
> if i then clear the date range, and set a new one, the api call does not trigger......

@turbo124 Sorry, I had to refactor the logic and that caused the issue. I recreated that case and tested many more scenarios, and it works perfectly fine on my end. Demo:

https://github.com/user-attachments/assets/9c7425b3-45f4-48ec-a478-4489d9616da2

Let me know your thoughts.

src/components/DataTable.tsx
```diff
@@ -550,21 +580,64 @@ export function DataTable<T extends object>(props: Props<T>) {
...
+        ];
+      });
+    },
+    [setDateRangeEntries]
+  );
+  const getDateRangeEntryForColumn = (
+    columnId: string
+  ): DateRangeEntry | undefined => {
+    return dateRangeEntries.find((entry) => entry.column === columnId);
   };
```
beganovich: I wonder if this is expensive call since it will re-render every time? Worth doing useCallback?

Civolilah: @beganovich Sure thing, it has been improved. Let me know your thoughts.

PR #2950, opened 2026-02-06
[Cleanup] Hiding Gateway Tokens Card And Auto Bill Action 

Civolilah:
@beganovich @turbo124 The PR includes improvements for hiding/filtering gateways and the auto_bill action. The PR covers three cases: hiding the gateways card on the client show page if `company_gateways_ids = '0'`, filtering the current list of gateways in the same card if `company_gateways_ids` is populated with a comma-separated list of gateways ids, and hiding or displaying the `auto_bill` action for invoices under the same conditions as the gateways card on the client show page. Let me know your thoughts.


Civolilah:
@turbo124 The priority is always given to the client level of settings first, and then the group level is checked as well.

PR #2958, opened 2026-02-12
[Cleanup] Clearing Filters And Preferences After "Purge Data"

Civolilah:
@beganovich @turbo124 The PR includes adding logic which will clear up `react_settings`, which contains table filters, common actions, and similar preferences after the "Purge Data" action is triggered. Let me know your thoughts.


turbo124:
@Civolilah i don't think the react settings should change? we don't wipe out any of the company settings, so i don't think the react settings should be any different?

beganovich (approved):
Thanks 👍

PR #2971, opened 2026-02-18
[Bugfix] Cancelling Queries On Company Switch And Logging Out

Civolilah:
@beganovich @turbo124 The PR fixes multiple queries being cancelled when switching companies or logging out. The reason for this was unnecessary query invalidation and cache removal calls in both cases. Since we already perform a full redirect in both cases, which does exactly the same thing for the query client as those functions, we can simply rely on the redirect and remove those calls. Let me know your thoughts.


turbo124:
looks good!

PR #2975, opened 2026-02-19
Fixes for tour settings

beganovich:
This fixes the issue of tour not persisting saves when on builder page.


turbo124:
@beganovich if you have tested this, please merge.

PR #2984, opened 2026-02-20
[Feature] Implementing Purge User Action

Civolilah:
@beganovich @turbo124 The PR implements the purge user action for users. Screenshots:

Screenshot 2026-02-20 at 20 13 08

Screenshot 2026-02-20 at 20 16 48

`Note`: @turbo124 You mentioned that we should prevent the user from deleting the owner, as they should never be deleted. However, I've checked and we don't need any prevention because no one can even see the owner.

Let me know your thoughts.


turbo124:
@Civolilah 

i've tested this, after receiving the successful response with Status Code 204,

the react app attempt to $refresh that user record.

image


this needs to be prevented.

PR #2998, opened 2026-02-25
[Bugfix] Quotes Expiration Badge Logic

### LINKED ISSUE #2997: Quote expires immediately on valid until date instead of at the end of it
## Setup
- Version:  5.12.65
- Environment:  self-host

## Interface
- Flutter: []
- React: []
- Both: []

## Checklist
- Can you replicate the issue on our v5 demo site https://demo.invoiceninja.com or https://react.invoicing.co/demo? yes
- Have you searched existing issues? yes
- Have you reported this to Slack/forum before posting? no
- Have you inspected the logs in storage/logs/laravel.log for any errors? yes

## Describe the bug
A quote that is marked sent will be marked expired on the date of the expiration and sometimes the quote will show up in both the upcoming and expired tables in the dashboard. Because of this the platform makes it difficult to approve the quote (you can't).

### Steps To Reproduce
create a quote and mark valid until for "today". when marked sent it will be immediately changed to expired

### Expected Behavior
a quote should be valid until close of business (whatever that happens to be) on the expiration date.

### Logs

```

```

Civolilah:
@beganovich @turbo124 The PR includes fixes for which badge should be displayed on quotes. So, we have a case where the current day is when the quote is expiring and the status sent from the backend is `-1` (which is the expired status), but we should not display it because the quote is still valid until midnight. So, the fixed logic is like this: if the quote has a `-1` status and is expiring today, the "sent" status badge will be displayed instead, otherwise, the "expired" status badge will be displayed in the case of a `status_id='-1'`. Screenshot:

Screenshot 2026-02-25 at 16 49 34

Closes #2997 

Let me know your thoughts.

### DIFF
--- a/src/pages/quotes/common/components/QuoteStatus.tsx
+++ b/src/pages/quotes/common/components/QuoteStatus.tsx
@@ -13,6 +13,7 @@ import { Badge } from '$app/components/Badge';
 import { useTranslation } from 'react-i18next';
 import { QuoteStatus as QuoteStatusEnum } from '$app/common/enums/quote-status';
 import { useStatusThemeColorScheme } from '$app/pages/settings/user/components/StatusColorTheme';
+import dayjs from 'dayjs';
 
 interface Props {
   entity: Quote;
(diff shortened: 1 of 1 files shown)


beganovich (approved):
Thanks 👍

PR #2999, opened 2026-02-26
[Bugfix] Tasks Tab Hours Rounding

Civolilah:
@beganovich @turbo124 The PR includes a fix for rounding hours in the Tasks tab. Instead of displaying decimal values, hours will be rounded down to the nearest whole number (e.g., 2.6712 → 2h). Screenshot:

Screenshot 2026-02-26 at 01 07 53

Let me know your thoughts.

### DIFF
--- a/src/pages/invoices/common/components/TasksTabLabel.tsx
+++ b/src/pages/invoices/common/components/TasksTabLabel.tsx
@@ -37,7 +37,7 @@ export function TasksTabLabel({ lineItems }: Props) {
       <div className="flex space-x-2">
         <span>{t('tasks')}</span>
 
-        <span className="font-medium">({Math.floor(totalQuantity)} h)</span>
+        <span className="font-medium">({Math.round(totalQuantity)} h)</span>
       </div>
     );
   }


Civolilah:
> @Civolilah is floor the optimal approach here?
> 
> ie
> 
> ```js
> Math.floor(4.9);   // 4
> Math.floor(4.1);   // 4
> Math.floor(4.0);   // 4
> ```

@turbo124 Hmm, I also wondered about the same question, if it's 1.2 hours, is it optimal to show 2 hours? So we had to decide whether to display the nearest lower or higher value. Yeah, I understand the point, but at the same time, do we want to display 5 hours for a user if they spent 4.1? They might ask themselves, "But I didn't spend 5 hours?"

Civolilah:
> i would have thought that we would round to the _nearest_ value?
> 
> ie 1.1 => 1 1.9 => 2

@turbo124 Hmm, yeah, I think the same. I've updated the solution to round to the nearest value, 1.4 = 1, but 1.7 = 2. I think that's good enough.

PR #3002, opened 2026-02-27
[Bugfix] Group Settings Configuration

Civolilah:
@beganovich @turbo124 The PR includes fixes for group settings configuration, disabling the language and currency selector, and the title for the localization card. After testing a lot of scenarios, I was not able to recreate the case when `company_settings` are sent in the payload of the `group_settings` endpoint. However, after a lot of research, I think that most likely a Redux update/rendering issue caused this. So, I made some logic updates to prevent this and ensure that the payload is always correct. Screenshot:

Screenshot 2026-02-27 at 20 36 41

Let me know your thoughts.

### DIFF
--- a/src/pages/settings/group-settings/common/hooks/useConfigureGroupSettings.ts
+++ b/src/pages/settings/group-settings/common/hooks/useConfigureGroupSettings.ts
@@ -48,9 +48,9 @@ export function useConfigureGroupSettings(params?: Params) {
     );
 
     if (!withoutNavigation) {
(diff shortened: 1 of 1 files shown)


turbo124:
@beganovich for review, edge case was seen where company settings were pushed to group settings, the issue is not 100% recreatable but we have logging which shows this is occurring. This PR addresses this issue; for your review (and merge to main pls)

beganovich (approved):
Thanks 👍

PR #3006, opened 2026-03-04
[Bugfix] Unnecessary Opening Send Time Modal | Client Settings

Civolilah:
@beganovich @turbo124 The PR fixes an unnecessary opening of the send time modal for client settings configuration. The reason the modal opened unnecessarily was due to the condition we had for comparing changes to the current company configuration:

```
if (
      currentCompany?.settings.entity_send_time !==
      company?.settings.entity_send_time
    ) {
```

So, the current changes are `undefined` because the checkbox is still not checked, but the current company has a numeric value. Once we compare them, they are different, and the modal shows. Now, this has been fixed.

Let me know your thoughts.

### DIFF
--- a/src/pages/settings/email-settings/EmailSettings.tsx
+++ b/src/pages/settings/email-settings/EmailSettings.tsx
@@ -73,8 +73,9 @@ export function EmailSettings() {
 
   const handleOnSaveClick = () => {
     if (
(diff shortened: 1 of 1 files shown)


beganovich (approved):
Thanks 👍

PR #3049, opened 2026-03-17
[Cleanup] Admin Toggle Behavior Changes | User Permissions

Civolilah:
@beganovich @turbo124 The PR includes the removal of logic that turned on all permissions when "Admin" was enabled. Now, when "Admin" is turned on, the rest of the permissions remain available for the same action without being disabled. Screenshot:

Screenshot 2026-03-17 at 14 37 42

Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #3052, opened 2026-03-19
[Bugfix] Changing Client/Vendor When Entity Is Cloned Fixes

Civolilah:
@beganovich @turbo124 The PR includes fixes for not properly applying client/vendor contacts when the client or vendor is changed after triggering the clone action. Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #3053, opened 2026-03-19
Merge company translations to main i18n

beganovich:
This merges `company.settings.translations` into main i18n.


turbo124:
image

seeing circular reference issue, here i tried to change the quotes label to Estimates

this only succeeded in changing the localization section translation, but didn't impact sidebar translation.

image


Not sure there is a nice way to implement this with the impact that would be expected.

PR #3072, opened 2026-03-27
[Bugfix] Payload For Recovering Password Form

Civolilah:
@beganovich @turbo124 The PR includes fixes for the payload in the recover password form. I was not able to recreate this issue, but I assumed what could happen here and made some fixes. Let me know your thoughts.


turbo124:
@Civolilah i don't think this works.

what are we not using

```
value={form.values.email}
````

so that all autofill input is captured and updated as needed?

we'll also want to consider any other uncontrolled form inputs where autofill may be available (register.tsxcomponents) and update accordingly.

finally, the root issue I think is using onFocus, ie type an email address and hit enter, because we do not lose focus the form field does not update formik

beganovich (approved):
Thanks 👍

PR #3078, opened 2026-03-30
[Bugfix] Expense Amount Calculations

Civolilah:
@beganovich @turbo124 The PR includes expense amount calculation fixes for the add_to_invoice action, matching the invoicing expense action. Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #3080, opened 2026-04-01
[Bugfix] Expense Category Creation By Clicking 'Enter'

Civolilah:
@beganovich @turbo124 The PR includes changes for overriding the `onBlur` event, which we mostly use throughout the app, and in this case, leaving the `onChange` function to manage changes. Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #3082, opened 2026-04-01
[Bugfix] Cloning Expense To Recurring Expense

Civolilah:
@beganovich @turbo124 The PR includes setting frequency_id to a default value when cloning an expense to a recurring expense. Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #3083, opened 2026-04-02
Reimplement Playwright Tests

turbo124:
Adds back in the necessary fixes for Playwright tests to pass.

Also include a bonus fix of expense category create component issue


beganovich (approved):
Thanks 👍

PR #3085, opened 2026-04-02
[Bugfix] Pre Selection Invoices And Payments

Civolilah:
@beganovich @turbo124 The PR includes the addition of pre-selection logic for invoices and payments for matched transactions. Let me know your thoughts.


turbo124:
@Civolilah i'm not seeing this work as expected for transactions that have been matched?

image

Civolilah:
@turbo124 As we discussed on Slack, the functionality has been fixed by adding a "with" query parameter to return all needed entities in the response and pre-select them. Let me know your thoughts.

turbo124:
@Civolilah this works, as a result instead of using invoice #1, we should have the related invoice.number prop available?

src/pages/transactions/components/TransactionMatchDetails.tsx
```diff
@@ -213,6 +215,18 @@ export function TransactionMatchDetails(props: Props) {
...
+    const { invoice_ids, payment_id } = props.transactionDetails;
+
+    if (invoice_ids) {
+      setInvoiceIds(invoice_ids.split(',').filter(Boolean));
+    }
+
+    if (payment_id) {
+      setPaymentIds([payment_id]);
+    }
+  }, [props.transactionDetails.transaction_id]);
```
beganovich, line 228: Please just confirm this won't mess up flow since we don't have invoice_ids/payment_id in the `useEffect` dependencies.

beganovich (approved):

src/pages/transactions/components/TransactionMatchDetails.tsx
Civolilah: @beganovich No, it will not, no worries. So, `invoice_ids` as well as `payment_id` are part of the transaction query response. So, once `transaction_id` has been changed, everything else is changed at the same time.

```
transactionDetails={{
            base_type: transaction?.base_type || '',
            transaction_id: transaction?.id || '',
            status_id: transaction?.status_id || '',
            invoice_ids: transaction?.invoice_ids || '',
            payment_id: transaction?.payment_id || '',
          }}
```

beganovich (approved):

PR #3088, opened 2026-04-02
Add groupby for reports

turbo124:
Add support for group by in reports allowing aggregation by user selected column


beganovich (approved):
Thanks 👍

PR #3091, opened 2026-04-03
[Bugfix] Reseting Status Selector Value | Reports

Civolilah:
@beganovich @turbo124 The PR includes fixes for resetting the status selector value on reports when the report has been changed. Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #3093, opened 2026-04-06
[Bugfix] Tax Rates List | Quickbooks Connection Issue

Civolilah:
@beganovich @turbo124 The PR includes the removal of the UI that was displayed after Quickbooks was connected on the tax_settings page. So, from now on, we always display the same UI regardless of whether Quickbooks is connected or not. Screenshot:

Screenshot 2026-04-06 at 19 18 09

Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #3099, opened 2026-04-08
[Feature] Implementing CC Only Toggle For Client And Vendor Contacts

Civolilah:
@beganovich @turbo124 The PR includes the implementation of the "CC Only" toggle for client and vendor contacts. In addition, it adds logic to hide contacts that have this cc_only property set to true on the create and edit pages. Screenshot:

Screenshot 2026-04-08 at 11 09 35

Let me know your thoughts.


Civolilah:
@turbo124 Currently, I could not test everything fully as, for me, the endpoint does not save the cc_only property. So, let me know if I missed anything.

beganovich (approved):

PR #3105, opened 2026-04-10
[Feature] Adding Name Field For Schedulers

### LINKED ISSUE #3097: Custom title for Scheduled Reports
**What version of Invoice Ninja are you running? ie v4.5 / v5**
Latest (hosted)

**Have you searched existing issues/requests?**
Yes

**Screenshots**

Image

Image

**Additional context**
Hello, I would like to request the ability to add custom title for scheduled reports. We started using them and I immediately realized that it would get messy real quick. 

It would be just one more field

Thank you for your consideration

Civolilah:
@beganovich @turbo124 The PR includes the addition of a name field for schedulers, except that instead of a custom value on the table, we display the 'name' property value. Screenshot:

Screenshot 2026-04-10 at 20 50 35

Closes #3097 

Let me know your thoughts.


Civolilah:
@turbo124 I hope that I did not miss the point. The client requested a field, per your comment, it sounds to me as confirmation of it. If we should only display the name property on the table or similar things, please let me know.

turbo124:
@Civolilah as long as we use schedule.name as the value on the list view, this is correct.

turbo124:
@Civolilah we should still use the UI generated placeholder for NEW schedules. This certainly should be kept, however, if a user manually overrides, then their name should be preferred.

Civolilah:
> @Civolilah we should still use the UI generated placeholder for NEW schedules. This certainly should be kept, however, if a user manually overrides, then their name should be preferred.

@turbo124 Ah okay, got you! I returned the old logic, but now we have a name field which, if populated, will be displayed in the table. Let me know your thoughts.

turbo124:
@Civolilah 

image

When using the system generated name, we don't see it in the form, perhaps we could generate it and set it as placeholder text?

Civolilah:
> @Civolilah
> 
> image
> When using the system generated name, we don't see it in the form, perhaps we could generate it and set it as placeholder text?

@turbo124 Yes, you are right, it will be better to match the table with the logic. So, I've added a fallback template value to the field on the form. Screenshot:

Screenshot 2026-04-15 at 13 15 22

Let me know your thoughts.

beganovich (approved):
Thanks 👍

PR #3108, opened 2026-04-12
[Feature] Adding Additional Project Tabel View Columns

Civolilah:
@beganovich @turbo124 The PR includes the implementation of additional columns for the Project table. The important thing to mention is that we only apply those query parameters as they are needed (when columns are selected), as they are expensive. Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #3109, opened 2026-04-14
Add Item Net Cost as selectable product column for PDF

turbo124:
Adds an option in the Invoice Design to assign the net item cost as a column on the invoice design.


beganovich (approved):
Thanks 👍 

Do you want to merge this or wait for https://github.com/invoiceninja/invoiceninja/discussions/11877#discussioncomment-16553005?

PR #3111, opened 2026-04-14
[Cleanup] Dashboard Card Addition Improvements

Civolilah:
@beganovich @turbo124 The PR includes improvements to dashboard cards, specifically for adding them. Screenshot:

Screenshot 2026-04-16 at 19 27 08

Let me know your thoughts.


turbo124:
@Civolilah 

when adding a new card, we should always have one selected.

image

Civolilah:
> @Civolilah
> 
> when adding a new card, we should always have one selected.
> 
> image

@turbo124 I improved the logic here, by default, we have a field selected when opening the modal. However, if we want to keep that behavior, we need to avoid allowing multiple fields to be added by opening the modal only once. This means the user would have to remove the default field each time if they don’t want to add that one. So, I think preventing the addition of multiple fields avoids a lot of complications. Screenshot:

Screenshot 2026-04-19 at 20 51 35

Let me know your thoughts.

turbo124:
image

I think one item we are missing here is the actual date range, it look like in the payload you are sending the start/end dates could you include these on a new line, formatted based on localization.

Civolilah:
> image
> I think one item we are missing here is the actual date range, it look like in the payload you are sending the start/end dates could you include these on a new line, formatted based on localization.

@turbo124 Yes, I agree with you. I played with the UI a little bit here. I hope you like the changes. Screenshot:

Screenshot 2026-04-20 at 22 32 19

Let me know your thoughts.

turbo124:
@Civolilah 

Thanks, ui looks great!

i see some issues.

1. Currency formatting

image

For users with multiple currencies, the currency symbol is not updating on the label. for the All currency option, this should be the company currency.

The other issues i've just realised is the confusion that is caused by the current label system:

image

Current Period is misleading, the cards are calculated based on the selected period (this month, last month)

So current period is always "relative" to what has been selected, and is only relevant when this month is selected.

Because of this, I think we should only show the date range when period == current period. the user will then understand the current period, and then know the previous period is relative to current period.

Civolilah:
> @Civolilah
> 
> Thanks, ui looks great!
> 
> i see some issues.
> 
> 1. Currency formatting
> 
> image
> For users with multiple currencies, the currency symbol is not updating on the label. for the All currency option, this should be the company currency.
> 
> The other issues i've just realised is the confusion that is caused by the current label system:
> 
> image
> Current Period is misleading, the cards are calculated based on the selected period (this month, last month)
> 
> So current period is always "relative" to what has been selected, and is only relevant when this month is selected.
> 
> Because of this, I think we should only show the date range when period == current period. the user will then understand the current period, and then know the previous period is relative to current period.

@turbo124 Yeah, really good catches. I think both of the mentioned things are covered right now. Screenshot:

Screenshot 2026-04-21 at 22 17 44

Let me know your thoughts.

turbo124:
looks great!

Civolilah:
@turbo124 As this PR was already active and not merged, I've improved the functionality as explained in: https://invoiceninja.atlassian.net/browse/RU-2248. The modal has been organized so that we have the current cards on the left side with drag-and-drop functionality. On the right side, we have the possibility of adding cards, with a message if that card combination has already been added. I hope that I have covered all cases. Screenshots:

Screenshot 2026-04-25 at 22 41 32


Screenshot 2026-04-25 at 22 41 18

Let me know your thoughts.

turbo124:
ping @beganovich ready now :)

src/pages/dashboard/components/DashboardCardSelector.tsx
```diff
@@ -102,21 +283,64 @@ export function DashboardCardSelector() {
...
+    requestAnimationFrame(() => {
+      scrollContainerRef.current?.scrollTo({
+        top: scrollContainerRef.current.scrollHeight,
+        behavior: 'smooth',
+      });
+    });
+  }, [
+    selectedField,
+    isDuplicate,
+    fields,
```
beganovich: Since `fields` is part of this dependencies, `handleAdd` gets recreated every render, right?

src/pages/dashboard/components/DashboardCardSelector.tsx
```diff
@@ -52,27 +61,199 @@ export const FIELDS_LABELS: Record<string, string> = {
...
+          </span>
+
+          <span className="text-xs truncate" style={{ color: colors.$17 }}>
+            {t(
+              decoded.period === 'current'
+                ? 'current_period'
+                : decoded.period === 'previous'
+                ? 'previous_period'
+                : 'total'
+            )}
```
beganovich: This is weird and bad.

```tsx
const PERIOD_LABEL_MAP: Record<Period, string> = {
  current: 'current_period',
  previous: 'previous_period',
  total: 'total',
};

{t(PERIOD_LABEL_MAP[period])}
```

beganovich (approved):
Looks good, couple things to clear up 👍

src/pages/dashboard/components/DashboardCardSelector.tsx
Civolilah: @beganovich Hmm, yeah, it will be recreated, but at the same time, we always need correct field values. So, to avoid recreation while always keeping correct usage of field values, I used a ref. Let me know your thoughts.

src/pages/dashboard/components/DashboardCardSelector.tsx
Civolilah: @beganovich Yeah, I agree with you. I improved it by using an object.

beganovich (approved):
Thanks 👍

PR #3115, opened 2026-04-19
[Cleanup] Refetching Invoices Query While Listening to InvoiceWasPaid

Civolilah:
@beganovich @turbo124 This PR includes a cleanup of the logic while listening for the `InvoiceWasPaid` event. So instead of displaying any warning banners, we refetch the query to display changes on the UI (status banner, etc.). Let me know your thoughts.

### DIFF
--- a/src/common/hooks/useGetSetting.ts
+++ b/src/common/hooks/useGetSetting.ts
@@ -27,10 +27,6 @@ export function useGetSetting({
   return (client: Client | undefined, propertyKey: keyof Settings) => {
     if (groupSettings && client) {
       if (client.settings[propertyKey] !== undefined) {
-        if (import.meta.env.VITE_IS_TEST === 'true') {
-          return `Client: ${client.settings[propertyKey]}`;
-        }
-
         return client.settings[propertyKey];
       }
 
@@ -38,10 +34,6 @@ export function useGetSetting({
         client.group_settings &&
         client.group_settings.settings[propertyKey] !== undefined
       ) {
-        if (import.meta.env.VITE_IS_TEST === 'true') {
-          return `Group: ${client.group_settings.settings[propertyKey]}`;
-        }
-
         return client.group_settings.settings[propertyKey];
       }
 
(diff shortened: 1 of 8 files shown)


turbo124:
@Civolilah on payment, we should still show the banner, but also refetch query.

Civolilah:
> Member
> ###

@turbo124 Okay, sure. Displaying the banner has been returned, but as we discussed, instead of a warning banner, we are just displaying an info banner (blue color). We just need a little update here. The current 'invoice_status_changed' translation keyword reads as: 'Please note that the status of your invoice has been updated. We recommend refreshing the page to view the most current version.' So we just need a small change in the text. Let me know your thoughts.

turbo124:
@Civolilah 

when websocket === InvoiceWasPaid, we can show banner with this translation string:

invoice_status_paid

also just to clarify, we are ONLY invalidating the single invoice id from query? one edge case we want to ensure, is that if a user is working on a different invoice, that invalidation does not wipe their data that they are in the process of updating...

Civolilah:
> @Civolilah
> 
> when websocket === InvoiceWasPaid, we can show banner with this translation string:
> 
> invoice_status_paid
> 
> also just to clarify, we are ONLY invalidating the single invoice id from query? one edge case we want to ensure, is that if a user is working on a different invoice, that invalidation does not wipe their data that they are in the process of updating...



> @Civolilah
> 
> when websocket === InvoiceWasPaid, we can show banner with this translation string:
> 
> invoice_status_paid
> 
> also just to clarify, we are ONLY invalidating the single invoice id from query? one edge case we want to ensure, is that if a user is working on a different invoice, that invalidation does not wipe their data that they are in the process of updating...

@turbo124 Yeah, I understand your points. So, I've changed the translation, and additionally, it was not the case earlier that we only invalidated the query from a specific invoice, it was all invoice queries. Currently, I've ensured that only that specific invoice query has been invalidated, so the user cannot experience that edge case you mentioned. Let me know your thoughts.

Civolilah:
@turbo124 As we discussed on Slack, the PR is updated with the correct invalidation, we are now invalidating the invoice from which the socket is triggered, as well as list queries. Also, the `VITE_IS_TEST` environment variable has been removed from the `useGetSetting` hook, as we discussed on Slack, to avoid confusion and incorrect responses. Let me know your thoughts.

turbo124:
@beganovich for this release please.

beganovich (approved):
Thanks 👍

PR #3116, opened 2026-04-19
[Cleanup] Decimal Filter Input Normalization

Civolilah:
@beganovich @turbo124 The PR includes normalization of EU-style decimal input in filter fields. When the input matches a digits-comma-digits pattern (e.g. `23,32`), the comma is automatically replaced with a dot before passing the value to the filter handler. All other inputs remain unaffected. Let me know your thoughts.

### DIFF
--- a/src/components/DataTable.tsx
+++ b/src/components/DataTable.tsx
@@ -378,6 +378,10 @@ export function DataTable<T extends object>(props: Props<T>) {
     customFilters,
   });
 
+  const normalizeNumericCommas = (value: string): string => {
+    return value.replace(/(\d),(\d)/g, '$1.$2');
+  };
+
   useEffect(() => {
     const queryParam = buildDateRangeQueryParameter(dateRangeEntries);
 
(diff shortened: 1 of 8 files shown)


turbo124:
@Civolilah 

this works, however the filter transforms afterwards:

image


Ideally we would preserve the comma in the filter input.

Civolilah:
> @Civolilah
> 
> this works, however the filter transforms afterwards:
> 
> image
> Ideally we would preserve the comma in the filter input.

@turbo124 I understand your point. I improved the behavior in the direction you suggested. Please let me know your thoughts.

turbo124:
I understand the direction here. But i'm just not sure disabling the comma is bueno for this.

Civolilah:
> I understand the direction here. But i'm just not sure disabling the comma is bueno for this.

@turbo124 So, as we discussed on Slack, the best solution here is to normalize the value that is going into the query params. So, that is what I did here, and it works perfectly fine, avoiding any complications. Let me know your thoughts.

beganovich (approved):
Thanks 👍

PR #3117, opened 2026-04-19
Add password protection timeout variable back into UI

turbo124:
Add the default password protection timeout selector back  into the UI.


beganovich (approved):
Thanks 👍

PR #3127, opened 2026-04-26
[Cleanup] Refactoring Templates And Reminders

Civolilah:
@beganovich @turbo124 The PR includes a refactoring of the UI for templates and reminders, making it much easier to pick the template you want to adjust or edit. Screenshot:

Screenshot 2026-05-02 at 19 52 33

Let me know your thoughts.


Civolilah:
@turbo124 I've updated the PR with the redesign that we discussed on Slack. Please let me know if this direction is better.

turbo124:
@Civolilah i do think this is much better, i think we need to align the UI close to tabs thou, at least for the top row:

image


in practice, they don't have to be "TABS" but the UI does :)

Civolilah:
@turbo124 So, as we discussed on Slack, we will make the first row as tabs, but the second one as some kind of button to pick the variant of reminder or payment. Screenshots:

Screenshot 2026-05-05 at 18 24 01

Screenshot 2026-05-05 at 18 24 10

Let me know your thoughts.

turbo124:
lgtm!

beganovich (approved):
Thanks 👍

PR #3130, opened 2026-04-28
Option for attaching audit logs to outbound client emails

turbo124:
@beganovich we'll need this for next release pls.


beganovich (approved):
Thanks 👍

PR #3141, opened 2026-05-04
[Cleanup] Adding Actions For Templates

Civolilah:
@beganovich @turbo124 The PR includes the implementation of actions on the edit page of the Template. Screenshot:

Screenshot 2026-05-06 at 21 27 38

Let me know your thoughts.

### DIFF
--- a/src/pages/documents/pages/blueprints/builder/BlueprintBuilder.tsx
+++ b/src/pages/documents/pages/blueprints/builder/BlueprintBuilder.tsx
@@ -14,6 +14,7 @@ import { Card } from '$app/components/cards';
 import { Button, InputField, SelectField } from '$app/components/forms';
 import { Default } from '$app/components/layouts/Default';
 import { Modal } from '$app/components/Modal';
+import { ResourceActions } from '$app/components/ResourceActions';
 import { TabGroup } from '$app/components/TabGroup';
 import {
   Alertbox,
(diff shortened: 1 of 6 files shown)


Civolilah:
@turbo124 We are missing translations for `restored_template`, `archived_template`, and `deleted_template`. So, if we are going to keep those, please just add them to the files.

turbo124:
@Civolilah we are missing archive/restore from list view?

image

Civolilah:
> @Civolilah we are missing archive/restore from list view?
> 
> image

@turbo124 Yes, you are right, we missed them because the DocuNinja API returns a different data type for `archived_at` and similar properties, and the current table logic did not display them. So, I handled the display through custom actions. Screenshot:

Screenshot 2026-05-07 at 11 59 20

Let me know your thoughts.

src/pages/documents/pages/blueprints/builder/BlueprintBuilder.tsx
```diff
@@ -268,11 +271,17 @@ function BlueprintBuilder() {
   const { id } = useParams();
   const colors = useColorScheme();
 
+  const { data: blueprintResponse } = useBlueprintQuery({ id });
+
+  const [isEditModalOpen, setIsEditModalOpen] = useState<boolean>(false);
   const [isDocumentSaving, setIsDocumentSaving] = useState<boolean>(false);
-  const { data: blueprintResponse, isLoading } = useBlueprintQuery({ id });
 
   const [blueprint, setBlueprint] = useState<Blueprint>();
```
beganovich, line 279: Just saw this - was wondering can't we consume the `blueprintResponse` directly instead of this double-state tracking?

Civolilah: @beganovich Yes, we can do that since we never use the setter. Let me know your thoughts.

beganovich (approved):

PR #3145, opened 2026-05-06
[Bugfix] Payment Date Calculation

Civolilah:
@beganovich @turbo124 The PR includes fixes for the presentation of date values for payments. Except on the Payment edit page, we display the payment date on the invoice slider, so I assumed we should fix it there as well. Please let me know if there is a different case for 'paymentables'.


turbo124:
@Civolilah we need to change this. this will force all timestamps to use UTC, this may not necessarily be the desired outcome.

we need to create a _new_ method that selectively changes just this timestamp value for the payment overview created_at timestamp please.

turbo124:
@beganovich for merge to main please.

beganovich (approved):

PR #3147, opened 2026-05-06
[Cleanup] Implementing Docuninja Document Filters

Civolilah:
@beganovich @turbo124 The PR includes the implementation of DocuNinja document filters. Screenshot:

Screenshot 2026-05-06 at 22 28 56

Let me know your thoughts.

### DIFF
--- a/src/common/helpers/payment.ts
+++ b/src/common/helpers/payment.ts
@@ -1,0 +1,16 @@
+import dayjs from 'dayjs';
+import utc from 'dayjs/plugin/utc';
+
+dayjs.extend(utc);
+
+export function dateUTC(date: number | string, format: string) {
+  if (date === 0 || date === '' || date === undefined) {
+    return '';
+  }
+
+  if (typeof date === 'number') {
+    return dayjs.unix(date).utc().format(format);
+  }
+
+  return dayjs(date).utc().format(format);
+}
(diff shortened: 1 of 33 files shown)


turbo124:
@Civolilah looks good, should the initial default 'active' be selected on first load?

Civolilah:
> @Civolilah looks good, should the initial default 'active' be selected on first load?

@turbo124 Hmm, yes, I think it will be better to have 'active' selected by default, so I did it. But the reason it was not is that those filters are not regular active/archived/deleted, we have a lot more of them. But yeah, it is better to have something by default. Let me know your thoughts.

turbo124:
@Civolilah actually, after testing this, i see a problem.

we are combining active/archived/deleted into the same group here.

we need to split active/archived/deleted into their own selector.

Civolilah:
> @Civolilah actually, after testing this, i see a problem.
> 
> we are combining active/archived/deleted into the same group here.
> 
> we need to split active/archived/deleted into their own selector.

@turbo124 Yeah, I'm aware that it would be a much better convention for Invoice Ninja, but should we then combine values into one query parameter 'status', or should it be different for the rest of the status values?

Civolilah:
@turbo124 As we discussed on Slack, we separated the filtering into two dropdowns to align with the convention. But we still merge it into one query parameter "status". Screenshot:

Screenshot 2026-05-11 at 21 58 17

Let me know your thoughts.

beganovich (approved):
Thanks 👍

PR #3149, opened 2026-05-09
[Bugfix] Reloading Beta Screen Fixes

Civolilah:
@beganovich @turbo124 Initially, the user does not have a DocuNinja account, so the login route returns an error and we redirect to `/docuninja/beta?ie=true`. However, this query parameter forces the beta join screen to display regardless of whether the user has already joined. The same issue occurs when a user joins for the first time, the page reloads but retains the `?ie=true` parameter, which causes the join screen to appear again as if the enrollment was unsuccessful. The fix is simply to reload the page without this parameter. Let me know your thoughts.


turbo124:
@beganovich unable to test this one, but looks good.

beganovich (approved):
Thanks 👍

PR #3154, opened 2026-05-11
[Cleanup] Stripe Integration Helper Modal

Civolilah:
@beganovich @turbo124 The PR includes implementation of a Stripe helper modal when the Stripe configuration is done and the user is redirected to the edit page. The modal only displays if the URL contains `show_onboarding=true`. The modal explains what the user can additionally configure, and once clicked, the user will be navigated to another tab. Screenshot:

Screenshot 2026-05-11 at 22 51 34

Let me know your thoughts.

### DIFF
--- a/src/pages/settings/gateways/edit/Edit.tsx
+++ b/src/pages/settings/gateways/edit/Edit.tsx
@@ -42,7 +42,7 @@ export function Edit() {
   const { documentTitle } = useTitle('edit_gateway');
 
   const [t] = useTranslation();
-  const [searchParams] = useSearchParams();
+  const [searchParams, setSearchParams] = useSearchParams();
 
   const { id } = useParams();
   const actions = useActions();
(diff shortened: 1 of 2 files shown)


src/pages/settings/gateways/edit/Edit.tsx
```diff
@@ -112,6 +115,12 @@ export function Edit() {
     }
   }, [gateway]);
 
+  useEffect(() => {
+    if (searchParams.get('show_onboarding') === 'true' && companyGateway) {
+      setShowOnboardingModal(true);
+    }
+  }, [companyGateway, searchParams]);
+
```
beganovich, line 131: Do we need to clean up param after successful onboard? 

This means if person refreshes the page and/or navigates back, dialog will show, right?

Civolilah: @beganovich Yes, I agree, the logic for parameter removal has been added. Let me know your thoughts.

beganovich (approved):

PR #3156, opened 2026-05-12
Update documents redirects

beganovich:
This redirects directly to builder if status is draft & we have no signatories. 

Otherwise it goes to timeline view.


turbo124:
Merge when happy @beganovich

PR #3162, opened 2026-05-16
[Bugfix] Templates And Reminder Behavior When Configuring Settings

Civolilah:
@beganovich @turbo124 The PR includes logic fixes for templates and reminders when configuring client/group settings, enabling each template or reminder separately, making tabs selectable and never disabled, but only fields if a specific template or reminder is not selected. Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #3163, opened 2026-05-16
[Cleanup] Payment Terms Tooltip | Invoices & Quotes

Civolilah:
@beganovich @turbo124 The PR includes implementation of a payment terms tooltip for invoices and quotes on the edit page, showing payment terms settings details along with a configure link if the user wants to add more payment terms. Screenshot:

Screenshot 2026-05-16 at 17 58 09

Screenshot 2026-05-16 at 17 58 17

Let me know your thoughts.

### DIFF
--- a/src/components/PaymentTermsTooltip.tsx
+++ b/src/components/PaymentTermsTooltip.tsx
@@ -1,0 +1,67 @@
+/**
+ * Invoice Ninja (https://invoiceninja.com).
+ *
+ * @link https://github.com/invoiceninja/invoiceninja source repository
+ *
+ * @copyright Copyright (c) 2022. Invoice Ninja LLC (https://invoiceninja.com)
+ *
+ * @license https://www.elastic.co/licensing/elastic-license
+ */
+
+import { useColorScheme } from '$app/common/colors';
+import { useGetSetting } from '$app/common/hooks/useGetSetting';
+import { Client } from '$app/common/interfaces/client';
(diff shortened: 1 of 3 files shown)


turbo124:
@Civolilah the link should provide the ability to navigate to the CLIENTS edit page where they can set their preferred payment terms for this particular client.

image

Civolilah:
> @Civolilah the link should provide the ability to navigate to the CLIENTS edit page where they can set their preferred payment terms for this particular client.
> 
> image

@turbo124 Ah, okay, got you. Navigation has been changed to the `clients/:id/settings` page, where the payment terms select field has been placed. Let me know your thoughts.

turbo124:
@Civolilah just need to be careful that we don't display the (i) without having the client record.

ie. when cloning / creating a new invoice prior to saving, we have no client record, so the payment terms just show as none... which may not actually be correct.

image

Civolilah:
> @Civolilah just need to be careful that we don't display the (i) without having the client record.
> 
> ie. when cloning / creating a new invoice prior to saving, we have no client record, so the payment terms just show as none... which may not actually be correct.
> 
> image

@turbo124 Ah yes, sorry, my bad. So, if we have a client object, we will display an icon with details in a tooltip. If we do not have one, we will completely hide the icon. Screenshot:

Screenshot 2026-05-19 at 15 41 18

Let me know your thoughts.

beganovich (approved):

PR #3164, opened 2026-05-19
[Cleanup] Sum Columns | Reports

Civolilah:
@beganovich @turbo124 The PR includes implementation of sum columns for reports. Let me know your thoughts.

### DIFF
--- a/src/pages/reports/common/components/SortableColumns.tsx
+++ b/src/pages/reports/common/components/SortableColumns.tsx
@@ -324,7 +324,7 @@ export function SortableColumns({ report, columns }: Props) {
 
   const colors = useColorScheme();
 
-  const { update } = usePreferences();
+  const { update, preferences } = usePreferences();
 
   const { data: persistedData, defaultColumns } = useColumns({
     report,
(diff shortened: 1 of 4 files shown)


src/pages/reports/common/components/SortableColumns.tsx
```diff
@@ -339,9 +339,13 @@ export function SortableColumns({ report, columns }: Props) {
 
   const syncToPreferences = useCallback(
     (newData: Record[][]) => {
+      if (Array.isArray(preferences.reports.columns)) {
+        update('preferences.reports.columns', {});
+      }
```
beganovich, line 344: This looks very destructive. Won't this wipe EVERY report configuration each time this function runs?

Civolilah: @beganovich Actually, this is not part of the functionality at all, this is a bug fix that I can always recreate locally on develop. By default, columns are an empty object:

```
reports: {
    columns: {},
  },
```

but for some reason that I'm completely not aware of, we have an empty array as the value of the 'columns' property.

Maybe it's because I was developing some functionalities locally and saved it as an empty array, and right now it's causing an issue (just for me as user).

But I couldn't leave it just like that. I think it should not be the case that PHP returns an empty array instead of an empty object, because this is a whole JSON where we send everything we want and it just saves.

In this case, can you please test the 'Customize Columns' functionality on develop?

beganovich (approved):

PR #3171, opened 2026-05-22
[Cleanup] Filtering Active Payment Terms

Civolilah:
@beganovich @turbo124 The PR includes filtering active payment terms instead of displaying all of them for selectors. Let me know your thoughts.


turbo124:
this is better, but then it does lead to issues like this:

image


most likely this is just a compromise for the given issue we are dealing with here thou..

Civolilah:
> this is better, but then it does lead to issues like this:
> 
> image
> most likely this is just a compromise for the given issue we are dealing with here thou..

@turbo124 Okay, I understand the issue of duplicating payment terms with the same number of days. I have made a solution. Let me know your thoughts.

turbo124:
@Civolilah this reverts everything back to the precommit experience here.

For example, if I have a client with 30 day payment terms. Then Archive the 30 day option.

The clients option never appears in the selector.

image

image

Civolilah:
> @Civolilah this reverts everything back to the precommit experience here.
> 
> For example, if I have a client with 30 day payment terms. Then Archive the 30 day option.
> 
> The clients option never appears in the selector.
> 
> image image

@turbo124 Ah okay, I got you. So, if a payment term is selected on a client and then archived/deleted, it will still appear in the dropdown. But if you go to any other client where that specific payment term is not selected or a different one is selected, it will not be displayed in the dropdown at all. Let me know your thoughts.

beganovich (approved):
Thanks 👍

Issue #3175, opened 2026-05-26
Login without network connectivity results in misleading error message


ludgerey:
### Setup
- Version: 03.05.2026.1

### Describe the bug

Trying to login in without a network connectivity results in a `These credentials do not match our records` error, which is highly misleading, as user will expect that the values have been checked and that the validation has failed. In reality nothing has reached the server at all.

### Steps To Reproduce
* Go to `/login`
* Go offline (e.g. via Developer Console and select `Offline` in network throughput selection)
* Enter some credentials and press "Login"
* See `These credentials do not match our records`

### Expected Behavior
A distinguishable error message when login failed without any actual login attempt. E.g. `An error occurred during login. Try again later.`.

Preferably "no network" get its own message `Server not reachable. Check your internet connection and try again later.`

### Screenshots

<img width="497" height="753" alt="Image" src="https://github.com/user-attachments/assets/e109fd6c-a1fc-4804-be96-4a163a29d567" />

ludgerey:
I added a PR: https://github.com/invoiceninja/ui/pull/3176

PR #3176, opened 2026-05-26
Fix login without network connectivity results in misleading error message

ludgerey:
Fixes [Login without network connectivity results in misleading error message #3175]( https://github.com/invoiceninja/ui/issues/3175)

* Use a generic fallback message for login failures, in case we get a response, but message body has no `message`. I reused the existing `error_refresh_page` translation key. As this is really an edge case I think the message fits good.
* Add a dedicated error message for server not reachable, in case we get no response (DNS failures, Port not open, Server not running, Wifi disconnected, ...). I added a new translation key `server_not_reachable` for `en` and `de`. The error message is shown as a toast message. It works on any request, not just login, as I added it in `request.ts` helper file.

Bildschirmfoto 2026-05-26 um 18 48 47xx

Bildschirmfoto 2026-05-26 um 18 45 05

Bildschirmfoto 2026-05-26 um 18 51 50

In case you wonder, the credential not valid message still appears as a server generated message:
Bildschirmfoto 2026-05-26 um 18 52 57


src/common/helpers/request.ts
```diff
@@ -43,6 +43,11 @@ client.interceptors.response.use(
   (error: AxiosError<ValidationBag>) => {
     const url = error.response?.config.url;
 
+    if (!error.response) {
+      toast.error('server_not_reachable');
+      return Promise.reject(error);
+    }
+
```
turbo124: I think this is too narrow, I would think error.code would be more appropriate and switch on it based on:

```
ERR_NETWORK
ECONNABORTED (timeout)
ERR_CANCELED
```

ludgerey: Yes, reporting timeouts and especially cancelled requests as server not reachable could lead to false assumptions.

I changed the code so that only `ERR_NETWORK` is treated as `server_not_reachable`. Timeouts and cancellations remain their previous behavior.

I thought about giving timeouts a dedicated error message too, but against it because it's more of an exceptional case. What you think?

Cancellations should be kept silent, that's good that we exclude that from reporting.

turbo124:
Thanks!

beganovich (approved):
Thanks!

PR #3177, opened 2026-05-27
Fix german translation for key `calculate_taxes_help`

ludgerey:
Fix german translation for key `calculate_taxes_help`

Before:
```
"calculate_taxes_help":"Archivieren Sie Angebote automatisch, wenn sie in Rechnungen umgewandelt werden."
```
(It's copy pasted from `auto_archive_quote_help`, so it's an unrelated text)

After:
```
"calculate_taxes_help":"Steuern beim Speichern von Rechnungen automatisch berechnen"
```
(Directly translated from english)


turbo124:
@ludgerey thanks for this, we manage all translations here:

https://app.transifex.com/invoice-ninja/

I've updated this particular key myself, feel free to update any invalid ones directly from transifex going forward.

PR #3178, opened 2026-05-28
Add paymentable for invoice payment resolution.

turbo124:
Changes the functionality around displaying an invoices payment from mthe Invoice context.

Previously we passed in the Payments, but this was not always displaying the correct data (ie applying an amount from the same payment was shown multiple times). 

In this implementation we pass in the invoices paymentables which are the most accurate representation of the total amount of payments applied to the invoice.

To support this, and backward compatibility with links to payments, we have introduced a payment_id prop on the paymentable transformer.

We have also extracted the components for reuse.


beganovich (approved):
Thanks 👍

PR #3182, opened 2026-06-01
Refactor react_settings to prevent multiple re renders.

turbo124:
This PR reduces unnecessary rerenders caused by changes to `react_settings`.

Previously, `react_settings` lived inside the Redux user state. Any update to preferences could churn the larger user/change tree and cause unrelated consumers to rerender. This PR moves committed `react_settings` into a dedicated Jotai atom, with more focused access patterns for consumers that only need specific settings.

## Persistence Model

This changes the persistence model for `react_settings`.

Updates are now written optimistically to the local atom and persisted via the dedicated preferences endpoint. On a successful HTTP response, the frontend treats the local atom as the source of truth and does not rehydrate it from the PUT response.

The assumption is that a successful response means the submitted settings were accepted by the API. Avoiding response-based rehydration prevents stale server response payloads from overwriting newer local changes that may have occurred while a save request was in flight.

Queued saves re-read the atom at send time, so later saves persist the latest committed local state rather than an older captured snapshot.

## Draft Handling

This PR also separates committed settings from draft settings.

Cancelable flows, such as preferences modals and User Details preference edits, keep changes in local or draft state until the user explicitly saves. Cancel/discard flows can therefore drop dirty edits without mutating the committed atom or accidentally persisting those edits through an unrelated save.

For User Details, draft commits patch only the dirty paths into the committed atom. If persistence fails, only those paths are rolled back, preserving any concurrent immediate settings changes such as sidebar collapse or dark mode.

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -100,8 +100,9 @@ export default defineConfig({
    * Build first with: npx vite build --mode testing
    * This ensures VITE_* vars are loaded from .env.testing */
   webServer: {
-    command: 'npx vite preview',
+    command: 'npx vite build --mode testing && npx vite preview',
     port: 4173,
     reuseExistingServer: true,
+    timeout: 180_000, // give the build ~3 min to settle
   },
 });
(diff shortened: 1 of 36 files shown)


beganovich (approved):
Thanks 👍

PR #3187, opened 2026-06-02
Add banner for quickbooks no deposits enabled

turbo124:
Adds a warning banner if deposits are disabled in QB and user attempts to add a deposit in IN.


beganovich (approved):
Thanks 👍

PR #3189, opened 2026-06-02
[Cleanup] Adding Entity Number To Schedule Template Name

Civolilah:
@beganovich @turbo124 The PR includes the implementation of adding an entity number to the schedule template name if it is not already specifically entered. Screenshots:

Screenshot 2026-06-02 at 20 55 52

Screenshot 2026-06-02 at 20 56 02

Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #3190, opened 2026-06-02
[Cleanup] Improving Behavior For Report Previews

Civolilah:
@beganovich @turbo124 The PR includes improving behavior for report previews when selecting new columns, removing existing ones, or applying filters to an already generated preview. Let me know your thoughts.


turbo124:
@Civolilah is the design here deliberately unsetting the preview if i remove a row from the report columns?

PR #3191, opened 2026-06-03
Fixes for performance in report previews + multi column sorting

turbo124:
This pr addresses performance issues when generating large preview reports and then attempting to sort/add/remove columns from the initial report columns.

Each draggable causes an update to the reactsettings atom which then causes a cascading impact on all of the previews table rows, mainly due to the heavy read on the colors prop in react settings.

- Now, local edits to sortablecolumn does not bubble up and convert into a write until the user clicks Preview, Export or Schedule.
- useColorScheme() no longer subscribes to the entire settings object, now we focus only to the dark_mode prop.
- Drop deep clone, and consider the data as immutable arrays

Also, included a small change to make multi sort the default option when sorting the previews.

### DIFF
--- a/src/common/colors.ts
+++ b/src/common/colors.ts
@@ -8,7 +8,7 @@
  * @license https://www.elastic.co/licensing/elastic-license
  */
 
-import { useReactSettings } from './hooks/useReactSettings';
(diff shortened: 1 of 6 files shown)


beganovich (approved):
Thanks 👍

PR #3192, opened 2026-06-03
Remove tags normalization for put/post

turbo124:
This PR removes the normalized of tags from PUT/POST requests and shift the normalization requirement to the API.

As tags will be introduced to all entities in the future. The front end should not be burdened with the tasks of normalizing the payload.

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -107,8 +107,9 @@ export default defineConfig({
   // },
 
   webServer: {
-    command: 'npx vite --mode testing --port 4173',
+    command: 'npx vite build --mode testing && npx vite preview --port 4173',
     port: 4173,
     reuseExistingServer: true,
+    timeout: 180_000,
   },
 });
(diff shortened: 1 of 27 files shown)


src/pages/clients/show/components/Details.tsx
```diff
@@ -53,6 +56,26 @@ export function Details(props: Props) {
     });
   };
 
+  const getTestingSettingText = (propertyKey: keyof Settings) => {
```
beganovich, line 57: Don't like this part being here. Sending fix.

playwright.config.ts
```diff
@@ -99,10 +99,16 @@ export default defineConfig({
...
+  // webServer: {
+  //   command: 'npx vite build --mode testing && npx vite preview',
+  //   port: 4173,
+  //   reuseExistingServer: true,
+  //   timeout: 180_000, // give the build ~3 min to settle
+  // },
+
   webServer: {
-    command: 'npx vite build --mode testing && npx vite preview',
+    command: 'npx vite --mode testing --port 4173',
```
beganovich: We're no longer testing against production build? Can you double check this is what you wanted?

beganovich (approved):
Looks good, I've pushed tiny fix for settings resolver. Please retest and merge if good 👍

turbo124:
thks

PR #3194, opened 2026-06-03
add Resume button to extend last timelog entry (fixes #3064)

### LINKED ISSUE #3064: Ability to resume / continue a subtask's timer to extend the existing timelog entry
Running the Enterprise SaaS version from https://app.invoicing.co in a desktop web browser, with task timelog descriptions enabled (essentially sub-tasks, as seen in the screenshot in #3063), there are sometimes (often) situations where I stopped the timer for a task, and then after a few minutes I got a call back the client and have to resume the task. In those situations, the short time break between stopping and starting is insignificant and the description is the same, so I would want to be able to actually resume the last subtask in the task.

Currently, the UI offers no way to resume the timer in a way that reuses and extends the last timelog entry within a task, instead of inserting a whole new timelog entry; and it doesn't offer a way to "merge" two timelog entries together post-facto either. I end up having to do it manually, by stopping the task timer, then remembering the end time, deleting the last entry, editing the previous-to-last entry's end time to match the new end time, and _then_ saving.

muhammadshaharyaraulakh:
## Summary

Fixes #3064

When a user stops a task timer and needs to resume it shortly after 
(e.g. a client calls back), the current UI forces a brand new timelog 
entry instead of continuing the last one. This PR adds a Resume button 
that reopens the last stopped entry by clearing its end time.

### New file: `src/pages/tasks/common/hooks/useResume.ts`
1- `useResume()` hook — clears the `end_time` of the last timelog entry 
  and calls `?start=true` to reopen it instead of creating a new entry
2-`canResumeLastEntry()` helper — returns `true` only when the last 
  entry was stopped within the past 60 minutes (avoids showing Resume 
  on old tasks)

### Modified: `src/pages/tasks/common/components/TaskDetails.tsx`
1-Added Resume button next to the Start button in both desktop and 
  mobile layouts
2- Resume button only appears when `canResumeLastEntry()` is true

## Behavior

1- Stop a task timer
2- Within 60 minutes, open the task edit page
3- A **Resume** button appears next to Start
4- Clicking Resume extends the last entry instead of creating a new one
5- After 60 minutes, only the regular Start button is shown

## Testing

1- Create a task and start the timer
2- Stop the timer
3- Go to edit the task — Resume button should appear
4- Click Resume — verify the last timelog entry's end time is cleared 
   (not a new row added)

### DIFF
--- a/src/pages/tasks/common/components/TaskDetails.tsx
+++ b/src/pages/tasks/common/components/TaskDetails.tsx
@@ -22,6 +22,7 @@ import { UserSelector } from '$app/components/users/UserSelector';
 import { TaskStatusSelector } from '$app/components/task-statuses/TaskStatusSelector';
 import { TaskStatus as TaskStatusBadge } from './TaskStatus';
 import { useStart } from '../hooks/useStart';
+import { useResume, canResumeLastEntry } from '../hooks/useResume';
 import { useStop } from '../hooks/useStop';
 import { isTaskRunning } from '../helpers/calculate-entity-state';
 import { formatTime, TaskClock } from '../../kanban/components/TaskClock';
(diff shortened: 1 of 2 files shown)


turbo124:
@muhammadshaharyaraulakh thanks for the PR, i'm not sure this is the right way forward for this particular feature.

image

We now have 3 ways to increase the timelog, and it isn't really clear the intent. personally, I would have thought this functionality would have placed a secondary button on the last time entry to "resume" that specific entry

image

It also requires changes to the api to ensure the last time log entry is adjusted as expected... as there is no functionality currently to support this.

@nekohayo what would your expectation be here?

@beganovich interested in your thoughts also.

PR #3196, opened 2026-06-03
[Cleanup] Reports Info Toaster For Large Reports

Civolilah:
@beganovich @turbo124 The PR includes the implementation of an info toaster as a notification when the preview of the report could not be displayed because the report is too large. Also, I've added a scroll to the start of the preview table when the response has been returned. Screenshot:

Screenshot 2026-06-03 at 23 20 09

`Note`: In the toaster, I've used the `report_too_large_to_preview` translation keyword. If you agree with it, please just add it to the files.

Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #3199, opened 2026-06-04
isHosted guard for tasks + fixes

turbo124:
Add isHosted() guard for calendar connections
Remove time totals from calendar view as they may not always be correct when a task spans over multiple days
Add missing translations


beganovich (approved):
Thanks 👍

PR #3202, opened 2026-06-05
Fixes for calendar connection interface

turbo124:
Change required to move calendar connection onto its own prop via user.settings, rather than overload referral_meta.
Small fix to ensure Purchase Order preview interceptor is passed in.
Change to ensure that browser time is used when presenting a task in the calendar rather than UTC.


beganovich (approved):
Thanks 👍

PR #3203, opened 2026-06-05
[Cleanup] Increasing Debounce For Data Table Field

Civolilah:
@beganovich @turbo124 The PR includes increasing the debounce from 300 to 500 ms for the filter data table field. Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #3206, opened 2026-06-07
Add dependency files for tiny mce

turbo124:
@beganovich this PR include the tinymce files as  I believe we bundle these with the application rather than generate them from the build?


beganovich (approved):
Thanks 👍

PR #3210, opened 2026-06-07
[Bugfix] Filtering Invoice Design Columns

Civolilah:
@beganovich @turbo124 The PR includes implementation of correct filtering of invoice design columns when they do not need to be displayed by settings configuration. Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #3211, opened 2026-06-08
Improve templates UX

beganovich:
https://github.com/DocuNinja/builder2.0/pull/144

 - Signatory labels are now clickable. Tap a signatory name to assign a person, no more hunting for a tiny swap icon.                                                                             
- Single-signatory templates skip the editor. Pick who to send to in a popup, and it saves + sends automatically.                                                   
- Send button now saves first then sends, one click instead of two, no extra confirmation dialog.                                                                                                       
- Buttons renamed. "Send" and "Save for later" make your intent clear.                                                      
- Send button works immediately after assigning signatories, no need to refresh or save first. 
- When removing signatories (placeholders or invited), we show confirmation dialog now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "28.06.2025.1",
       "dependencies": {
         "@azure/msal-browser": "^3.28.1",
-        "@docuninja/builder2.0": "^0.0.96",
+        "@docuninja/builder2.0": "^0.0.99",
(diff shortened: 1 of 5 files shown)


turbo124:
@beganovich this looks great. Only changes that may be worth considering is hiding the send button _until_ the signatories have been selecting, and then disabling after being sent the first time until a full page reload (to prevent double send?)

PR #3215, opened 2026-06-11
[Cleanup] Default Widths | Reports

Civolilah:
@beganovich @turbo124 The PR includes adjustments for the max widths of columns in the report preview table, along with adding the ability to scroll horizontally. Let me know your thoughts.

### DIFF
--- a/src/pages/reports/common/components/EnhancedPreview.tsx
+++ b/src/pages/reports/common/components/EnhancedPreview.tsx
@@ -25,7 +25,8 @@ import {
 import { isSummableColumn } from '../constants/columns';
 import { useNumericFormatter } from '$app/common/hooks/useNumericFormatter';
 // Import the preview types and hook from Preview.tsx
-import { usePreview } from './Preview';
+import { PreviewCell, usePreview } from './Preview';
+import { resolveCellWidth } from '../constants/column-widths';
 
 interface EnhancedPreviewProps {
   enableMultiSort?: boolean;
@@ -106,6 +107,19 @@ export function EnhancedPreview({
 
     return { columns: preview.columns, rows };
   }, [preview, activeFilters, sortConfigs]); // Dependencies ensure re-computation when any change
+
+  const columnWidths = useMemo(() => {
+    if (!preview) {
+      return new Map<string, ReturnType<typeof resolveCellWidth>>();
+    }
+
+    return new Map(
+      preview.columns.map((column) => [
+        column.identifier,
+        resolveCellWidth(column.identifier, column.display_value),
+      ])
+    );
+  }, [preview]);
 
   const columnTotals = useMemo(() => {
     const summable = new Set<string>();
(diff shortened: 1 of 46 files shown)


turbo124:
@Civolilah i'm not seeing a significant change here, can we add in resizeable columns for the preview table pls.

image

turbo124:
@Civolilah looks good. i can see WIP label still in place. i've reviewed and functionality looks correct, feel free to request review from @beganovich

Civolilah:
> @Civolilah looks good. i can see WIP label still in place. i've reviewed and functionality looks correct, feel free to request review from @beganovich

@turbo124 I just wanted to check code very well! Now available for @beganovich review also!

src/pages/reports/common/components/EnhancedPreview.tsx
```diff
@@ -46,9 +45,7 @@ export function EnhancedPreview({
 
   const [sortConfigs, setSortConfigs] = useState<SortConfig[]>([]);
   const [filterValues, setFilterValues] = useState<Record<string, string>>({});
-  const [sortTypeOverrides, setSortTypeOverrides] = useState<
-    Record<string, SortType>
-  >({});
+  const [sortTypeOverrides] = useState<Record<string, SortType>>({});
```
beganovich: I think this is dead code? Not used anywhere?

src/pages/reports/common/constants/column-widths.ts
```diff
@@ -0,0 +1,175 @@
...
+  '.notes',
+  '.item_notes',
+  '.description',
+  '.terms',
+  '.footer',
+  '.reminder_schedule',
+];
+
+const matchesSuffix = (identifier: string, suffixes: string[]) => {
+  return suffixes.some((suffix) => identifier.endsWith(suffix)) ?? false;
```
beganovich: No need for ?? false.

src/pages/reports/common/components/Preview.tsx
```diff
@@ -100,6 +118,274 @@ export function usePreview() {
...
+        if (!el) {
+          return;
+        }
+
+        if (typeof w === 'number' && w > 0) {
+          el.style.width = `${w}px`;
+          el.style.minWidth = `${w}px`;
+        } else {
+          el.style.width = '';
+          el.style.minWidth = '';
```
beganovich: This means intiial load won't have anything so it might result in UI flashing, right? 

Could this work?

```
if (typeof w === 'number' && w > 0) {
  el.style.width = `${w}px`;
  el.style.minWidth = `${w}px`;
} else {
  const defaultWidth = resolveMinColumnWidth(column.identifier);
  el.style.width = `${defaultWidth}px`;
  el.style.minWidth = `${defaultWidth}px`;
}
```

Civolilah: @beganovich Good catch! On initial load there was nothing setting widths so columns just distributed evenly, not great UX. Went with your suggestion using `resolveMinColumnWidth` as the default. Since it runs in `useLayoutEffect`, widths apply before paint so no flashing.

src/pages/reports/common/constants/column-widths.ts
Civolilah: @beganovich Yeah, removed!

src/pages/reports/common/components/EnhancedPreview.tsx
Civolilah: @beganovich Yes, this was part of the old logic, which is no longer in use.

beganovich (approved):
Thanks 👍

PR #3218, opened 2026-06-12
Improve dates across task views

turbo124:
Allows better tracking of dates when moving between day/week/month views.


beganovich (approved):
Thanks 👍

PR #3220, opened 2026-06-12
[Bugfix] Hiding Task Start Button

Civolilah:
@beganovich @turbo124 The PR includes the implementation of hiding the task start button when the last stop time log is greater than or equal to the current UTC time. Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #3221, opened 2026-06-13
[Bugfix] Fixing Formatting Date For Falsy Values

Civolilah:
@beganovich @turbo124 The PR includes fixes for date formatting of falsy values (when we do not have a populated property like `last_login_at`). Let me know your thoughts.

### DIFF
--- a/src/common/helpers.ts
+++ b/src/common/helpers.ts
@@ -69,7 +69,7 @@ export function classNames(...classes: any) {
 }
 
 export function date(date: number | string, format: string) {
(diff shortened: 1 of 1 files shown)


beganovich (commented):
!date should cover all these? 

0, '', undefined, null, NaN, false

Civolilah:
> !date should cover all these?
> 
> 0, '', undefined, null, NaN, false

@beganovich I agree! I just replaced it with `!date`, all of those are just falsy values.

beganovich (approved):
Thanks 👍

PR #3224, opened 2026-06-16
[Cleanup] Adjusting Passing Filters From Main Tables & Dashboard Tables Query Params

Civolilah:
@beganovich @turbo124 The PR includes logic adjustments so that search filters are not passed to sub-tables, only status filters are passed. It also includes fixes for dashboard tables so they do not take any preferences from the main tables, but instead just use the passed endpoint. Let me know your thoughts.

### DIFF
--- a/src/common/hooks/useDataTablePreferences.ts
+++ b/src/common/hooks/useDataTablePreferences.ts
@@ -40,7 +40,7 @@ interface Params {
   withoutStoringPerPage: boolean;
   enableSavingFilterPreference?: boolean;
   withoutStoringPage?: boolean;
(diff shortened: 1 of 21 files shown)


Civolilah:
@turbo124 Please just confirm that the query parameters on the dashboard tables are behaving well.

turbo124:
@Civolilah i didn't realise both issues were included here. deleted filters are still impacting Dashboard tables even on this PR. 

Dashboard tables should _NEVER_ have their query parameters altered

beganovich (approved):
Thanks 👍

PR #3226, opened 2026-06-16
[Cleanup] Adding Invoice Selector On Payment Page

Civolilah:
@beganovich @turbo124 The PR includes the implementation of an invoice selector on the payment creation page to enable the ability to create a payment by selecting an invoice. Screenshot:

Screenshot 2026-06-17 at 13 46 27

Let me know your thoughts.

### DIFF
--- a/src/components/invoices/InvoiceSelector.tsx
+++ b/src/components/invoices/InvoiceSelector.tsx
@@ -19,6 +19,7 @@ interface Props extends GenericSelectorProps<Invoice> {
   onClearButtonClick?: () => void;
   errorMessage?: string | string[];
   clientId?: string;
+  endpoint?: string;
 }
 
 export function InvoiceSelector(props: Props) {
@@ -37,11 +38,14 @@ export function InvoiceSelector(props: Props) {
       inputOptions={{
         value: props.value ?? null,
       }}
-      endpoint={endpoint(
-        `/api/v1/invoices?include=client&filter_deleted_clients=true&status=active${
-          props.clientId ? `&client_id=${props.clientId}` : ''
-        }`
-      )}
+      endpoint={
+        props.endpoint ??
+        endpoint(
+          `/api/v1/invoices?include=client&filter_deleted_clients=true&status=active${
+            props.clientId ? `&client_id=${props.clientId}` : ''
+          }`
+        )
+      }
       onChange={(invoice: Entry<Invoice>) =>
         invoice.resource && props.onChange?.(invoice.resource)
       }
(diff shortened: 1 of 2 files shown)


turbo124:
image


the invoices list needs to be filtered with payable.

currently, i am able to select an invoice that has already been paid (balance 0)

Civolilah:
> image
> the invoices list needs to be filtered with payable.
> 
> currently, i am able to select an invoice that has already been paid (balance 0)

@turbo124 Yeah, you're right. When we have a client selected, we have a payable query parameter to include in the query. But when we don't have one selected, we simply pull a list of invoices with a specific `client_status` in this case, `client_status=unpaid,partial` , which will give us only invoices that are not paid or are partially paid. I hope that works 100% correctly. Let me know your thoughts.

turbo124:
on the invoices route we have ?payable=true as a passable query param

Civolilah:
> on the invoices route we have ?payable=true as a passable query param

@turbo124 Ah, okay! I didn't know that about that one. So, I replaced `client_status` with `payable=true` when the client is not selected, but the problem is that I noticed the responses differ between endpoints.

When the query parameter is `payable=true`, it does not return any invoices, but there should be one returned:

`https://invoicing.co/api/v1/invoices?include=client&filter_deleted_clients=true&sort=date|desc&payable=true&status=active&filter=`

When the query parameter is set to `payable=[id]`, it returns that one invoice for specific client:

`https://invoicing.co/api/v1/invoices?include=client&filter_deleted_clients=true&sort=date|desc&payable=lej2kND4ej`

Can this be an API issue, or could something be wrong with the endpoint?

beganovich (approved):
Looks good, thanks!

Ping @Civolilah for merge conflicts.

beganovich (approved):
Looks good, one thing to clear up.

We have divergent state. Is this allowed?

1. We select invoice from dropdown
2. In data table we uncheck that invoice
3. Invoice selector will show that invoice

Please just confirm this is intended behaviour.

@turbo124 @Civolilah 

Image

Civolilah:
> Looks good, one thing to clear up.
> 
> We have divergent state. Is this allowed?
> 
> 1. We select invoice from dropdown
> 2. In data table we uncheck that invoice
> 3. Invoice selector will show that invoice
> 
> Please just confirm this is intended behaviour.
> 
> @turbo124 @Civolilah
> 
> Image

@beganovich I agree with you, this has been improved. So, the invoice selector is used for easier selection and easier client resolution. Once an invoice has been deselected from the table, it doesn't make sense to have it still selected in the invoice selector. So, I improved it. Let me know your thoughts.

turbo124:
good catch, and it appears that this particular issue has been resolved in the latest iteration @Civolilah @beganovich

beganovich (approved):
Thanks 👍

PR #3227, opened 2026-06-17
[Bugfix] Enter Payment Bulk Action

Civolilah:
@beganovich @turbo124 If you check the current production, you will notice that the pre-selection of invoices when performing the "Enter Payment" bulk action does not work. This PR fixes that and makes it work the same as the single "Enter Payment" action. Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #3228, opened 2026-06-17
[Feature] Implementing Last Error Message For Health Check Modal

Civolilah:
@beganovich @turbo124 The PR includes the implementation of a "last error message" label on the health check modal. Screenshot:

Screenshot 2026-06-17 at 18 06 58

Let me know your thoughts.

### DIFF
--- a/src/components/AboutModal.tsx
+++ b/src/components/AboutModal.tsx
@@ -109,6 +109,7 @@ export function AboutModal(props: Props) {
   const [systemInfo, setSystemInfo] = useState<SystemInfo | undefined>(
     currentSystemInfo
   );
+  const [lastError, setLastError] = useState<string>('');
 
   const handleHealthCheck = (allowAction?: boolean) => {
     if (!isFormBusy || allowAction) {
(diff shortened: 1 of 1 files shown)


Civolilah:
@turbo124 I wasn't able to test it 100% because of a 403 error on the staging API. So please just confirm that it works well on a real self-hosted environment.

src/components/AboutModal.tsx
```diff
@@ -327,26 +334,41 @@ export function AboutModal(props: Props) {
             <div className="flex flex-col">
               <span className="font-medium text-base mb-1">PHP</span>
               <span>
-                {t('web')}: {systemInfo?.php_version.current_php_version}
+                {t('web')}: {systemInfo?.php_version?.current_php_version}
               </span>
               <span>
-                {t('cli')}: {systemInfo?.php_version.current_php_cli_version}
+                {t('cli')}: {systemInfo?.php_version?.current_php_cli_version}
```
Civolilah, line 340: @beganovich The staging API does not have those responses, so to avoid experiencing errors locally while I'm developing, I always need to make changes. Right now, I've made it so that it will be avoided locally in the future.

turbo124:
image


I think the problem is the prop accessor, i've adjusted this.

Civolilah:
> image
> I think the problem is the prop accessor, i've adjusted this.

@turbo124 Yep, I assumed it would be the `message` prop, now it should be correct! Thanks!

beganovich (approved):
Thanks 👍

PR #3242, opened 2026-06-30
Tags settings page refactor + supporting DataTable preferences changes

turbo124:
Includes:
- Tags settings list/create/edit pages and tag entity-type interface
- entity_types custom-filter support in DataTable + useDataTableUtilities
- DataTable preferences persistence migrated to the reactSettings atom (renames withoutStoringStatusPreferences -> withoutStoringSearchFilter; updates the 12 client/vendor show-page consumers)
- Fix: tables passing defaultCustomFilterValues never enabled their query because the preferences-applied gate kept checking the now-seeded customFilter; rely on the appliedRef run-once guard instead
- en.json: add "global", correct "bank_transaction" label
- tags e2e tests


beganovich (approved):
Thanks 👍

PR #3247, opened 2026-07-02
[Cleanup] Implementing Pagination Using API Responses

Civolilah:
@beganovich @turbo124 The PR includes implementation of new logic for changing pages on tables. So, instead of handling it in React, we will take the next or previous property values from the pagination response object. Also, the PR prepares logic for more changes like this one. Let me know your thoughts.

### DIFF
--- a/src/components/tables/Pagination.tsx
+++ b/src/components/tables/Pagination.tsx
@@ -26,15 +26,18 @@ import {
 } from '$app/common/helpers/pagination';
 import { PaginationMeta } from '$app/common/interfaces/generic-many-response';
 
-const PaginationButton = styled.div<{ $disabled?: boolean }>`
+const PaginationButton = styled.button`
   background-color: ${(props) => props.theme.backgroundColor};
   border-color: ${(props) => props.theme.borderColor};
-  opacity: ${(props) => (props.$disabled ? 0.5 : 1)};
-  cursor: ${(props) => (props.$disabled ? 'not-allowed' : 'pointer')};
+  cursor: pointer;
 
-  &:hover {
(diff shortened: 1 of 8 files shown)


turbo124:
@Civolilah looks like a lot of this is dead code and not used? we should also have matching playwright tests that confirm functionality is working as expected.

Civolilah:
> @Civolilah looks like a lot of this is dead code and not used? we should also have matching playwright tests that confirm functionality is working as expected.

@turbo124 Yeah, I know there is dead code, but I actually wanted to prepare some stuff for potential future additions. However, I removed the dead code and left only what we need for now. I also added tests for the functionality that is successfully passing. Let me know your thoughts.

src/components/tables/Pagination.tsx
```diff
@@ -19,13 +19,22 @@ import { DoubleChevronLeft } from '../icons/DoubleChevronLeft';
...
-
-const PaginationButton = styled.div`
+import {
+  getPageNavigationState,
+  PageNavigationAction,
+  resolvePageNavigation,
+} from '$app/common/helpers/pagination';
+import { PaginationMeta } from '$app/common/interfaces/generic-many-response';
+
+const PaginationButton = styled.div<{ $disabled?: boolean }>`
```
beganovich: I wonder if this can be just styled.button? 

`divs` aren't good with a11y, which means no proper focus state, tab activation and other stuff.

Civolilah: @beganovich Yeah, I agree, it has been switched to button. Let me know your thoughts.

beganovich (approved):

PR #3248, opened 2026-07-03
[Cleanup] Implementing Flag/Preference For Storing Table Filters

Civolilah:
@beganovich @turbo124 The PR includes implementation of a flag for storing table filter preferences under the Preferences page. Let me know your thoughts.


turbo124:
@Civolilah this isn't working for me. the filters are now not passing at all regardless of the settings toggle.

Civolilah:
> @Civolilah this isn't working for me. the filters are now not passing at all regardless of the settings toggle.

@turbo124 As we discussed on Slack, in addition to implementing the flag functionality, I have fixed what was broken on the develop branch. So now everything should behave correctly. Let me know your thoughts.

beganovich (commented):
I think naming is a bit confusing here.

`withoutStoringStatusPreferences` blocks all persistance not just status? 

The old `withoutStoringSearchFilter` only prevented `storeSessionTableFilters()` from running. The rest of the save function still executed. 

The new `withoutStoringStatusPreferences` returns early from the entire save handler, and on load it resets custom filters / sort / perPage / status to defaults while only restoring the search‑filter text.

I don't mind this change if it was intentional, but naming is problematic then.

Please confirm @turbo124 @Civolilah

turbo124:
The intent here is to ensure that the filter input values are passed down to the client overview page _only_

The parent would also retain ALL filters and Status filters, so the parent state should always be preserved, but the client overview is ONLY passed the filter input, and nothing else.

beganovich:
Couple things then:

1. Naming mismatch: withoutStoringStatusPreferences doesn't match the intent. The prop controls far more than just status. Something like withFilterTextOnly or withoutInheritingPreferences would be clearer.

2. Global toggle breaks the feature: When persist_table_filters is false, withoutStoringPreferencesEffective becomes true, which skips all preference restoration including the filter text. So client overview pages won't receive the search input from the parent, contradicting the stated intent.

cc @Civolilah

Civolilah:
> Couple things then:
> 
> 1. Naming mismatch: withoutStoringStatusPreferences doesn't match the intent. The prop controls far more than just status. Something like withFilterTextOnly or withoutInheritingPreferences would be clearer.
> 2. Global toggle breaks the feature: When persist_table_filters is false, withoutStoringPreferencesEffective becomes true, which skips all preference restoration including the filter text. So client overview pages won't receive the search input from the parent, contradicting the stated intent.
> 
> cc @Civolilah

@beganovich Okay, I hope I understood you correctly. I improved the solution by renaming properties and adding a non-impacting `persist_table_filters` property to the text input filter. Let me know your thoughts.

beganovich (approved):

PR #3249, opened 2026-07-05
Add generic OIDC sign-in button + callback consumer

matheusliraofficial:
@beganovich @turbo124 Companion frontend PR for the backend generic OIDC
provider in invoiceninja/invoiceninja#12073 (which closes
invoiceninja/invoiceninja#10839).

**What's in this PR**

- **`components/OidcSignIn.tsx`** — new button rendered on the login page.
  Reads `GET /api/v1/oidc/config` on mount. Renders a "Sign in with {label}"
  button (matching the existing SignInProviders styling) only when the backend
  responds `oidc_enabled: true`. Clicking it does a full-page navigation to
  `${apiEndpoint()}/auth/oidc` — the backend takes over from there.
- **`OidcCallback.tsx`** — the SPA lands here after the backend Socialite
  round-trip. Reads `?code=` (a 64-char one-shot exchange code with a 60s TTL
  — the real CompanyToken never touches the URL, so it can't leak into
  browser history, Referer headers, or upstream access logs). POSTs the code
  to `POST /api/v1/oidc/exchange` to receive the actual CompanyToken, then
  swaps that for a full `CompanyUser[]` payload via `POST /api/v1/refresh_react`,
  dispatches `authenticate({type: TOKEN, ...})`, and navigates to `/dashboard`.
  Short-circuits on `?error=` from the IdP/backend and surfaces the message
  inline with a "Back to login" affordance. Same terminal state as the current
  `oauthApiLogin` path used by Google/Microsoft.
- **`routes.tsx`** — registers `/oauth-callback` under `PublicRoute`.
- **`Login.tsx`** — imports and renders `<OidcSignIn />` below `SignInProviders`.

**Backward compatibility**

Zero UI change when the backend doesn't have OIDC configured — the config
endpoint returns `oidc_enabled: false` and the button component early-returns
`null`. Google/Microsoft SignInProviders are untouched.

**Manual test**

Against a self-hosted Authentik: log out, click "Sign in with Authentik",
authenticate at the IdP, land on `/dashboard`.

No screenshots yet — happy to attach if you want to see the login page with
the extra button.


matheusliraofficial:
@turbo124 , can now we move with this one?

turbo124:
@matheusliraofficial yes, I plan to reviewing this one this week.

src/pages/authentication/components/OidcSignIn.tsx
```diff
@@ -0,0 +1,87 @@
...
+      });
+
+    return () => {
+      cancelled = true;
+    };
+  }, []);
+
+  if (!config?.oidc_enabled) {
+    return null;
+  }
```
turbo124: This returns an error for any installation that doesn't have the matching API shipped.

You would need to handle this gracefully using the skipIntercept and handling a 404 on that route.

Because this call is nested in a  useEffect() there are also issues with multiple calls to this endpoint on unrelated re-renders.

src/pages/authentication/OidcCallback.tsx
```diff
@@ -0,0 +1,99 @@
...
+      setError(oauthError);
+      return;
+    }
+
+    if (!code) {
+      setError('Missing OIDC exchange code in callback URL.');
+      return;
+    }
+
+    request('POST', endpoint('/api/v1/oidc/exchange'), { code })
```
turbo124: You need to skipIntecept here if you want to model the requests like this so that you can handle the failures (401/429/404) otherwise you end up in a clear localstorage / reload loop endlessly.

turbo124 (changes requested):
Please see comments:

Please use translation strings rather than hardcoded strings

@beganovich may be able to provide some additional implementation feedback also.

src/pages/authentication/components/OidcSignIn.tsx
matheusliraofficial: Fixed in 458b9ffee8e6a680962e8a023148d84d8dc3e64a.

- Added `skipIntercept: true` so the missing-endpoint 404 on older backends no longer trips the global interceptor (toast / clear-localstorage / reload).
- Swapped the `useEffect + useState` for `useQuery({ staleTime: Infinity, retry: false })` so parent re-renders/remounts share a single in-flight request instead of hitting the endpoint again.
- Wrapped the "Sign in with {label}" string in `t('sign_in_with', { provider: label })` — new key added to `en.json`.

Fires exactly once per session now, and silently hides the button on backends that don't ship the endpoint yet.

src/pages/authentication/OidcCallback.tsx
matheusliraofficial: Fixed in 458b9ffee8e6a680962e8a023148d84d8dc3e64a.

Added `skipIntercept: true` to **both** hops of the exchange:

- `POST /api/v1/oidc/exchange`
- `POST /api/v1/refresh_react`

Without it, a 401/429/404 from either request would hit the global interceptor and trigger `clearLocalStorage() + window.location.reload()`, which lands right back on `/oidc/callback` with the same failing code and loops forever. Errors now propagate to the local `.catch` and render a clean "Sign-in failed" screen with the backend's error message.

Also moved all hardcoded strings (`oidc_missing_code`, `oidc_no_token`, `oidc_login_failed`, `sign_in_failed`, `back_to_login`, `signing_you_in`) to `en.json` per the review body.

turbo124:
@beganovich did you have any additions here?

beganovich:
Looks good. Only thing I'd suggest is to guard/hide OIDC button with `isSelfHosted` since it makes no sense on hosted platform, AFAIK.

matheusliraofficial:
Thanks @beganovich — good catch. Fixed in e43ce92fb.

Guarded both the button render and the `GET /api/v1/oidc/config` discovery request on `isSelfHosted()` (via `enabled: isSelfHosted()` on the `useQuery`), so nothing fires against the hosted API and the button is fully hidden on hosted.

Also merged the latest `develop` in daa6e6d1f to clear the conflict (both sides only added new i18n keys; taking `develop`'s `@docuninja/builder2.0` bump to 0.0.109 since we don't touch deps). Full typecheck is clean.

turbo124:
@beganovich please review and merge if ready :+1:

src/pages/authentication/components/OidcSignIn.tsx
```diff
@@ -0,0 +1,84 @@
...
+            viewBox="0 0 24 24"
+            fill="none"
+            stroke="currentColor"
+            strokeWidth="2"
+            strokeLinecap="round"
+            strokeLinejoin="round"
+          >
+            <path d="M12 2l3 6 6 .9-4.5 4.3 1 6.3L12 16.9 6.5 19.5l1-6.3L3 8.9 9 8z" />
+          </svg>
+          <p>{t('sign_in_with', { provider: label })}</p>
```
Ninos, line 79: May just use provider label without `sign_in_with` prefix? OIDC and SSO are too technical for "Consumer" -> so I prefer `Single Sign-On`. With the hardcoded prefix the button gets unnessessary long (`Sign in with Single Sign-On`)

Ninos (commented):
Additionally would be nice to get some `OIDC_FORCE`/`DISABLE_LOGIN_FORM` for auto-redirecting to oidc provider/block native login/hide native login form. I can create a MR after this one 😇

src/pages/authentication/components/OidcSignIn.tsx
beganovich: cc @turbo124 on this one.

beganovich (approved):

PR #3253, opened 2026-07-07
[Feature] Documents Tab On Sliders

Civolilah:
@beganovich @turbo124 The PR includes the implementation of a Documents tab on existing sliders. For entities that do not have sliders, we have implemented it with Overview and Documents tabs. Screenshot:

Screenshot 2026-07-07 at 22 25 45

Let me know your thoughts.


src/pages/expenses/index/Expenses.tsx
```diff
@@ -46,6 +55,28 @@ export default function Expenses() {
...
+
+  useEffect(() => {
+    if (expenseResponse && expenseSliderVisibility) {
+      setExpenseSlider(expenseResponse);
+    }
+  }, [expenseResponse, expenseSliderVisibility]);
+
+  useEffect(() => {
+    return () => setExpenseSliderVisibility(false);
+  }, []);
```
beganovich, line 83: This will cause flashing data when switching between different entities.

Please make sure to clean up data on close, if possible.

```
useEffect(() => {
  setExpenseSlider(null);
}, [sliderExpenseId]);
```

Or even better have "loading" state with isPreviousData/isFetching.

beganovich (approved):
Looks good, just to fix that loading state if possible.

src/pages/expenses/index/Expenses.tsx
Civolilah: @beganovich Yeah, I totally understand the issue, and I think I came up with a good enough solution to cover it. Let me know your thoughts.

beganovich (approved):

Issue #3254, opened 2026-07-07
Localize Address Formats


ludgerey:
Version: 16.06.2026.1

## Summary

Invoice Ninja currently renders addresses using a single layout regardless of the recipient's country. However, postal address formats differ between countries, including the order of address components, the placement of the postal code, and the punctuation used.

As a result, addresses for many countries are rendered incorrectly.

## Current Behavior

The current implementation appears to assume a US/Australian-style address format. This leads to issues such as:

Incorrect placement of the postal code (for most European countries the postal code should come before the city).
Unwanted commas in countries where they are not used.
Extra/wrong whitespace when optional address components are missing.

(See attached screenshots)

## Proposed Solution

Rather than maintaining country-specific formatting rules manually, I would recommend using an existing address formatting dataset.

Established sources are Google's Address Data Service (see [libaddressinput](https://github.com/google/libaddressinput)) or Unicode [CLDR](https://cldr.unicode.org/). These projects provide localized address formatting rules for virtually every country.

There are libraries implementing these datasets for most common programming languages.

For the React frontend, [localized-address-format](https://github.com/DASPRiD/localized-address-format) looks like a good fit.

Using a standardized formatter would ensure that addresses are rendered correctly for all supported countries while avoiding the need to maintain formatting templates within Invoice Ninja.

## Screenshots

<img width="393" height="207" alt="Image" src="https://github.com/user-attachments/assets/28979f6d-d046-4ec9-b233-c5251358c72b" />
<img width="408" height="263" alt="Image" src="https://github.com/user-attachments/assets/6f5a9655-630f-47fa-ad57-59b8b14cd795" />

turbo124:
@Civolilah we should use the company.settings.pdf_variables.client_details as the standard for displaying the client addresses in the UI.

ludgerey:
@turbo124 This won't solve the postal code misplacement, which I guess is not addressable with that configuration. Organizations can also have customers all around the world, so I think it's better to format the address according to the recipient's country rather than the organization's configured layout.

turbo124:
But from the UI you wouldn't ever need a customer address displayed differently? Only in the format the admin viewing the address would expect?

ludgerey:
@turbo124 I don't think so. I'd expect an address to be rendered in the format of the recipient's country, regardless of where it's displayed. Especially on the invoice itself, I'd want the address to match the format used by the destination country's postal system.

[...]

PR #3261, opened 2026-07-10
Localize Address Formats (#3254)

ludgerey:
This PR proposes a solution for https://github.com/invoiceninja/ui/issues/3254. Rather than using a fixed layout, it renders postal addresses according to the format used in the address's country. The formatting is based on standardized country-specific data provided by an address formatting library.

## Summary

* Added a shared `FormattedAddress` component to render localized address lines.
* Updated client address cards, client locations, and vendor address display to use the shared formatter.
* Added the [localized-address-format](https://github.com/DASPRiD/localized-address-format) dependency.
* Preserved existing card/layout `classNames` at call sites while centralizing address formatting and country resolution.
* Use address format for client location cards, rather than individual field labels.

## Before
Bildschirmfoto 2026-07-09 um 17 41 19
Bildschirmfoto 2026-07-09 um 17 41 45
Bildschirmfoto 2026-07-09 um 17 41 54
Bildschirmfoto 2026-07-10 um 09 21 30
Bildschirmfoto 2026-07-10 um 09 35 07

## After
Bildschirmfoto 2026-07-10 um 09 30 42
Bildschirmfoto 2026-07-10 um 09 30 27
Bildschirmfoto 2026-07-10 um 09 30 34
Bildschirmfoto 2026-07-10 um 09 29 59
Bildschirmfoto 2026-07-10 um 09 30 57


src/components/FormattedAddress.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { useResolveCountry } from '$app/common/hooks/useResolveCountry';
+
+interface Props {
+  address: {
+    address1?: string;
+    address2?: string;
+    city?: string;
+    state?: string;
+    postal_code?: string;
+    country_id: string;
```
beganovich: This interface should be extracted.

```tsx
// src/common/interfaces/address.ts
export interface AddressFields {
  address1?: string | null;
  address2?: string | null;
  city?: string | null;
  state?: string | null;
  postal_code?: string | null;
  country_id?: string | null;
}

// src/components/FormattedAddress.tsx
interface Props {
  address: AddressFields;
  className?: string;
  style?: CSSProperties;
}

File: src/components/FormattedAddress.tsx [hunk 2]
```diff
@@ -0,0 +1,45 @@
...
+  };
+  className?: string;
+  style?: CSSProperties;
+}
+
+export function FormattedAddress(props: Props) {
+  const resolveCountry = useResolveCountry();
+  const country = resolveCountry(props.address.country_id);
+
+  const addressLines = [
```
beganovich, line 16: Please wrap this into `useMemo`, it reads from cache, so no need to waste work.

beganovich (changes requested):
Thanks for PR! Looks good, just couple tiny changes 👍

src/components/FormattedAddress.tsx
ludgerey: I can wrap it in `useMemo`, but it will effectively only memorize a `.find() ` on the country list. You think it's still worth it?

beganovich: Yeah, I think so, didn't do any benchmarks, but country list itself is quite big so in theory it should help :+1:

ludgerey: I added `useMemo`.

ludgerey:
@beganovich Please review again :)

beganovich (approved):

PR #3278, opened 2026-07-24
[Cleanup] Implementing Burn Up Chart For Projects

Civolilah:
@beganovich @turbo124 The PR includes the implementation of a burn-up chart for projects, and it is placed on the show page. Screenshots:

Screenshot 2026-07-27 at 02 13 00

Screenshot 2026-07-27 at 02 15 31

Let me know your thoughts.


turbo124:
@Civolilah did you follow #3243 when creating this? I see the burnup, but there is a lot missing here.

Civolilah:
> @Civolilah did you follow #3243 when creating this? I see the burnup, but there is a lot missing here.

@turbo124 Ah, sorry! I missed it somehow. I've now reimplemented the chart using the logic from the mentioned PR. Screenshot with mocked data (endpoint is not available yet):

Screenshot 2026-07-28 at 13 57 42

Let me know your thoughts.

turbo124:
image

padding issues.


image


also, i notice that the project navigation does not show for the first project in the list, but when you view the second project, it then appears.

src/common/queries/project-burnup.ts
```diff
@@ -0,0 +1,45 @@
...
+  enabled?: boolean;
+}
+
+export function useProjectBurnupQuery(
+  payload: ProjectBurnupRequest,
+  options?: QueryOptions
+) {
+  return useQuery<ProjectBurnupResponse>({
+    queryKey: [
+      '/api/v1/charts/project_burnup/:id',
```
beganovich: useRefetch invalidates `/api/v1/charts/project_burnup`.. Does this mean we're effectively always having stale data (cached forever)?

src/common/queries/project-analytics.ts
```diff
@@ -0,0 +1,46 @@
...
+  enabled?: boolean;
+}
+
+export function useProjectAnalyticsQuery(
+  payload: ProjectAnalyticsRequest,
+  options?: QueryOptions
+) {
+  return useQuery<ProjectAnalyticsResponse>({
+    queryKey: [
+      '/api/v1/charts/project_analytics/:id',
```
beganovich: I might be wrong but we never invalidate `project_analytics`?

beganovich: Also, AFAIK we never did :id, but project_analytics without id reference, so we can invalidate whole resource?

Civolilah: @beganovich Hmm, yeah you are right, great catch, :id is unnecessary and makes caching stale, that's a mistake there. So, I think the best way here is just to remove :id and keep it as is, because all necessary entities have this endpoint in their dependencies, so when they change, the project_analytics endpoint also invalidates. I fixed the invalidation. Let me know your thoughts.

PR #3290, opened 2026-08-03
Fix credit note action for peppol

turbo124:
Fixes the link for Credit Note creation from the Invoice Actions menu - was formally pointing to the reverse option"


beganovich (commented):
Looks good. One thing caught my eye.. 

`e_invoice` from the source invoice leaks into cloned credits now. Previously we were setting e_invoice to null, which we don't do anymore? 

Unless that's a problem, this is good to merge.

@turbo124

turbo124:
@beganovich thanks! fixed and updated.

beganovich (approved):

PR #3311, opened 2026-08-10
[Cleanup] Implementing Cache Behavior Across Tabs

Civolilah:
@beganovich @turbo124 The PR includes an implementation of persisting the cache for certain cases. So, opening the app from scratch will pull all queries just as before, and reloading will do the same. However, opening new tabs, which is the case we want to handle differently, will take data from the cache. Let me know your thoughts.


turbo124:
@Civolilah thanks for this. Looks good. 

I wrote some tests evaluating this concept, and I was under the impression that the query cache would be shared across tabs, but this was wrong. Only a snapshot of the current cache state is copied over. So there are huge issues as sync/state mutations would never carry  across. So I think there is just too much risk providing this to end users.

@beganovich do you agree on this? Or are there other options available?

Civolilah:
> @Civolilah thanks for this. Looks good.
> 
> I wrote some tests evaluating this concept, and I was under the impression that the query cache would be shared across tabs, but this was wrong. Only a snapshot of the current cache state is copied over. So there are huge issues as sync/state mutations would never carry across. So I think there is just too much risk providing this to end users.
> 
> @beganovich do you agree on this? Or are there other options available?

@turbo124 I share the same thoughts, it's too risky, and I'm not sure that's actually common practice. I have one idea for cases like this, there are many solutions for tracking user behavior. So instead of implementing a solution first, we could set everything up and track how often something actually triggers. If it's rare, we just don't do it. Let me know your thoughts.

beganovich:
Not fan of this to be honest.

Unless we have really good reason for this, I'd just skip. Not worth the potential headaches.

turbo124:
agreed, lets close for now.

PR #3334, opened 2026-08-16
Compat shims for QR + Apple

turbo124:


beganovich:
For any future packages, just export via helper.

```tsx
import QRCodeImport from 'react-qr-code';
import { unwrapDefault } from '$app/common/helpers/unwrap-default';

export const QRCode = unwrapDefault(QRCodeImport);
```

PR #3341, opened 2026-08-18
[Cleanup] Implementing Requering Details Modal

Civolilah:
@beganovich @turbo124 The PR includes the implementation of a modal for first and last name when they are not entered, only for the HOSTED platform. It pops up every time the user comes to the app and the details are not entered, or if they try to purchase a plan before attempting, the modal will appear. Screenshot:

Screenshot 2026-08-28 at 21 06 49

Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #3345, opened 2026-08-19
Enable single signatory flow

beganovich:
https://github.com/DocuNinja/builder2.0/pull/159

image


turbo124:
@beganovich merge when happy!

PR #3346, opened 2026-08-20
Notifications - refactor for localStorage quota exceeded

turbo124:
If a significant number of websocket events land in a short time period, the localstorage can quickly exceeds it quota.

The changes in this PR address this by introducing:

1. Jotai only storage with no persistence layer
2. Dedupe functionality to prevent duplicate messages appearing
3. Capping notififcations to 100
4. Ensuring we dump all notifications on logout / company switching


beganovich (approved):
Thanks 👍

PR #3348, opened 2026-08-24
Set design on selection.

turbo124:
Enables live_design to update appropriately when a selector (other than invoice) is selected.

Coverage the custom template options also.

### DIFF
--- a/src/pages/invoices/common/components/InvoiceViewer.tsx
+++ b/src/pages/invoices/common/components/InvoiceViewer.tsx
@@ -136,7 +136,9 @@ export function InvoiceViewer(props: Props) {
                 props.onLink && props.onLink(url);
               }
 
-              toast.dismiss();
+              if (props.withToast) {
+                toast.dismiss();
+              }
 
               return response;
             })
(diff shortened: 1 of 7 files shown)


src/pages/invoices/common/components/InvoiceViewer.tsx
```diff
@@ -42,17 +45,68 @@ export function InvoiceViewer(props: Props) {
...
+  const lastFetchedKeyRef = useRef<string | null>(null);
+  const inFlightAbortRef = useRef<AbortController | null>(null);
+  const activeResourceKeyRef = useRef('');
+
+  const resourceKey = useMemo(() => {
+    if (props.resourceKey) {
+      return props.resourceKey;
+    }
+
+    return JSON.stringify(props.resource);
```
beganovich: This will break any invoice viewer that uses GET request (which is most of them). To reproduce this, just open invoice PDF and download_pdf never triggers.

src/pages/invoices/common/components/InvoiceViewer.tsx
```diff
@@ -61,10 +115,19 @@ export function InvoiceViewer(props: Props) {
 
           return request(props.method, props.link, props.resource, {
             responseType: 'arraybuffer',
-            signal,
+            signal: signal ?? abortController.signal,
```
beganovich: There's no need for this custom abort controller, React Query ALWAYS provides one.

beganovich (changes requested):
Sending fixes for mentioned issues.

PR #3353, opened 2026-08-28
[Cleanup] Stripe Disconnection Implementation

Civolilah:
@beganovich @turbo124 The PR includes implementation of the disconnect Stripe functionality if the account is connected. Screenshot:

Screenshot 2026-08-31 at 07 15 15

Let me know your thoughts.


Civolilah:
@turbo124 The "disconnect_stripe_confirmation" translation keyword is used on the confirmation modal by passing accountId into it. Please let me know if this is okay.

turbo124:
@Civolilah looks good. Can you confirm that the disconnect button only appears if length > 0 for account_id ?

beganovich (approved):

beganovich:
Disconnect is not account_id.length > 0. It is isHosted() && Boolean(accountId).

@turbo124

Civolilah:
> Disconnect is not account_id.length > 0. It is isHosted() && Boolean(accountId).
> 
> @turbo124

@beganovich @turbo124 Just checked Jira again, and it seems the condition is correct, so it's for the hosted platform only, with a condition to check if the account ID is populated. Let me know your thoughts.

turbo124:
@Civolilah we unset accountId after disconnect, so we would want to ensure that the conditional to display the disconnect button is length > 0

Civolilah:
> @Civolilah we unset accountId after disconnect, so we would want to ensure that the conditional to display the disconnect button is length > 0

@beganovich @turbo124 Okay, instead of a truthy value check, I replaced it with a length > 0 check, and it should work correctly again. Let me know your thoughts.

beganovich (approved):
Thanks 👍

PR #3355, opened 2026-08-31
[Cleanup] Default Tab Preference

Civolilah:
@beganovich @turbo124 The PR includes implementation of the default tab preference. Based on that, the invoice, recurring invoice, and quote routes will be adjusted with a URL parameter that determines which 'products' or 'tasks' tab should be the default. Screenshot:

Screenshot 2026-08-31 at 19 04 30


beganovich (commented):
I think core idea of driving this by URL is complex because we're going to run in these cases.

Here's demo of two spots I randomly found not to be working:

https://github.com/user-attachments/assets/25038c03-7d6e-435f-be96-f9f3fa28edf3

Perhaps, we can drive default tab using user preference instead of link/URL?

Civolilah:
> I think core idea of driving this by URL is complex because we're going to run in these cases.
> 
> Here's demo of two spots I randomly found not to be working:
> 
>  Screencast.From.2026-09-02.09-30-05.mp4 
> Perhaps, we can drive default tab using user preference instead of link/URL?

@beganovich I agree with you. I've created a more global solution with settings for the default tab index based on the configuration and, in some cases, the URL. Let me know your thoughts.

beganovich:
Thanks, but looks like you missed couple spots. Please see following two demos: 


https://github.com/user-attachments/assets/f9e6b617-9894-4026-90ba-0e8dccbb00cf


https://github.com/user-attachments/assets/85b0cb1f-5242-4c01-972b-9d57206b87b0

Civolilah:
> Thanks, but looks like you missed couple spots. Please see following two demos:
> 
>  Screencast.From.2026-09-03.14-34-49.mp4 
>  Screencast.From.2026-09-03.14-35-25.mp4

@beganovich Completely agree with you, when it is logically appropriate per navigation to NOT listen for preference, we should do that. I've adjusted the logic, and it should work great across all cases. Let me know your thoughts.

beganovich:
Looks better, but this still will miss couple places, for e.g.

1. Set default tab to products
2. Create invoice with at least one task line item
3. Clone to invoice, and you'll land on the products table.

I think we can amplify this effect by adding dominant entries rule. 

Order should be:

1. Explicit ?table
2. Dominant line item type
3. default_tab preference

Let me know if this works, @turbo124.

Civolilah:
> Looks better, but this still will miss couple places, for e.g.
> 
> 1. Set default tab to products
> 2. Create invoice with at least one task line item
> 3. Clone to invoice, and you'll land on the products table.
> 
> I think we can amplify this effect by adding dominant entries rule.
> 
> Order should be:
> 
> 1. Explicit ?table
> 2. Dominant line item type
> 3. default_tab preference
> 
> Let me know if this works, @turbo124.

@beganovich I agree with your recommendation, Ben. I think forcing a default tab will create some edge cases that we might forget to cover, and the default tab won't always be the most relevant. As you mentioned, sometimes a user will create an invoice with entirely task-based items, and it would be relevant to set that tab as the default. Let us know your thoughts @turbo124

Civolilah:
@turbo124 @beganovich I adjusted the implementation here per Ben's recommendation for prioritizing which tab needs to be selected upon landing on the page. So, if an explicit table parameter is set, that tab will be selected by default. The next level is the dominant tab, which means that if you land on the edit page of an invoice that has mostly task line items, we will select that tab regardless of whether the default tab is products in preferences. Finally, if there is no parameter or dominant number of any line items, the default tab from preferences will be taken as the value. Let me know your thoughts.

beganovich (approved):

PR #3356, opened 2026-09-01
[Cleanup] New Calculation Fields For Dashboard

Civolilah:
@beganovich @turbo124 The PR includes implementation of the new calculated fields for task metrics. Six fields were added to the dashboard card selector, with the calculation and format controls constrained per field, so invalid combinations can't be constructed, the count fields don't send the `format` property at all, and duration responses are rendered through a duration formatter instead of raw seconds. Since older API versions don't support these fields, they are feature-detected with a single probe request and hidden if it returns 422, without affecting any of the existing options. Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #3357, opened 2026-09-01
[Cleanup] Implementing Check Mailer

Civolilah:
@beganovich @turbo124 The PR includes the implementation of a mailer check on the email settings page.

@turbo124 `Important note:` I could not test it 100% because the route is returning a 404 for me. So, apologies if something is missed, and please let me know if anything is missing. Screenshot:

Screenshot 2026-09-01 at 22 17 14

Let me know your thoughts.


turbo124:
@Civolilah 

We'll want to present any error messages back in context. I can see that the 422's are displayed correctly, however on 400 (failure) the response shape is:

{"message":"Could not send a test email using the selected mailer."}

Civolilah:
> @Civolilah
> 
> We'll want to present any error messages back in context. I can see that the 422's are displayed correctly, however on 400 (failure) the response shape is:
> 
> {"message":"Could not send a test email using the selected mailer."}

@turbo124 Just a quick confirmation, the solution is now updated, and 400 errors should also display correctly. Let me know your thoughts.

beganovich (approved):
Thanks 👍

PR #3358, opened 2026-09-01
[Cleanup] Pre-fill Recover Password Email

Civolilah:
@beganovich @turbo124 The PR includes implementation of logic for pre-filling the recover password email field with the value entered on the login screen. Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #3360, opened 2026-09-03
[Cleanup] Custom Fields For Design

Civolilah:
@beganovich @turbo124 The PR includes the implementation of an additional flow for when new invoice or product custom fields are added. This means that when we save custom fields, if new fields are detected, we will display the modal shown in the screenshots. The modal essentially lists the new custom fields detected, with a toggle for each, allowing the user to choose whether to include them in the invoice design. If the toggle is turned on during save, the field will also be saved to the design. There is also a sort button, which displays a second version of the modal that allows sorting of variables, which are saved as they are on the invoice design page. Screenshots:

Screenshot 2026-09-04 at 12 04 52

Screenshot 2026-09-04 at 12 05 16

Let me know your thoughts.


Civolilah:
@turbo124 Just to confirm, per your recommendation, which I also agree with, we simplified the flow a little bit. So, in the first step, we have custom fields that are added with "No" and "Yes" buttons. In the second step, if the user clicks the "Yes" button, we simply offer to sort it and save. Screenshots:

Screenshot 2026-09-07 at 23 03 35

Screenshot 2026-09-07 at 23 04 56

Let me know your thoughts.

turbo124:
@Civolilah if i add a product And invoice custom field, the popup only handle the invoice, and drops the product custom field completely?

Could you please write matching playwright tests to confirm the functionality and cover these types of scenarios.

Civolilah:
> @Civolilah if i add a product And invoice custom field, the popup only handle the invoice, and drops the product custom field completely?
> 
> Could you please write matching playwright tests to confirm the functionality and cover these types of scenarios.

@turbo124 No worries, it has already been covered as a possible case. So, if a user enters a product and an invoice custom field at the same time, both modals will be adjusted per input. The first modal will explicitly state which columns are from the product and which are from the invoice, with toggle functionality. The second one will provide the possibility of sorting and saving both. Also, tests have been implemented for this functionality and its cases. Screenshots:

Screenshot 2026-09-08 at 12 14 41

Screenshot 2026-09-08 at 12 16 52

Let me know your thoughts.

turbo124:
@Civolilah We should also support the surcharge fields

image

Civolilah:
> @Civolilah We should also support the surcharge fields
> 
> image

@turbo124 Sure thing, surcharge fields have been implemented as part of this UX improvement, but since they are part of the totals columns in the invoice design, we needed to add another column for this purpose, which I did. Screenshots:

Screenshot 2026-09-09 at 23 15 44

Screenshot 2026-09-09 at 23 17 01

Let me know your thoughts.

beganovich (approved):
Thanks 👍

PR #3365, opened 2026-09-04
[Cleanup] Displaying Delete Payment Action Conditionally

Civolilah:
@beganovich @turbo124 The PR includes the addition of logic for displaying the delete payment action. So, if a payment is made on any invoice that is in a deleted status, the action will not be displayed. Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #3366, opened 2026-09-05
Ensure expense_id set when cloning to PO

turbo124:
When converting invoice => PO, expense_id was not set causing a white screen.


beganovich (approved):
Thanks 👍

PR #3367, opened 2026-09-05
2026 06 09 fix table sorting preferences

turbo124:
Applies fixes for table filter preferences:

Currently the preferences ONLY update if a filter +/- status filter are set. this can cause sorting changes to not persist to preferences.


beganovich (approved):
Thanks 👍

PR #3370, opened 2026-09-07
[Cleanup] Implementing Convert To Purchase Order Action

Civolilah:
@beganovich @turbo124 The PR includes the implementation of the "Convert to Purchase Order" action for invoices and quotes. I picked a modal for invoices as well because I guess we will have more of those actions in the future, but for quotes, it is just one more action in the modal. Screenshots:

Screenshot 2026-09-09 at 21 13 00

Screenshot 2026-09-09 at 21 13 19

Let me know your thoughts.

### DIFF
--- a/src/common/hooks/useCommonActions.ts
+++ b/src/common/hooks/useCommonActions.ts
@@ -34,6 +34,7 @@ export function useAllCommonActions() {
       { value: 'client_portal', label: t('client_portal') },
       { value: 'cancel_invoice', label: t('cancel_invoice') },
       { value: 'reverse', label: t('reverse') },
+      { value: 'convert_to', label: t('convert_to') },
       { value: 'clone_to', label: t('clone_to') },
       { value: 'archive', label: t('archive') },
       { value: 'restore', label: t('restore') },
--- a/src/common/interfaces/purchase-order.ts
+++ b/src/common/interfaces/purchase-order.ts
@@ -75,6 +75,7 @@ export interface PurchaseOrder {
   paid_to_date: number;
   subscription_id: string;
   expense_id: string;
+  invoice_id: string;
   invitations: Invitation[];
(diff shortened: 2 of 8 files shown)


turbo124:
on conversion you need to take the response and pass that through to the create page, ie, the API only stubs the correct purchase order shape which is not yet persisted.

Civolilah:
> on conversion you need to take the response and pass that through to the create page, ie, the API only stubs the correct purchase order shape which is not yet persisted.

@turbo124 As we discussed on Slack, I expected the currently active shape of the response, which is an array, but we should switch it to a single object. So I assumed it like that, please let me know if it works great on your end.

turbo124:
image
@Civolilah we will want to ensure from Purchase Order list. If a quote_id or invoice_id is present. we provide a link back to either reference (could be both!)

Civolilah:
> image @Civolilah we will want to ensure from Purchase Order list. If a quote_id or invoice_id is present. we provide a link back to either reference (could be both!)

@turbo124 Sure thing, it has been implemented. There are icons in the status column on the purchase orders table which link to an invoice or a quote, or there can be both icons. Screenshot:

Screenshot 2026-09-14 at 10 27 56

Let me know your thoughts.

turbo124:
@Civolilah this looks good, is it possible to extend this and have a tool tip over that icon advising the document type? ie invoice/quote etc just to provide the right context?

Civolilah:
> @Civolilah this looks good, is it possible to extend this and have a tool tip over that icon advising the document type? ie invoice/quote etc just to provide the right context?

@turbo124 Sure thing! The tooltip has been added. Screenshot:

Screenshot 2026-09-14 at 12 23 32

Let me know your thoughts.

beganovich (approved):

PR #3371, opened 2026-09-08
[Cleanup] Implementing Validation For Tag Names

Civolilah:
@beganovich @turbo124 The PR includes implementation of validation for tag names, not allowing commas in the name either on creation or editing. Screenshots:

Screenshot 2026-09-08 at 21 55 36

Screenshot 2026-09-08 at 21 55 57

`Note`: As we do not have any translation keyword for this purpose, I assumed "commas_not_allowed". So, if you agree with it, please just add it to the translation files.

Let me know your thoughts.


beganovich (approved):
Thanks 👍

PR #3374, opened 2026-09-09
Fixes to dismiss toast after invoice project action

turbo124:
Fixes a bug where the processing toast does not dismiss after using the Invoice Project action.


beganovich (approved):
Thanks 👍

PR #3376, opened 2026-09-11
Fixes for product creation error

turbo124:
image

When creating a new invoice and using the + product selector. Pressing Enter without highlighting a product, pressing Tab, or clicking outside can emit an entry with resource: null. 

The dialog blindly appends product.resource as Product to its selected products causing an error to throw

This fix add an optional nullable prop which we can gate on to prevent this.


beganovich (approved):
Thanks 👍

PR #3378, opened 2026-09-11
Task / Project improvements

turbo124:
- New date range query to allow tasks which span over multiple days to be visible on the monthly calendar view, not just from its start date.
- Implement navigation aware modal in case debounce saving is midflight
- Rearrange tabs in Project overview
- Add expenses list to Project overview
- Add color picker to project create/edit screens.


beganovich (approved):
Thanks 👍

PR #3379, opened 2026-09-14
Support selecting company with query parameter

hillelcoren:
We're working on adding deep link support in the new mobile apps. If the mobile app isn't installed we fall back to the React web app. This change enables us to select the active company using a query parameter. 

Note: you may want to consider automatically adding the company id in the browser URL so users can more easily share URLs when they have multiple companies.


beganovich:
@hillelcoren please target `develop` branch. Thanks!

hillelcoren:
Thanks, will do...

PR #3383, opened 2026-09-16
[Feature] Quick Task Creation Project Show Page

Civolilah:
@beganovich @turbo124 The PR includes the implementation of quick task creation functionality from the project tasks table. Screenshot:

Screenshot 2026-09-17 at 00 47 59

Let me know your thoughts.


turbo124:
I like it, my only question is the location and label, i personally think it may be better to add this at the bottom, and use a different label to indicate the functionality. ie

```
Quick Add Task ↵
```

@beganovich thoughts?

beganovich:
> Quick Add Task ↵

image

Yeah, it makes most sense on bottom. I am fine with both labels, though.

@turbo124

turbo124:
@Civolilah I think the UI with the row just below the table + create makes sense.

Civolilah:
> @Civolilah I think the UI with the row just below the table + create makes sense.

@turbo124 Sure thing! So, as we discussed, the field is in the last row of the table, and the UX is also improved. Screenshot:

Screenshot 2026-09-21 at 10 14 28

Let me know your thoughts.

beganovich (approved):

PR #3388, opened 2026-09-18
Support selecting company with query parameter

hillelcoren:
We're working on adding deep link support in the new mobile apps. If the mobile app isn't installed we fall back to the React web app. This change enables us to select the active company using a query parameter. 

Note: you may want to consider automatically adding the company id in the browser URL so users can more easily share URLs when they have multiple companies. 

---

Supersedes #3379, retargeted from `main` to `develop` as requested. Same two commits, replayed onto `develop` — identical diff.


turbo124:
I'm not sure how flutter handles this particular scenario, but I think if this is merged, when the link is opened (typically from an email client it will open in a new window) any existing sessions that are logged in under a _different_ company will no longer be valid and will log themselves out when they regain focus.

hillelcoren:
Good catch. There are actually two issues here.

As written, the PR only changes the index. useAuthenticated still authenticates with the stored X-NINJA-TOKEN, so requests would still go to the old company.

Fixing that naively causes exactly what you describe, though it shows up as a silent company mix-up rather than a logout. The token is shared in localStorage and read on every request, so switching it in one tab affects all of them. The same thing already happens today if you switch company in a second tab.

I'll rework it so a ?company= link keeps its token and index in sessionStorage, scoped to that tab, and leaves other tabs alone.

hillelcoren:
I've pushed a change that fixes it.

Before this change, a ?company= link had two problems:

- Wrong token: It only changed the company index, not the token. Requests kept going to the previous company, so the linked record would fail to load.
- Cross-tab switching: Fixing that the obvious way would have caused the problem you described. The token is shared by all tabs in localStorage and read on every request, so other open tabs would have silently switched to the linked company's data. Nothing in the UI shows this, which makes it worse.

Now a link to a different company stores that company's token and index in sessionStorage, which only that tab can see. Other tabs aren't affected. Using the company switcher, logging in or logging out in that tab clears it, and the tab goes back to the company all tabs share. The ?company= parameter is removed from the URL after it's read, so a reload can't undo a company switch.

Limitations:

- DocuNinja: DocuNinja's token is also stored per company in localStorage, so a link tab still switches DocuNinja in other tabs. Switching company in a second tab already does the same today. I'd suggest fixing this in a separate PR since it touches around 35 files.
- First load only: ?company= only applies when the page first loads. A tab that's already signed in ignores it. That's fine in practice since links from email open a new tab.

